PR #1, opened 2011-04-18
Nexus 3977 - make tasks stoppable

cstamas:
Minor changes in task abstract classes

REST Resource made honor a flag, to be able to "cancel only" (but not remove) task.

ReindexTask made stoppable

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/scheduling/NexusTask.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/scheduling/NexusTask.java
@@ -65,6 +65,8 @@ public interface NexusTask<T>
      * @return task name (or null if not available)
      */
     String getName();
+    
+    boolean isExposed();
 
     /**
      * Should an alert email be sent?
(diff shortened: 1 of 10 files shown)


velo:
Seems great, but, shouldn't we make all tasks stoppable?  Or that is for a second phase?

velo:
Sorry, pressed the wrong button ;)

cstamas:
Yes, the 2nd commit contains methods from DefaultIndexManager "pinted" with stop support. Just as an example.

As I keep telling, and actually makes me feel a bit uncomfortable with this, is that our task components are mostly trivial one-liner classes, that call into another component (like ReindexTask calling into DefaultIndexManager.reindex() method).

Hence, the source being executed as "background task" is completely unaware where it runs, moreover, I think (unverified yet) that we have some methods called from both: synchronously from core, and from tasks.

The newly introduced TaskUtil contains helper methods to get the current progress listener, and every progress listener method also performs a "stop check". So, code below is -- from stopping perspective -- identical, both will stop:

TaskUtil.checkInterruption()  
TaskUtil.getCurrent().(beginTask | work | endTask)

So, the two lines above will stop anyway.

cstamas:
Ok, I am putting this request on hold, since Toby came up with different solution on last scrum.

So, I'd like to see his solution too and then have a discussion about comparing the two and choosing which one to implement. This change is "parked" on my private branch anyway, see below:

https://github.com/cstamas/nexus/tree/nexus-3977

PR #4, opened 2011-08-23
it's its, not it's: http://garyes.stormloader.com/its.html

laszlovandenhoek:
I know, I know, it's a nitpicky detail, but I twitch every time I read it in my log file :P


cstamas:
Thanks!

PR #6, opened 2011-10-14
NXCM-4570, DefaultFSLocalRepositoryStorage will now ignore ItemNotFoundExceptions when listing directories.

bdemers:


cstamas:
Accepted, but still: the fact this may happen means that locking is not applied to children... and that should be fixed. Right now, this change just "shoves" the actual problem under the carpet.

PR #8, opened 2011-10-24
NEXUS-4588: SnapshotRemover task cancellation is not properly handled

cstamas:
Snapshot remover makes two passes ("walks" in Nx lingo). Walker and entry was checking for cancellation, but 2nd pass did not.

Facts:
- The cancellation detection was added to the very beginning (entry) of the workhorse method
- The Walker does detects cancellation
- The 2nd pass was a simple for-cycle of paths collected in 1st pass, and while it was not doing a lot, it was iterating over the collection even if cancel happened.


cstamas:
Btw, even before this patch, the SnapRemover task upon cancellation would leave your repository in broken state: some snapshot JARs might be already deleted, but the metadata (maintained in 2nd pass) would not reflect their removals... This is just a FYI, we should probably note somewhere in doco...

bdemers:
+1, but add the missing test license headers before merging

nabcos:
missing license header and javadoc TODOs, otherwise +1

cstamas:
Added license headers and some javadoc explaining what happens in tests.

PR #9, opened 2011-10-24
NEXUS-4574 - Give Nexus179* test time to run async events

velo:
I ran it here several times and was not able cause problem here.

CI is also running fine for 20 times now (other failures but not this one)

Debugging I noticed async events fire, so I gave time to run.


bdemers:
+1

nabcos:
+1

PR #11, opened 2011-10-25
Making it possible to run IT's in IDEA

bdemers:


bdemers:
Looking for thoughts/opinions

nabcos:
+1, works around the IDEA bug with custom type dependencies

cstamas:
Merged, since the same hack was done on Overlord and was working okay.

PR #13, opened 2011-10-25
NXCM-3463 - wait for all async events and tasks to finish before trying t

velo:
Both NXCM-3463 and NXCM-3462 seem to be related to file downloading.

I think we should wait all events to run before we try to download anything.  Anyone disagrees?

### DIFF
--- a/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
+++ b/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
@@ -1080,6 +1080,15 @@ public abstract class AbstractNexusIntegrationTest
     public File downloadFile( URL url, String targetFile )
         throws IOException
     {
+        try
+        {
+            TaskScheduleUtil.waitForAllTasksToStop();
+            getEventInspectorsUtil().waitForCalmPeriod();
+        }
+        catch ( Exception e )
+        {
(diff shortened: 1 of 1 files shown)


velo:
Also related to NXCM-3471

velo:
One more NXCM-3472

bdemers:
Why do we need to wait for tasks to finish to download files?  This seems wrong

velo:
4 different tests failed at same condition, trying to download something that was not available due to some async condition.

But what is your concern here?

bdemers:
It is more of where the change is, you made all downloads wait for tasks to finish.  So it becomes harder to verify that I could download file XYZ-release.jar while the snapshot remover task is running.

peterlynch:
I want to make sure that by waiting we are not hiding the real problem? Users will not wait for calm period...

velo:
@bdemers ow, you are absolutely right.  Will nuke this pull and do this changes up the hill.

velo:
@peterlynch well, if nexus has a async event and until the async event finishes a item won't be downloadable that is something users would need to deal with.

Is like what happen to indexer.... you upload something, a async event is adding the uploaded artifact entry to the index. Until this event is executed there is no entry on the index.  So tests need to wait all events to be executed.  But users are most unlike to face this since this events take seconds or even miliseconds to run.  Even the ITs do not notice this most of the times, but then if there is some performance drop and the async event took 3 seconds to run the IT would fail cause the is no entry on the index....

peterlynch:
+1 ok good explanation - hard to see the entire context in small diff

velo:
although this do not apply to this, as toby pointed out we may have tests that I wanna download in the middle of a task

PR #14, opened 2011-10-25
Nxcm 3230 test log levels

peterlynch:
Modules which use nexus-test-common artifact in test scope will now expose logging level of the ITs, UTs and nexus under test to a system property evaluated via logback.

```
mvn clean test -Dit.test=MyNoisyTest -Pdebug
```

or

```
mvn clean verify -Dit.test=MyNoisyIT -Dit.test.log.level=WARN -Dit.nexus.log.level=WARN -Dit=true
```

for example


velo:
Liked mvn -Pdebug  idea
+1

bdemers:
+1

PR #16, opened 2011-10-25
NEXUS-4332: fixing the logging of REST clients to adapt to Maven logging level and more

cstamas:
"verboseOutput" parameter removed, and Maven's "-X" is honored only.

Swapped logging to logback too.

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/pom.xml
+++ b/nexus-maven-plugins/nexus-maven-plugin/pom.xml
@@ -115,6 +115,9 @@
       <scope>test</scope>
     </dependency>
     <!-- end of required ordering... -->
+
+    <!-- mojo-commons shall be used by this (and any other Sonatype) Maven plugin -->
+    <!-- Currently, mojo-commons is not yet fully usable for this case, but Stuart will make it usable -->
     <!--dependency>
       <groupId>org.sonatype.maven</groupId>
       <artifactId>mojo-commons</artifactId>
--- a/nexus-maven-plugins/nexus-maven-plugin/pom.xml
+++ b/nexus-maven-plugins/nexus-maven-plugin/pom.xml
@@ -20,7 +20,7 @@
 
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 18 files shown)


bdemers:
I would prefer the loggers not be in the constructor

nexus-maven-plugins/nexus-maven-plugin/pom.xml
```diff
@@ -114,15 +115,18 @@
       <scope>test</scope>
     </dependency>
     <!-- end of required ordering... -->
+    <!--dependency>
+      <groupId>org.sonatype.maven</groupId>
+      <artifactId>mojo-commons</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
+    </dependency-->
```
peterlynch, line 125: So you decided not to use this? Why commented out.

cstamas, line 125: Because it's not suitable quite well -- and initially I thought it is.

What happens in, Mojo Log is _injected_ into Mojo instance. But, there is the "discoverer" component, that uses Mojo Javadoc annotation of @component... Meaning, when I get to the point to invoke "syncLogbackWithMavenLog...", with _injected_ Log, I already have instantiated discoverer too, that was using static logger initialization.

Now, this would be fine, as I again thought, if I'd would be in situation like Insight scanner is: a pure SISU module could be started... but, the discover (indirectly), depends on components coming from Maven core, so I'd need a SISU container (that would be made synced and logback enabled), that gets SISU and plexus components, but is also "aware" of it's encapsulating Plexus, where Maven and this mojo exists....

A bit complicated, so this is why I stopped using this artifact (there are already comments how to improve it anyway), and I implemented a hack: I made AbstractNexusMojo Contextualizable, doing some dirty work there _and_ lazy looking up the discoverer _after_ I set up logger (and not having instance of Mojo Log injected yet). From tests, we do direct setter invocation, so change does not matter at all.

This is also a good candidate to move out to mojo-commons.

peterlynch, line 125: Any reason to keep the commented out pom snippet?

cstamas, line 125: I'd rather keep it to avoid mistakes from past: as mojo-commons are shaping in this very moment, it would be a "reminder" to whoever picks up this task in the future to checks where it is, since ultimately, we should all use that new artifact... even if it's unused, since the problem resolved here is very specific. Maybe a new MavenAwareSISUContainer or MavenAwareModule could be added, that is able to suck components from surrounding plexus container for example? That would solve the problem here,

Anyway, the commented part does not hurt at all in any other way.

peterlynch, line 125: Don't mind leaving commented out dependency in pom, but if you leave it there without a comment explaining why you want to keep it there, someone will inevitably come by and remove it later unknowingly.

```
<!-- keep as reminder we would like to use this module one day to fix xxx -->
```

PR #18, opened 2011-10-25
Adding tests to veryify sessions are handled the same way for anonymous users (we may change this, as anon users do NOT need a session)

bdemers:


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
```diff
@@ -110,6 +109,68 @@ public class Nexus4257CookieVerificationIT
...
+
+        Cookie sessionCookie = this.getSessionCookie( httpClient.getState().getCookies() );
+        Assert.assertNotNull( sessionCookie, "Session Cookie not set" );
+
+        
+        httpClient.getState().clear(); // remove cookies, credentials, etc
+        // set the cookie expect greatness
+        httpClient.getState().addCookie( sessionCookie );
+        getMethod = new GetMethod( url );
+        Assert.assertEquals( httpClient.executeMethod( getMethod ), 200 );
```
nabcos: Not sure what this would assert, it is the same call as before (username and password are unused). How do you know the cookie is used and not recreated? What exactly is tested here? Needs more doc IMO.

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
```diff
@@ -110,6 +109,68 @@ public class Nexus4257CookieVerificationIT
...
+        TestContext context = TestContainer.getInstance().getTestContext();
+        String username = context.getAdminUsername();
+        String password = context.getPassword();
+        String url = this.getBaseNexusUrl() + "content/";
+
+        // default useragent is: Jakarta Commons-HttpClient/3.1[\r][\n]
+        HttpClient httpClient = new HttpClient(); // anonymous access
+
+        GetMethod getMethod = new GetMethod( url );
+        Assert.assertEquals( httpClient.executeMethod( getMethod ), 200 );
```
peterlynch: if assert fails you never release the method. Better to use org.sonatype.nexus.test.utils.ResponseMatchers and/or RequestFacade just in case

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
```diff
@@ -110,6 +109,68 @@ public class Nexus4257CookieVerificationIT
...
+        Cookie sessionCookie = this.getSessionCookie( httpClient.getState().getCookies() );
+        Assert.assertNotNull( sessionCookie, "Session Cookie not set" );
+
+        
+        httpClient.getState().clear(); // remove cookies, credentials, etc
+        // set the cookie expect greatness
+        httpClient.getState().addCookie( sessionCookie );
+        getMethod = new GetMethod( url );
+        Assert.assertEquals( httpClient.executeMethod( getMethod ), 200 );
+        getMethod.releaseConnection();
```
peterlynch: same- assert failing prevents cleanup

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
```diff
@@ -110,6 +109,68 @@ public class Nexus4257CookieVerificationIT
...
+        String url = this.getBaseNexusUrl() + "content/";
+
+        Header header = new Header("User-Agent", "Java/1.6" );
+
+        // default useragent is: Jakarta Commons-HttpClient/3.1[\r][\n]
+        HttpClient httpClient = new HttpClient(); // anonymous access
+
+        GetMethod getMethod = new GetMethod( url );
+        getMethod.addRequestHeader( header );
         Assert.assertEquals( httpClient.executeMethod( getMethod ), 200 );
```
peterlynch, line 169: does method get released?

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
```diff
@@ -110,6 +109,68 @@ public class Nexus4257CookieVerificationIT
 
         GetMethod getMethod = new GetMethod( url );
         getMethod.addRequestHeader( header );
+        getMethod.releaseConnection();
+        Assert.assertEquals( httpClient.executeMethod( getMethod ), 200 );
+
```
peterlynch, line 111: releasing method and then executing it?

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4257/Nexus4257CookieVerificationIT.java
bdemers: I had to use http client directly so i could handle the cookie setting.  The connection release has been cleaned up in the latest commit though.

peterlynch:
not going to -1 this but I'd rather avoid using testng OR junit Assert classes in the future.
 too many differences, especially the order of arguments for messages is in reverse which can lead to errors...

nabcos:
+1 now, but I agree with Peter wrt. assert methods.

bdemers:
I agree with the Asserts, but are we saying that we should change the asserts when we touch a class?  (In this case it isn't a big deal to do)  Maybe we should add a wiki page with items we should consider changing when touching a class?

Also, I just invalidated this new test, we need to stop creating sessions for the Anon user,  I'll resubmit this pull request, as the intent has changed

nabcos:
I like the attitude of always leaving a class you touch in better state than before... Although I have to admit I don't do that always as well. (I would like to quote from "Clean Code" but I can't find the quote right now. ;)

PR #19, opened 2011-10-26
Use path from inside a repository in repository target permission pattern matching

nabcos:
the path has to be /repositories/id/g/a/v/... for the DefaultRepositoryRouter, but only /g/a/v/... for the repository target pattern.

Reference Build: (still running at the time of this pull request)
https://builds.sonatype.org/job/nexus-oss-its-feature/jdk=java-6x,label=linux/1/


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4548/Nexus4548RepoTargetPermissionMatchesPathInRepo.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * IT testing the pattern matching of repository targets for security permissions.
+ * A pattern should fully match paths inside the repo, e.g.
+ * {@code ^/g/a/v/.*} should match the path inside the repository and not need to take
+ * the URL path (/repositories/id/g/a/v/) into account.
+ * <p/>
+ * The nexus config is set up to allow the user test/test access to ^/g/a/v/.* via a repo target permission.
+ */
+public class Nexus4548RepoTargetPermissionMatchesPathInRepo
```
velo, line 49: The test is not gonna run...

Name must be Nexus*IT

PR #22, opened 2011-10-26
Do not run long running tests as part of default pre push profile

adreghiciu:
Renamed all tests (its) that are running for longer then 15 seconds(ish) to LRTest (as in Long Running Test) and excluded them from default profile (prepush).
The tests are still run on another profiles as in "it" profile.

On my machine 2.3 i7/quad it reduced the default build from 16 minutes to about 7.

### DIFF
--- a/nexus/pom.xml
+++ b/nexus/pom.xml
@@ -1769,6 +1769,21 @@
         <enunciate.skip>true</enunciate.skip> -->
         <test-env.skip>true</test-env.skip>
       </properties>
+      <build>
+        <pluginManagement>
+          <plugins>
+            <plugin>
+              <artifactId>maven-surefire-plugin</artifactId>
+              <configuration>
+                <excludes>
+                  <exclude>**/*LRTest.java</exclude>
+                </excludes>
+              </configuration>
+            </plugin>
+          </plugins>
+        </pluginManagement>
+      </build>
+
     </profile>
 
     <!-- dummy profile for activating integration tests -->
--- a/nexus/nexus-api/src/test/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUidLRTest.java
+++ b/nexus/nexus-api/src/test/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUidLRTest.java
@@ -26,7 +26,8 @@ import org.junit.Test;
 import org.sonatype.nexus.proxy.access.Action;
 import org.sonatype.nexus.proxy.repository.Repository;
 
-public class DefaultRepositoryItemUidTest
+// This is an IT just because it runs longer then 15 seconds
+public class DefaultRepositoryItemUidLRTest
 {
     private DummyRepositoryItemUidFactory factory;
 
--- a/nexus/nexus-api/src/test/java/org/sonatype/nexus/proxy/item/SimpleLockResourceLRTest.java
+++ b/nexus/nexus-api/src/test/java/org/sonatype/nexus/proxy/item/SimpleLockResourceLRTest.java
@@ -22,7 +22,7 @@ import java.util.Random;
 
 import org.junit.Test;
 
-public class SimpleLockResourceTest
+public class SimpleLockResourceLRTest
 {
     private static Random rnd = new Random( System.currentTimeMillis() );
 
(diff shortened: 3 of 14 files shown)


bdemers:
I would prefer *IT.java, and run them with the IT profile. what do you guy think ?

adreghiciu:
My preference for this name is to easy spot them.

nexus/pom.xml
```diff
@@ -1769,6 +1769,21 @@
         <enunciate.skip>true</enunciate.skip> -->
         <test-env.skip>true</test-env.skip>
       </properties>
```
velo, line 1771: Shouldn't a new profile be created to run the *LRTest?

bdemers:
Think of the names:
DownloadSomethingLRTest
DownloadToSomethingTest

At first glance they are very similar, which is my main objection, *IT.java was just my first thought.
What about removing the abbreviation? *LongRunTest.java, *LongTest.java, or even *SlowTest.java.

velo:
I agree with toby -1

peterlynch:
Overall great improvement
I agree with toby as well though -1

If they are Unit Tests, then keep the '*Test.java'

If they are IT Tests - something that requires a running Nexus, change to IT.java and use testng.

Lastly I prefer *Slow(Test).java to LR abbrev.. Also if they are ITs, use testng groups may be an option.

(I wish junit @Categories (requires a suite) and surefire's support of them ( ie none) were not so lame so we could just use them instead.)

peterlynch:
your idea is fine for isolating slow tests - but another idea is write a JUnit TestRule that checks the test method or test class for an annotations extending Category ( ie. interface SlowTest extends Category) and compare that with test.category.includes and test.category.excludes systemproperty that lists the category of tests to execute.

Could add as part of litmus...and would not need to wait for surefire, junit..and should be ide friendly too

adreghiciu:
@peter :) that was what I was looking into

Yet I was looking if we could not make the tests appear as ignored, but that is not possible without changing the junit runner. I will make such a rule but also discuss with junit guys. it should be pretty easy to implement a strategy for @Ignore

adreghiciu:
@peter looking more into this, the rule based approach can be misleading because the only thing we could do there is to not run the test. But the runner will still think that the test has being run and report that. Which mean that on test results you will be seing that test has been run when actually it wasn't.
This is a no go for me.

adreghiciu:
@toby I just want to make build shorter. Really I do not care about the name. We can always add an JUnit category to annotate them so is easy to find them. *IT is fine with me.

BTW, using Slow will result in "DownloadToSomethingSlowTest" which tells me that it will download something to something slow, not that test is slow. When you see LRTest you will always ask "what is LR for". And then you either ask somebody or you do not care.

I will change them to IT.

peterlynch:
plynch: actually I can live with LRTest
[4:19pm] adreghiciu: to me it does not matter
[4:19pm] plynch: someone will do MakeSureNexusStartupIsNotSlowTest.java
[4:19pm] plynch: but less likely to do MakeSureNexusStartupIsNotLRTest
[4:20pm] plynch: and besides the LRTest bit is in surefire config ( hopefully with a comment hint )

+1 LRTest

adreghiciu:
Lets vote:

A: keep LRTest
B: use SlowTest

adreghiciu:
My vote A for the reasons I mentioned above

velo:
Well, I still convinced on calling they IT.  If not that as it is right now is good enough for me.

adreghiciu:
@velo

Problems with using IT:
-  naming the slow running ITs as "*IT" introduces a bunch of problems as I have to configure surefire to also look for *IT which will mean that on some places like plugins surefire will try to run also the actual ITs. Alternative is to configure surefire to run on all modules.
- Peter has a point to make a diff between ITs that actually start a nexus server and this tests that does not do that

bdemers:
+1 Slow, or Non abbreviated LRTest -> LongRunningTest

velo:
Failsafe deals out of box with *IT for you

Well, I do not agree with IT being what start nexus server, anyway, the slow ITs in fact start nexus, just not the HTTP server, but everything else gets up and running.
Any test that touches (directly or indirectly) DefaultNexus it is an IT, or at least it will bring enough components to life to earn that title.

But I'm fine with the change we have right now (option A right?)

adreghiciu:
Failsafe plugin is not running by default on all modules. is included only for those modules that have IT, which are quite a few

PR #23, opened 2011-10-27
NEXUS-4520 et al: undefined variables

cstamas:
This pull request, same as issue, is a bit "mish-mash".

Original intent of issue was to fix the lost capability to honor env vars and system properties to get into appcontext that has prefix "PLEXUS_" and "plexus." accordingly.

As the issue itself notes, this was already fixed and is in core.

This pull request contains:
- fix from nexus-mock module that was during build producing a folder "${nexus-work}" next to pom.xml
- fir for IT that was creating a folder "storage" next to pom.xml
- cleaning up git ignores
- IT for verifying the NEXUS-4520 _is fixed_ for case with system properties


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4520/Nexus4520PlexusPrefixedVariablesArePickedUpIT.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonatype.nexus.integrationtests.AbstractNexusIntegrationTest;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * TODO
+ *
+ * @author: cstamas
+ */
```
nabcos, line 39: javadoc missing

nabcos:
aside from javadoc TODO +1

cstamas:
Javadoc + license header of the new IT fixed.

nexus/nexus-mock/src/test/resources/ehcache.xml
```diff
@@ -0,0 +1,372 @@
...
+            maxElementsInMemory="10000"
+            eternal="true"
+            overflowToDisk="false"
+            diskPersistent="false"
+            diskExpiryThreadIntervalSeconds="120"
+            memoryStoreEvictionPolicy="LFU"
+            />
+
+    <!--
+    DiskStore configuration
```
peterlynch: Do people like to have this massive boiler plate comment in this file? - makes hard to read - why not remove it and add a link to the online docs instead?

nexus/nexus-proxy/src/main/resources/ehcache-default.xml
```diff
@@ -32,12 +34,31 @@
...
+            overflowToDisk="false"
+            diskPersistent="false"
+            diskExpiryThreadIntervalSeconds="120"
+            memoryStoreEvictionPolicy="LFU"
+            />
+
+    <cache name="shiro-activeSessionCache"
+            maxElementsInMemory="10000"
+            eternal="true"
+            overflowToDisk="false"
```
peterlynch, line 61: again would like to remove the massive boiler plate comment in this file

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4520/Nexus4520PlexusPrefixedVariablesArePickedUpIT.java
```diff
@@ -0,0 +1,102 @@
...
+import org.codehaus.plexus.util.StringOutputStream;
+import org.sonatype.nexus.integrationtests.AbstractNexusIntegrationTest;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * IT for issue NEXUS-4520, covering the "System properties" case. In short, before Nexus is booted, it sets a
+ * system propery like {@code plexus.foo=bar} (simulating option -Dplexus.foo=bar in wrapper.conf for example) and
+ * then checks for the presence of {@code foo=bar} key-value pair in AppContext.
```
peterlynch, line 36: thx for the comment

nexus/nexus-mock/src/test/resources/ehcache.xml
cstamas: Actually, I just realized that this file here (in nexus-mock) is unneeded. The problem was that ehcache-default.xml in nexus-proxy was outdated. Meaning, this file should be removed.

And wrt your "boiler plate" comment: these files are _delivered within jar_, and are just "fallbacks" when no real config is found. They are not meant for end users editing (and they are anyway unable to edit these since they are packed into nexus-proxy JAR).

But, these files do originate from vanilla ehcache distro, meaning, leaving the as is and having only applied "our" (nexus specific) changes is handy, since you can track the changes using diff.

peterlynch: ok sounds good.

peterlynch:
+1

cstamas:
Matcher changed to `containsString`, as suggested, but also reduced the matched string, to not have IT failures if AppContext "dump format" (something totally unrelated to Nexus) changes in future.

PR #24, opened 2011-10-27
NEXUS-4548: IT to test pattern matching for security repo target permission

nabcos:
Last pull request + 'fix' was premature, this IT also works with current master, so no change is needed for the cases tested here (get/put/upload via rest).


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus4548/Nexus4548RepoTargetPermissionMatchesPathInRepoIT.java
```diff
@@ -0,0 +1,168 @@
...
+            client.getState().setCredentials( AuthScope.ANY, new UsernamePasswordCredentials( "test", "test" ) );
+            putMethod.setRequestEntity( new FileRequestEntity( getTestFile( "pom-a.pom" ), "text/xml" ) );
+
+            assertThat( client.executeMethod( putMethod ), Matchers.is( code ) );
+        }
+        finally
+        {
+            if ( putMethod != null )
+            {
+                putMethod.releaseConnection();
```
peterlynch: so just checking ... we don't need to do anything special with cleaning up connection manager used by httpclient? like call shutdown?

So this is second case I saw where HttpClient needed to be used directly - it would make more sense to me to have the boilerplate release in a class similar to REST Light ResponseMatcher

velo:
Any reason to use org.apache.commons.httpclient.\* instead of using our RequestFacade (based on org.restlet.*)?

nabcos:
I had problems with RequestFacade always using admin access, but maybe I just did not set the TestContext at the right time... IT uses RequestFacade now.

cstamas:
+1 for IT and this pull request, but wtf is the NEXUS-4548 about then?

nabcos:
Closed with 'cannot reproduce', I was not able to trigger the described behavior.

PR #26, opened 2011-10-27
Disable sessions for the anonymous user

bdemers:


peterlynch:
+1

adreghiciu:
I'm not good at this stuff but here are my 2c:
Wy we are not creating the cookie just in case that a certain header is set? like when "X-Nexus-Session" is set?
Of course this is an alternative only if we can send that header from Nexus UI once user logs in.

This will eliminate any need to check for user agent.

bdemers:
@adreghiciu

I agree, i'd like to do that too, but that may have a larger impact, and there are a few directions we could take.  

This solution is for a patch, and should help the problem, but it is not the ultimate solution

nabcos:
+1

PR #27, opened 2011-10-28
[NEXUS-4361] Set UTF-8 as default character set for post/ajax requests

adreghiciu:
[NEXUS-4361] Set UTF-8 as default character set for post/ajax requests to server

Looks like from the 3 browsers I tested with (Safari,Chrome,FF) only FF will automatically add UTF-8 as character set. To make the UTF-8 encoding work also in the other browsers, the fix will set the charset on Content-Type request header to UTF-8.

### DIFF
--- a/nexus/nexus-webapp/src/main/webapp/js/Sonatype.config.js
+++ b/nexus/nexus-webapp/src/main/webapp/js/Sonatype.config.js
@@ -27,7 +27,7 @@
 
   // Set default HTTP headers //@todo: move this to some other common init
   // section
-  Ext.lib.Ajax.defaultPostHeader = 'application/json';
+  Ext.lib.Ajax.defaultPostHeader = 'application/json; charset=utf-8';
 
   // set Sonatype defaults for Ext widgets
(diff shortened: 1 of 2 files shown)


velo:
+1

bdemers:
+1, I wish we could verify this with selenium

adreghiciu:
I'm not sure that actually this is "the solution". Looks like even without this it was working fine in nexus 1.8. What I see as the diff is that we started using jetty 7 compared with jetty 6 in 1.8

So, I'm doing further investigations

adreghiciu:
I spent the last 6 hours on debugging and comparing 1.8 with 1.10 (1.10 without above change)
I.87 works just fine wrt charset encoding but 1.10 does not. The obvious diff between them is the jetty version where 1.8 uses jetty 6 and 1.10 used jetty 7.
But, I compared the extjs generated html and relevant headers and they all seem the same. I even run 1.10 with java script from 1.8. Still 1.8 work, but 1.10 does not.

Really I can not say what the diff. I just give up and use the changes I made above, even if to me they seam as an workaround if everything worked before.

PR #32, opened 2011-10-31
NEXUS-4570, adding catch for FileNotFoundException

bdemers:
No real way to write a test for this


peterlynch:
this usually means code is written to be untestable. You could refactor the code or use mocks and make sure ItemNotFoundException is thrown when your mock throws FileNotFoundException.

These are subtle changes that are affecting what can get thrown from this method. I would try to write a test for it. At the very least, add a simple comment next to code referencing what you are fixing.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSLocalRepositoryStorage.java
```diff
@@ -325,6 +326,10 @@ public class DefaultFSLocalRepositoryStorage
                         request, file );
                 }
             }
+            catch ( FileNotFoundException e )
```
velo, line 329: Looking at code this looks like an unreachable branch.

This is probably a multi thread situation where on thread is running other went and deleted the file.  Best case scenario we need a TODO to have  better locking on files.

bdemers, line 329: I can add a better comment.  And it's _possible_ the file is getting removed by another process

velo:
+1

peterlynch:
+1 nice changes

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSLocalRepositoryStorage.java
cstamas, line 329: The file is definitely deleted by something else that is not using UID locking, and that's the culprit. This change here only makes the symptom not visible. But the ultimate question is what modifies the repo in RW manner and avoids locks?

PR #36, opened 2011-11-02
[NXCM-3121] Filter repositories for automatic selection based on user id/ip/agent

adreghiciu:


adreghiciu:
Not yet ready: missing a maven 3 strategy for extracting default user agent

cstamas:
Maven2UserAgentExtractor vs SettingsUserAgentExtractor..... they both do the same: WagonManager _does_ honor the settings.xml value.

The problem is with Maven3 way, you need to get it from session.

Now, to make this work in both maven2 and maven3 (and this plugin is supported on both), you'd need a bit of reflection too, since on one the another is not present and other way around (think maven using AHC Connector, no Wagon at all).

adreghiciu:
Yep, maven 3 is on the way.
That's why I closed the issue without merging.

PR #39, opened 2011-11-03
NEXUS-3996 Make date validation of 'once' scheduled tasks more lenient

nabcos:
Nexus tried to validate the date of a task that is run once without any information about the clients timezone, which failed when the timezone offset from GMT on the client is bigger than on the server.
(Client's midnight in server's timezone was on a different day)

### DIFF
--- a/nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3996/Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT.java
+++ b/nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3996/Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT.java
@@ -46,6 +46,12 @@ import static org.hamcrest.MatcherAssert.assertThat;
  * every midnight that is earlier than local time and not on the same day of the year.
  * <p/>
  * This test simulates a client whose timezone offset is two hours more than the local timezone.
+ * <p/>
+ * Related issues with timezone problems in the scheduled tasks:
+ * <ul>
+ * <li> https://issues.sonatype.org/browse/NEXUS-4617 </li>
+ * <li> https://issues.sonatype.org/browse/NEXUS-4616 </li>
+ * </ul>
  *
  * @since 1.10.0
  */
@@ -67,6 +73,8 @@ public class Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT
         task.setProperties( Lists.newArrayList( property ) );
 
         Calendar cal = Calendar.getInstance();
+
+        // simulating client timezone here: calculate offset from current timezone
 
         // client is tz+2 -> 3 hours ahead for local tz is one hour ahead for client tz
         cal.add( Calendar.HOUR_OF_DAY, 3 );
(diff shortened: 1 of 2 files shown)


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3996/Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT.java
```diff
@@ -0,0 +1,87 @@
...
+ * Nexus would not create tasks for the same day if the client is in a timezone with a larger offset from UTC than
+ * the server's timezone.
+ * <p/>
+ * The date is sent separate from the time as a timestamp for midnight of the selected day. The midnight is calculated
+ * on the client based on the local timezone.
+ * <p/>
+ * The server does not know which timezone the client is in and validates the date based on his timezone, rejecting
+ * every midnight that is earlier than local time and not on the same day of the year.
+ * <p/>
+ * This test simulates a client whose timezone offset is two hours more than the local timezone.
```
peterlynch, line 48: I love useful comments!

peterlynch:
+1

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3996/Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT.java
```diff
@@ -0,0 +1,87 @@
...
+        task.setName( "name" );
+        task.setSchedule( "once" );
+        task.setTypeId( EmptyTrashTaskDescriptor.ID );
+        ScheduledServicePropertyResource property = new ScheduledServicePropertyResource();
+        property.setKey( EmptyTrashTaskDescriptor.OLDER_THAN_FIELD_ID );
+        task.setProperties( Lists.newArrayList( property ) );
+
+        Calendar cal = Calendar.getInstance();
+
+        // client is tz+2 -> 3 hours ahead for local tz is one hour ahead for client tz
```
velo: Where is client tz defined?

nabcos: it's not defined but simulated by manually taking the offset from current timezone into account. So the time is set to 3 hours into the future (we need only HH:mm format here, no TZ info), but the date timestamp is 2 hours earlier than it would be in the current timezone.

So the point in time that nexus parses out of this is one hour ahead in the local timezone and is valid as a start time.

cstamas:
+1

bdemers:
Do we still end up with a timezone miss match?
1.) Server sends the date 5:00 UTC
2.) Client assumes this is EDT
3.) User makes changes to a none time field, and saves data
4.) Client sends date as 12:00 EDT
5.) server updates the time (unexpected)

I may be way off here, this was one of the possible problems we discussed.

nabcos:
@bdemers that is a different issue that may cause a different validation error (not about the start date, but the final 'run time' may end up being before current time). I also talked about that at scrum.

To clean that up, we need either to send the server timezone in the response or switch to using timestamps and not this hybrid "timestamp + clock time" thing. I'd prefer the timestamps, but that means some additional javascript work and changing the REST API.

I filed some issues about that yesterday:
- https://issues.sonatype.org/browse/NEXUS-4617
- https://issues.sonatype.org/browse/NEXUS-4616

And I found this issue yesterday evening, also describing the local/server timezone mismatch:
- https://issues.sonatype.org/browse/NEXUS-4202

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3996/Nexus3996ScheduledTasksTimezoneDifferentDayValidationIT.java
velo: +1

bdemers:
@nabcos ok, thanks for including that here

+1

PR #40, opened 2011-11-04
NEXUS-4257, Catch unknown session exception when logging in the anon user

bdemers:


nabcos:
commit is missing useful log message

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
```diff
@@ -235,15 +236,31 @@ public class NexusHttpAuthenticationFilter
             request.setAttribute( ANONYMOUS_LOGIN, Boolean.TRUE );
 
             subject.login( usernamePasswordToken );
+        }
+        catch( UnknownSessionException e )
+        {
+            Session anonSession = subject.getSession( false );
 
-            if ( getLogger().isDebugEnabled() )
+            this.getLogger().warn( "UnknownSessionException was thrown for with session: '{}'", anonSession, e );
```
nabcos: - typo ("for with")
- log message could be more clear ("Unknown session while logging in anonymous" or sth) because warn is visible with the default log settings
- should this be warn level anyway?

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
```diff
@@ -219,11 +221,10 @@ public class NexusHttpAuthenticationFilter
 
     protected boolean executeAnonymousLogin( ServletRequest request, ServletResponse response )
     {
-        if ( getLogger().isDebugEnabled() )
-        {
-            getLogger().debug( "Attempting to authenticate Subject as Anonymous request..." );
-        }
+        getLogger().debug( "Attempting to authenticate Subject as Anonymous request..." );
```
peterlynch, line 224: couldn't this log message be more informative if the subject information was included in the message?

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
```diff
@@ -307,7 +330,7 @@ public class NexusHttpAuthenticationFilter
             {
                 getSubject( request, response ).logout();
             }
-            catch ( SessionException e )
+            catch ( SessionException e ) //TODO: investigate why this is getting thrown
```
peterlynch: adding reference the issue number on this line would prove useful for the future

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
```diff
@@ -219,11 +221,10 @@ public class NexusHttpAuthenticationFilter
 
     protected boolean executeAnonymousLogin( ServletRequest request, ServletResponse response )
     {
-        if ( getLogger().isDebugEnabled() )
```
velo, line 222: Why removing the getLogger().isDebugEnabled() check?  Just curious

bdemers, line 222: There is not string concatenation.  isDebugEnabled() is better used when doing something like:
log.debug("something: "+ myObject);
or
log.debug("something: "+ someMethodCall() );

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
bdemers, line 224: In this case the subject will always be anonymous, and subject.getPrincipal() would be null. (until after the login)

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
bdemers: You might be right about the log level.  I personally would like to see it, but I don't think the customer wants to see a warning, for something that is recoverable.

bdemers:
@nabcos, good call, I'll fix the commit message when i merge

nabcos:
+1

velo:
+1

peterlynch:
+1

bdemers:
Merged manually (to update the commit message)

PR #41, opened 2011-11-04
Nexus 4539 ITs testing autoblock

velo:
Created 3 ITs to check autoblock.  
Also changed API to make autoblock configurable using System.properties

### DIFF
--- a/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty7NexusBooter.java
+++ b/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty7NexusBooter.java
@@ -156,9 +156,6 @@ public class Jetty7NexusBooter
         // The proper fix is to make all Indexer related ITs behave "properly" (with some heuristics?), and have some
         // sort of "try-wait-try-failAfterSomeRetries" the search operation itself.
         System.setProperty( "mavenIndexerBlockingCommits", Boolean.TRUE.toString() );
-        
-        // NEXUS-4539 - to get test faster reduced autoblock check from 5 minutes to 30 seconds
-        System.setProperty( "plexus.autoblock.remote.status.retain.time", String.valueOf( 30*1000 ) );
 
         // ---------------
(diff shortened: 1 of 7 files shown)


velo:
Nice, it already include the discussion we started during commit to pull request! =D

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty7NexusBooter.java
```diff
@@ -156,6 +156,9 @@ public class Jetty7NexusBooter
         // The proper fix is to make all Indexer related ITs behave "properly" (with some heuristics?), and have some
         // sort of "try-wait-try-failAfterSomeRetries" the search operation itself.
         System.setProperty( "mavenIndexerBlockingCommits", Boolean.TRUE.toString() );
+        
+        // NEXUS-4539 - to get test faster reduced autoblock check from 5 minutes to 30 seconds
+        System.setProperty( "plexus.autoblock.remote.status.retain.time", String.valueOf( 30*1000 ) );
```
cstamas: Why doing this _IT wise_? This way you affect nexus behavior in all ITs, not only in this one.

This should be moved to some of the NEXUS-4539 related IT...

velo: ok, done!

cstamas:
Not a blocker, but just created https://issues.sonatype.org/browse/NEXUS-4620

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty7NexusBooter.java
nabcos: We should add backup/restore of the system properties in @BeforeClass/@AfterClass for the AbstractNexusIntegrationTest.

nabcos:
+1

PR #42, opened 2011-11-04
[NXCM-3121] Filter out repositories for automatic selection based on user

adreghiciu:
[NXCM-3121] Filter out repositories for automatic selection based on user ip and user agent

To do so we make maven resolve an artifact from a "special" repository, handled by a rest resource. The response is a property file containing info about whom made the call. If the artifact could be resolved we then use that info to filter out the repositories for automatic selection.

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
+++ b/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
@@ -299,34 +299,7 @@ public abstract class AbstractStagingMojo
         {
             final ArtifactRepository whoAmIRepository = new DefaultArtifactRepository(
                 serverId, formatUrl( getNexusUrl() ) + SVC_BASE + "/staging/", new DefaultRepositoryLayout()
-            )
-            {
-                @Override
-                public String getUsername()
-                {
-                    if ( AbstractStagingMojo.this.getUsername() != null )
-                    {
-                        return AbstractStagingMojo.this.getUsername();
-                    }
-                    return super.getUsername();
-                }
-
-                @Override
-                public String getPassword()
-                {
-                    if ( AbstractStagingMojo.this.getPassword() != null )
-                    {
-                        return AbstractStagingMojo.this.getPassword();
-                    }
-                    return super.getPassword();
-                }
-
-                @Override
-                public String toString()
-                {
-                    return "(WhoAmI) " + super.toString();
-                }
-            };
+            );
 
             fakeLocal = new DefaultArtifactRepository(
                 "local", createTempFile( "whoami-", "", null ).toURI().toASCIIString(), new DefaultRepositoryLayout()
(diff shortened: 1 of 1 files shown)


nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -234,6 +287,117 @@ public abstract class AbstractStagingMojo
         }
     }
 
+    protected Identity whoAmI()
```
velo, line 299: Missing jdoc

cstamas:
Looks unbelievably involved just to perform so simple thing... also looks very fragile to me.

But as we went over this multiple times... if there is really no proper way than "pinging" remote to get the needed info... :(

velo:
A topic for a future discussion, this maven plugin may should be part of nexus pro, since the fix is partially here and partially there...
https://github.com/sonatype/nexus-enterprise/pull/24/files

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
adreghiciu, line 299: Added in c75d165b

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -234,6 +287,126 @@ public abstract class AbstractStagingMojo
...
+                "local", createTempFile( "whoami-", "", null ).toURI().toASCIIString(), defaultRepositoryLayout
+            );
+
+            // looks like Maven 2 does not give any chance of artifact provided username/password so try a fallback
+            final AuthenticationInfo backupAuthInfo = wagonManager.getAuthenticationInfo( whoAmIRepository.getId() );
+            wagonManager.addAuthenticationInfo( whoAmIRepository.getId(), getUsername(), getPassword(), null, null );
+
+            try
+            {
+                artifactResolver.resolve( whoAmIArtifact, asList( whoAmIRepository ), fakeLocal );
```
velo, line 328: I assume
artifactResolver.resolve( whoAmIArtifact, asList( whoAmIRepository ), fakeLocal );
is used so nexus can tell you the user agent, is that it?

adreghiciu:
@cstamas Do you see an alternative approach? 
For IP we must for sure make a nexus call, but that could be done without involving Maven.
For User-Agent is in my view less error prone then trying to dig it out from Maven.

And even if we run into problems, such as nexus not responding, maven not being able to connect, ..., we then just fall back to current approach = we will not filter out the repositories based on user ip/agent.

velo:
Far enough
+1

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
adreghiciu, line 328: user agent and user ip

PR #46, opened 2011-11-07
NEXUS-4609: After content manipulation, index is not up2date immediately.

cstamas:
A typical case when index synchronicity must be enforced. But this is not true to _every_ case.


nabcos:
+1

bdemers:
+1

PR #47, opened 2011-11-08
NEXUS-4600 Fix failing IT

nabcos:
- Dropped unnecessary FileLock
- Let IT wait until import is done


cstamas:
+1

bdemers:
+1, Seems fine, but I don't know enough about the bridge to say.  Velo may have some thoughts

velo:
+1 seems ok, I don't recall on any concurrency scenarios that would cause problems

PR #49, opened 2011-11-08
NEXUS-4521: more fixes for leak spotted during test debug.

cstamas:
If IO error occurs in the middle of body transfer, the connection
would be left dangling until HttpClient CM timeouts on it. On heavily
loaded systems this might lead to connection pool depletion. Now
the method.releaseConnection() will be invoked as quick as possible.

Also, in the UT the config change was not applied explicitly.

This are still not the solutions, see comments on NEXUS-4521 issue
about setSoTimeout() method.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/commonshttpclient/HttpClientInputStream.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/commonshttpclient/HttpClientInputStream.java
@@ -30,20 +30,88 @@ import org.sonatype.nexus.util.WrappingInputStream;
 public class HttpClientInputStream
     extends WrappingInputStream
 {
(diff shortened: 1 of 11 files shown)


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/commonshttpclient/HttpClientInputStream.java
```diff
@@ -30,23 +30,91 @@ import org.sonatype.nexus.util.WrappingInputStream;
...
+     */
     private final HttpMethod method;
 
     /**
      * Instantiates a new http client input stream.
-     * 
+     *
      * @param method the method
-     * @param is the is
+     * @param is     the is
```
velo, line 43: =(

velo:
+1

nabcos:
+1

PR #51, opened 2011-11-09
[NEXUS-4593] Don't autoblock proxy repos for remotes answering 403

nabcos:


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepository.java
```diff
@@ -1091,7 +1091,9 @@ public abstract class AbstractProxyRepository
                         }
                         catch ( StorageException ex )
                         {
-                            if ( ex instanceof RemoteStorageException )
+                            if ( ex instanceof RemoteStorageException
```
velo, line 1094: Why don't you just add an extra catch?
                catch ( RemoteAccessDeniedException e )
                {
                    // 403 exception
                }

nabcos, line 1094: I didn't because there is a cleanup of local item in that catch block, which would have to be duplicated.

velo, line 1094: BTW, you need to autoUnBlockProxying(); when it happen.

Repo could be timing out, got back up and start 403.

nabcos:
This is only a partial solution so far, I missed the code path when you already have the artifact and only need a check for lastModified. That is also what's used for the autoUnblock checks. Closing this pull request and reworking.

nabcos:
Some ITs have to be adapted: https://builds.sonatype.org/job/nexus-oss-its-feature/jdk=java-6x,label=linux/27/

nabcos:
Actually the IT failures did not have to do anything with the branch? Next build succeeded without changes: https://builds.sonatype.org/job/nexus-oss-its-feature/jdk=java-6x,label=linux/28/

PR #52, opened 2011-11-09
NEXUS-4595: add a filter to request LDAP server

alecharp:
Add a field in LDAP Plugin to filter the user requests


bdemers:
Sweet!  At first glance it looks good.  I'll take a closer look in a bit.
Can you add a test for it ?

alecharp:
I will work on it

## 

Adrien Lecharpentier

2011/11/9 Brian Demers <
reply@reply.github.com>

alecharp:
I added tests, maybe not enough. If it needs more, just tell me

nabcos:
see https://github.com/sonatype/nexus/pull/70

PR #53, opened 2011-11-09
[NEXUS-4624] Log skipped reindexing as re-indexing is already in place

adreghiciu:
We should be able to easier spot the cases when re-indexed is expected to happen but is skipped because indexing is already taking place (mainly happens in tests)


nabcos:
+1

cstamas:
♥

peterlynch:
+1

PR #54, opened 2011-11-09
NEXUS-4625: Consolidating core logging.

cstamas:
Consolidate current mess we have in core wrt logging. Realized as "side effect" of working on NEXUS-4521.

What we have here:
- Dropping all the uses of AbstractLogEnabled, switching to AbstractLoggingComponent
- but, since we have classes in core meant to be extended in plugins, keeping "oldie" Plexus logger wherever needed (like AbstractRepository)

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/AbstractApplicationStatusSource.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/AbstractApplicationStatusSource.java
@@ -28,13 +28,12 @@ import javax.inject.Inject;
 
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.logging.Logger;
+import org.sonatype.nexus.logging.Slf4jPlexusLogger;
 
 public abstract class AbstractApplicationStatusSource
     implements ApplicationStatusSource
 {
-    @Requirement
-    @Inject
-    private Logger logger;
+    private Logger logger = Slf4jPlexusLogger.getPlexusLogger( getClass() );
 
     /**
      * System status.
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/logging/AbstractLoggingComponent.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/logging/AbstractLoggingComponent.java
@@ -18,9 +18,8 @@
  */
 package org.sonatype.nexus.logging;
 
-import javax.inject.Inject;
+import static com.google.common.base.Preconditions.checkNotNull;
 
-import org.codehaus.plexus.component.annotations.Requirement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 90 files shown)


adreghiciu:
+1
Just wonder why you did not opt for something similar we had in Overlord:
Have a root abstract class ComponentSupport that has getLogger() that creates the logger lazy.
Doing so, as we will always use getLogger() you do not get an import in your class meaning that an eventual changing of logger type would not need sources changes.

peterlynch:
+1
I like Alan's idea for new components - we should depend in sisu-goodies ( which has this ComponentSupport class ) in the near future

peterlynch:
Idea: Do a full build including ITs with and without this patch, for each find all *.log files and compute size. Verify through random sampling this gives expected results

cstamas:
@Alan (sic!): getLogger() is what you can use here too, since Logger is a private member.... true, is not lazily created... but the idea is same, we can still play with it in future, since all that extending classes can reach is getLogger() method.

cstamas:
@peterlynch: I was doing for one day following

```
  $ mvn clean install -Dtest=SimpleRemoteLeakTest -o
```

And I know what the diff is, believe me. But, anyone is free to do the two and compare the -output.txt files...

cstamas:
@MrFord (sic!) did you mean this class?

https://github.com/sonatype/sisu-goodies/blob/master/common/src/main/java/org/sonatype/sisu/goodies/common/ComponentSupport.java

It is not lazy, and it makes you depend (and thus have import) of the logger member, right?

adreghiciu:
@Tamas I personally do not like that approach. It may be done like that just to save the time for calling the getLogger() method.

Yet if we switch to goodies then we have a common base all over the place. Insight guys were also interested in goodies.

cstamas:
Now am puzzled. That class I linked above is from goodies... Is that the one you dislike, or the AbstractLoggingComponent or both?

adreghiciu:
The goodies one.

cstamas:
Just to explain:
- I dislike the usual `protected static Logger logger` solution with static logger initializations
- I don't think @Injecting loggers is good either. See SISU-89 but also, why would we do it in 1st place? Slf4j API itself is already a level of indirection...
- agree with Alin about 'protected Logger logger' vs 'getLogger()' use (prefer latter)

But... I don't want to go with "magic" (oldie) Plexus way either. I like explicit things, so these two AbstractLoggingComponent and AbstractPlexusLoggingComponent are the maximum i'd go with logging.

Also, we _must_ wrap slf4j Logger into Plexus Logger, since some of components (see AbstractRepository) already had method `protected getLogger()` returning Plexus Logger. Changing this would break all the plugins out there, that contributes with new repository type (P2, OBR, etc), hence until 2.0 we must keep Plexus Loggers in play.

But with this change, we will have the Slf4jPlexusLogger in Nexus codebase, meaning, we will dump one more Plexus dependency. What is left is _lifecycle_... once we sort that out, we can go SISU right away.

Also, I am all for SISU Goodies, but since construction like above would break existing compatibility (code extending Goodies' ComponentSupport expects a protected `logger` field inherited), and it is not yet released (would go with OL? Or am wrong here?), I did not push the changes above in there. But ultimately, if we agree on sensible solution, yes, SISU Goodies is the way to go.

nabcos:
+1

cstamas:
Discussion between @nabcos and me on IRC that sheds more light to this:
- _hanzelm_ cstamas, didn't peter say to disregard OL compat so you could push changes to sisu-goodies? Not sure I understood correctly, but your comment sounds like you would rather have the logging component in sisu-goodies?
- _cstamas_ hanzelm: yes, but i was unsure (and i wrote about it in comment) where sisu-goodies are used in... OL? This was cleared up yesterday, when Peter told me disregard OL... but, I am still unsure about "maturity" of sisu-goodies, and this change will actually _easy_ the transition to it once we pull it into nexus.... currently we do not use it in nexus.... stuff like Closer and Flusher for exmaple are reduntant IMO, since sisu-goodies depend on guava, and guave itself has those stuff
- so, in short, yes, we can move to goodies, but the move should not block this commit, and actually _after_ this commit, the move would be much simpler
- just modify abstractloggingcomponent to extend componentsupport and done
- _hanzelm_ ok, +1 it is then.

peterlynch:
do it +1

adreghiciu:
+1

PR #55, opened 2011-11-10
NEXUS-4205 - set content validation to true

velo:
Enable fileTypeValidation by default for new proxy repositories and for the proxy's that ship in the default configuration.

### DIFF
--- a/nexus/nexus-configuration-model/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepositoryConfiguration.java
+++ b/nexus/nexus-configuration-model/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepositoryConfiguration.java
@@ -53,7 +53,7 @@ public abstract class AbstractProxyRepositoryConfiguration
 
     public boolean isFileTypeValidation()
     {
-        return Boolean.valueOf( getNodeValue( getRootNode(), FILE_TYPE_VALIDATION, "false" ) );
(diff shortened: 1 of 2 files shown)


cstamas:
And how about

``` java
    public boolean isFileTypeValidation()
    {
+        return Boolean.valueOf( getNodeValue( getRootNode(), FILE_TYPE_VALIDATION, "true" ) );
-        return Boolean.valueOf( getNodeValue( getRootNode(), FILE_TYPE_VALIDATION, "false" ) );
    }
```

in AbstractProxyRepositoryConfiguration that renders all these changes (and those in nexus-enterprise, and also those in P2 and OBR plugins) unneeded. Also, this is more correct, since this change makes sure UTs will receive same defaults as ITs are, while with changing templates only (and nexus.xml) the change is visible in ITs only....

cstamas:
-1 pull back and redo with change above please

velo:
Done

nabcos:
Do we still need all those configuration/template changes when the default value is set correctly?

cstamas:
No, all we need is f8b851261ef8ff19cb8488514f6250197f1e14cf and nothing more. No change in Nexus EE or OBR or P2 plugin is needed either.

velo:
Ok, I closed nexus ee, obr and p2 pull request.

cstamas:
☺

velo:
BTW, added a missing conversion at rest api

velo:
I assume tamas smile face counts as +1 =D

PR #56, opened 2011-11-10
Bump Jetty to latest 7.x

cstamas:
A lot of fixes in place since we bumped to then latest 7.4.5....


velo:
+1

nabcos:
+1

peterlynch:
+1
Probably not 'required' but what about updating https://github.com/sonatype/sisu-jetty7/blob/develop/pom.xml

cstamas:
Bumped sisu-jetty7

PR #60, opened 2011-11-11
NEXUS-4628: added "resolution" to lastRequested update

cstamas:
The initial implementation only for review.

Added UTs too. Ready for votes.


velo:
too late, 57 already merged ;)

velo:
+1
but I would appreciate less code format changes  ☻

cstamas:
Rebase happened against current master and UTs added.

nabcos:
+1

bdemers:
+1 but don't forget the performance tests

bdemers:
Also the issue says 12 hours.  I'm not saying 24 is bad, we just need to agree, then update one or the other.

cstamas:
True, issue said 12h. Fixing the default value in a moment and will merge this.

cstamas:
Ah, backing up: while the issue _does mention_ 12h as "resolution", the "Evict unused proxied items" task on UI allows configurations in _days_ (24h). So, IMO, it does not make sense to have lastRequest have more grained "resolution" then a day...

bdemers:
One could make an argument that it would make a [minor] difference.  Personally I don't care, just as long as the jira issue is updated.

PR #61, opened 2011-11-11
NEXUS-4205 - Convert line endings to LF and fix checksums

velo:


nabcos:
+1, builds ok here

cstamas:
+1 Did not build, but the hashes Velo recreated match those in here 

https://gist.github.com/5a8c94601e70f763407d

So build probably passes on me too.

PR #62, opened 2011-11-11
nexus-oss-webapp-tattletale: Tattletale config fix 

cstamas:
This change only _fixes_ the folder that Tattletale is about to check.

Unsure how long, but the plugin config is wrong. This branch needs more work, since once this change is merged in, the build _will fail_.

Fix incoming, just a heads up that Tattletale does NOOP for some time...


velo:
+1

cstamas:
Hunting season open, shoot people!

PR #65, opened 2011-11-14
Fix broken artifacts

velo:
Extra bunch of invalid artifact caught by file content validation


nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3709/Nexus3709FileTypeValidationIT.java
```diff
@@ -53,7 +53,7 @@ public class Nexus3709FileTypeValidationIT
         RepositoryProxyResource resource =
             (RepositoryProxyResource) repoUtil.getRepository( this.getTestRepositoryId() );
         // this should be false to start with
-        Assert.assertFalse( resource.isFileTypeValidation(), "Expected fileTypeValidation to be false after startup." );
+        Assert.assertTrue( resource.isFileTypeValidation(), "Expected fileTypeValidation to be false after startup." );
         resource.setFileTypeValidation( true );
```
cstamas: Is this line needed?

Also, learned from previous builds.... (when assertFalse was failing since core had already your change turning it on)... instead of Assert, why not throw some exception? With assertion, there was _no trace of reason_ why build failed... moreover, all the rest of the ITs were skipped...

nexus/nexus-test-harness/nexus-test-harness-its/resources/nexus1329/files/basic/nexus1329/sample/1.0.0/sample-1.0.0.xml
```diff
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bdemers: I don't think we can depend on all xml documents having the correct header, it is pretty common for people to _forget_ this

Also, validating XML is not going to help if the user is redirected to an XHTML page

velo: Then we would need to chat about this.  What would be the plan for XML or even if there is a plan.

Tomorrow? Before scrum?

bdemers: This is why we didn't validate XML files originally.  We need to be extremely careful about this because we are upgrading existing installs.  We should err on the letting the trash though side.

velo: XML validation is there since the initial commit...
I think we should talk before scrum.

cstamas:
Yes, I believe what velo did is "just turning validation ON". The validation algorithm itself is unchanged, and it was requiring the XML headers was _always_ required by it, since it's introduction....

velo:
Ok, I rolled back XML changes since tamas is gonna deal with then.

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3709/Nexus3709FileTypeValidationIT.java
velo: Ok, fixed

cstamas:
☀

nabcos:
+1, there is nearly no change in here anymore ;)

PR #67, opened 2011-11-15
[NEXUS-4579] Option for snapshot removal to delete immediately or move to trash

nabcos:


velo:
+1

adreghiciu:
+1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/Repository.java
```diff
@@ -512,6 +513,10 @@ public interface Repository
         throws UnsupportedStorageOperationException, IllegalOperationException, ItemNotFoundException, StorageException;
 
     @Deprecated
+    void deleteItem( boolean fromTask, ResourceStoreRequest request, DeleteOperation operation )
```
cstamas, line 516: Why do we need a new delete() method exposed?

nabcos, line 516: because this needs to be set by "request" (default behavior trash, snapshot remover depending on task configuration delete or move to trash), and before this change you could only set the delete operation per (singleton) instance of the wastebasket IIRC.

nabcos:
Redoing parts of this per discussion with @cstamas, introduced delete method should be removed and switch to use ResourceStoreRequest context controlling the DeleteOperation.

PR #69, opened 2011-11-16
Mime support fixes

cstamas:
Improving MIME support in Nexus Core.
- deprecated the MimeUtil component, it was just too "bozo"
- introduced MimeSupport component, all the core uses of MimeUtil redirected to this one
- introduced MimeRulesSource, making possible to set downstream MIME types per-repository (ie. *.pom should be XML in Maven repository, but might be some binary in some other repository type)
- added validation support for BZIP2 files
- improved content and file type validation logging
- cleaned up littered MIME related configurations and classes
# How it works

In general, we had two issues:
- how Nexus reports MIME type downstream
- how Nexus validates MIME type (from upstream, in case of proxy repositories)

Originally, MIME type was just "relayed", meaning, if some broken remote repository reported POMs as "foo/bar", we sent that downstream too. Many issues were reported and related to this (see NEXUS-3972 for example).

In Nexus 1.4 we added "proper MIME support" (NEXUS-2323), but it was relatively poor: it allowed us to _globally_ set the downstream MIME types (ie. unifying what Nexus claims about POM and JAR MIME types), but this solution still prevented us to have per-repository MIME type tweaks if needed (and conflicting mime types will be present soon, with introduction of Atom support for example).

MimeRulesSource is here to make a repository able to express _how it wants to represent MIME types for known repository type specific content downstream_. It does not take part in validation. Using it, a repository is now able to "fine tune" the reported MIME types overriding globally configured values (ie. some repo might for some reason report "*.xml" files as application/atom+xml instead of application/xml as today would be reported).

As for "upstream" validation, support has been broadened (adding BZIP2 to Velo's work), but also original MavenFileTypeValidator refactored into more reusable components (as of today, we only support file type validation for Maven repository, maybe some other repo type want to use this feature too?). Now we have reusable MIME magic type detection based validator, but also support for "XML like files" detection (that was also part of MavenFileTypeValidator to express expectancies against POM and Maven Repository Metadata files for example).

And finally, there is a "tweak" settings to enable/disable XML "lax validation" as we agreed. By default, it's on, and it will never report XML file as INVALID (it will stay put saying "neutral", or VALID if XML detected as proper XML file), unless turned OFF. Aside of "tweak" setting (ie. some user might want to _enforce_ real XML validation, considering XML files without XML preamble INVALID), there is a possibility to enable/disable XML "lax validation" per-validation request.

Locally, the build did pass (master was still failing, since some IT had XML resources that were actually "fake" XML files, and since validation was turned ON by default). This change makes the build pass, since XML "lax validation" is enabled by default, and those "fake" XMLs in ITs are not considered "fake" anymore by file type validation (this is not affecting the actual assertions of ITs).


cstamas:
- wrt @Since... do we have those tags at all in Nexus?
- wrt assertion vs ISE: this is a method in IT annotated with @Before... and it seems TestNG speciality is: if Assert fails not in @Test but @Before method, simply skips this and all the remaining tests... without leaving any clue why it failed... Btw, Velo removed that line, hence, it will be removed too once this pull req merged.
- wrt MavenFileTypeValidatorTest will fix it

cstamas:
@peterlynch: wrt Javadoc, please review changes in general (using Diff tab above) not commit by commit, since Javadocs you marked as missing were added in latter commit. So, we just generate noise...

peterlynch:
I believe i was doing that... At least was my intent
On Nov 16, 2011 3:08 PM, "Tamas Cservenak" <
reply@reply.github.com>
wrote:

cstamas:
Will close this pull, rebase the branch against current master (to get rid of misleading IT fixes) and create new pull.

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus3709/Nexus3709FileTypeValidationIT.java
```diff
@@ -53,7 +53,10 @@ public class Nexus3709FileTypeValidationIT
         RepositoryProxyResource resource =
             (RepositoryProxyResource) repoUtil.getRepository( this.getTestRepositoryId() );
         // this should be false to start with
-        Assert.assertFalse( resource.isFileTypeValidation(), "Expected fileTypeValidation to be false after startup." );
```
velo, line 56: I would suggest merging this with master, I removed this entire section since it is no longer required and any problems on @Before\* will hold all other tests (which I think is bad)

velo:
And I was about to +1 this one ¬¬

PR #70, opened 2011-11-16
[NEXUS-4595] Add user filter expression for LDAP realm

nabcos:
Main patch contributed by @alecharp.

### DIFF
--- a/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/pom.xml
+++ b/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/pom.xml
@@ -173,7 +173,7 @@
               <goal>xpp3-writer</goal>
             </goals>
             <configuration>
-              <version>1.0.0</version>
+              <version>1.0.1</version>
               <models>
                 <model>src/main/mdo/ldap.xml</model>
               </models>
--- a/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/test/java/org/sonatype/security/ldap/dao/LdapUserDAOTest.java
+++ b/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/test/java/org/sonatype/security/ldap/dao/LdapUserDAOTest.java
@@ -85,6 +85,18 @@ public class LdapUserDAOTest
             // good
         }
 
+        configuration.setLdapFilter( "description=nexus" );
+        // must succeed because cstamas has the attribute description set to nexus
+        user = lum.getUser( "cstamas", initialContext, configuration);
+        try {
+            // must fail because of the ldapFilter that jvanzyl user don't have
+            user = lum.getUser( "jvanzyl", initialContext, configuration);
+            fail();
+        }
+        catch ( NoSuchLdapUserException e)
+        {
+            // good
+        }
     }
 
 }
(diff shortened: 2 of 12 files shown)


nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/pom.xml
```diff
@@ -173,7 +173,7 @@
               <goal>xpp3-writer</goal>
             </goals>
             <configuration>
-              <version>1.0.0</version>
+              <version>1.0.1</version>
```
velo, line 176: We just stop generating 1.0.0??

Shouldn't we generate both and migrate cfg from 1.0.0 to 1.0.1?

nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/test/java/org/sonatype/security/ldap/dao/LdapUserDAOTest.java
```diff
@@ -85,6 +85,18 @@ public class LdapUserDAOTest
...
+        user = lum.getUser( "cstamas", initialContext, configuration);
+        try {
+            // must fail because of the ldapFilter that jvanzyl user don't have
+            user = lum.getUser( "jvanzyl", initialContext, configuration);
+            fail();
+        }
+        catch ( NoSuchLdapUserException e)
+        {
+            // good
+        }
```
peterlynch, line 99: What type of exception if filter is malformed? Hopefully that is clearly reported in logs. Where is the test for a malformed query filter?

nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/pom.xml
nabcos, line 176: 1.0.1 is a superset of 1.0.0, so 1.0.1 can parse old config and use the default value (no filter).
 There is no version field in the ldap.xml, and no upgrade infrastructure in place for the OSS LDAP plugin.
I do not think that it's necessary to introduce all that just to insert a version number into the XML on upgrade.

velo, line 176: then do we need to change version at all?

nabcos, line 176: no. it is more clear from a development perspective though.

nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/test/java/org/sonatype/security/ldap/dao/LdapUserDAOTest.java
nabcos, line 99: Same message that you get when you input garbage for the other fields. As all fields are concatenated to one search query, it cannot be overly specific. Last part of the message is the hint. (Field value for the filter was only 'description' here (missing '=' sign)).

https://issues.sonatype.org/secure/attachment/50041/nexus-4595-malformed-filter.png

peterlynch:
+1

cstamas:
+1

velo:
+1

PR #72, opened 2011-11-16
[NEXUS-4579] Option for snapshot removal to delete immediately or move to trash

nabcos:
Reimplemented by using ResourceRequest context instead of exposing another method on interfaces.


velo:
Nice to have a IT for this
+1

peterlynch:
can you add a screenshot of new task config to issue?
https://issues.sonatype.org/browse/NEXUS-4579

cstamas:
+1

And next step would be to "move" the "default operation" from wastebasket into Repository (and have sensible defaults).

nexus/nexus-app/src/main/java/org/sonatype/nexus/maven/tasks/SnapshotRemovalRequest.java
```diff
@@ -35,8 +35,11 @@ public class SnapshotRemovalRequest
     
     private final Set<String> processedRepos;
 
+    private boolean deleteImmediately;
+
     public SnapshotRemovalRequest( String repositoryId, int minCountOfSnapshotsToKeep,
-        int removeSnapshotsOlderThanDays, boolean removeIfReleaseExists )
+                                   int removeSnapshotsOlderThanDays, boolean removeIfReleaseExists,
+                                   final boolean deleteImmediately )
     {
```
bdemers: This is a public API change.  To keep the API from changing, we could just create a new constructor.

nexus/nexus-app/src/test/java/org/sonatype/nexus/maven/tasks/DefaultSnapshotRemoverLRTest.java
```diff
@@ -88,7 +87,7 @@ public class DefaultSnapshotRemoverLRTest
         itemFile.setLastModified( tenDaysAgo );
 
         SnapshotRemovalRequest snapshotRemovalRequest =
- new SnapshotRemovalRequest( snapshots.getId(), 1, 10, true );
+ new SnapshotRemovalRequest( snapshots.getId(), 1, 10, true, true );
```
bdemers: This test changes what is under test.  I think this is fine _if_ the trash path is covered by other tests

nexus/nexus-app/src/main/java/org/sonatype/nexus/maven/tasks/SnapshotRemovalRequest.java
cstamas: Since when is SnapshotRemovalRequest part of _public API_???

I dont see any @ExtensionPoint annotation on it... it's clearly internal.

nexus/nexus-app/src/test/java/org/sonatype/nexus/maven/tasks/DefaultSnapshotRemoverLRTest.java
nabcos: right, I had the option the other way 'round initially.

nabcos:
Added 'compatibiilty' constructor and undid test changes.

PR #73, opened 2011-11-16
[NEXUS-4629] Nexus977GroupOfGroupsDownloadIndexesTaskIT

adreghiciu:
It looks like we made wrong assumptions on when a task is completed as the test presumes that the task was already started

### DIFF
--- a/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
+++ b/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
@@ -148,15 +148,7 @@ public class TaskHelperPlexusResource
 
     private boolean isTaskCompleted( ScheduledTask<?> task )
     {
-        // always have to wait for internal tasks
-        if ( RunNowSchedule.class.isAssignableFrom( task.getClass() ) )
-        {
-            return false;
-        }
-
-        // otherwise, just check the status
         return !task.getTaskState().isActiveOrSubmitted();
-
     }
 
     private ScheduledTask<?> getTaskByName( String name )
--- a/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
+++ b/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
@@ -155,12 +155,8 @@ public class TaskHelperPlexusResource
         }
 
         // otherwise, just check the status
-        if ( TaskState.RUNNING.equals( task.getTaskState() ) || TaskState.SLEEPING.equals( task.getTaskState() ) )
-        {
-            return false;
-        }
+        return !task.getTaskState().isActiveOrSubmitted();
 
-        return true;
     }
 
     private ScheduledTask<?> getTaskByName( String name )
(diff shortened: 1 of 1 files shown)


velo:
+2

nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
```diff
@@ -148,19 +148,7 @@ public class TaskHelperPlexusResource
...
-        }
-
-        // otherwise, just check the status
-        if ( TaskState.RUNNING.equals( task.getTaskState() ) || TaskState.SLEEPING.equals( task.getTaskState() ) )
-        {
-            return false;
-        }
-
-        return true;
+        return !task.getTaskState().isActiveOrSubmitted();
```
peterlynch: so if task is SLEEPING it is completed? doesn't this hide task is sleeping and not yet run?

nabcos: the method name is misleading, it translates to:

```
        return this.equals( SUBMITTED ) || this.equals( RUNNING ) || this.equals( SLEEPING ) || this.equals( WAITING ) || this.equals( CANCELLING );
```

But this is actually not going to work - IIRC SUBMITTED is every task that has an execution date set in the future (e.g. hourly/daily), so if you have such a task configured this method will never return true.

adreghiciu: Which is OK as that means that the task is not done as it was not runed. in that case the timeout will kick in and you get an exception.
Th eSUBMITTED is the initial state of the task, including an run now task.

nabcos: TaskState#isActive() should be sufficient here. although WAITING is a weird state here, task was running and waits for the next execution - this state is when a task finished running and has another execution date set (daily/hourly).
(@cstamas, can you confirm?)

So I think that boils down the states to look for to RUNNING, SLEEPING and CANCELLING again .... basically what we had before(?)

adreghiciu: @peterlynch I do not know when a task is sleeping nor waiting. I just know that a task is done only when is broken, canceled or finished

adreghiciu:
@nabcos nope, is not sufficient. when a task is created the tasks has an SUBMITTED state. until it starts running the task is still to be run. if we mke the call to isCompleted before it starts running then we get a false positive. that is in my view why we actually get the intermittent failures

nabcos:
yes, that's right. 
and a task with a schedule will get into waiting and this call will time out?
I think we should test for 'isActiveOrSubmitted && ( ! WAITING )', that might be okay for all cases

so this is basically a +1 pending further discussion or acceptance.

cstamas:
☀

peterlynch:
+1 Whatever is decided to be correct, add an explanation to the javadoc on this method so the logic is not revisited.

nabcos:
+1

nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
cstamas: I just verified that this change makes ITs behave very badly: they run now for 5-15 minutes per test!

Here is an example stack trace:
https://gist.github.com/a56275cec70465d30397

This is from my branch (rebased against master just now), but I just confirmed the same on _master branch_. Still unsure what causes this, but it's obvious that for some reason the `org.sonatype.nexus.plugins.tasks.api.TaskHelperPlexusResource.get(TaskHelperPlexusResource.java:108)` sleep is where server side request processing is stuck for much longer time than before.

I am looking more into this, but one mistake: the `SUBMITTED` state is tricky: _manually scheduled_ tasks does get back into `SUBMITTED` state and stay there until next run.

PR #76, opened 2011-11-16
Nexus 4635 nexus version

velo:
Add nexusVersion to nexus.xml


nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/events/NexusVersionChangedEvent.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.sonatype.nexus.proxy.events;
+
+/**
+ * This event is fired every first time a new nexus version is launched
+ * 
+ * @author velo
+ */
+public class NexusVersionChangedEvent
+    extends AbstractVetoableEvent<Object>
```
cstamas: a) bad named event: NexusUpgradedEvent would be better unless we plan to programmatically support "downgrading" too?
b) why is this Vetoable event? Who can put veto on this and why?

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/source/ApplicationConfigurationSource.java
```diff
@@ -48,4 +48,14 @@ public interface ApplicationConfigurationSource
     void backupConfiguration()
         throws IOException;
 
+    /**
+     * Is nexus instance upgraded
+     */
+    boolean isInstanceUpgraded();
```
cstamas, line 54: This should go to ConfigurationSource, but it's kept as hostage (by security) in base-config :(

So, for now should be okay.

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/source/ApplicationConfigurationSource.java
```diff
@@ -48,4 +48,14 @@ public interface ApplicationConfigurationSource
...
+
+    /**
+     * Setter for nexus instance upgraded
+     */
+    void setInstanceUpgraded( boolean instanceUpgraded );
```
cstamas: Is setter really needed?

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/source/FileConfigurationSource.java
```diff
@@ -201,6 +210,26 @@ public class FileConfigurationSource
...
+        if ( currentVersion.equals( previousVersion ) )
+        {
+            setInstanceUpgraded( false );
+        }
+        else
+        {
+            setInstanceUpgraded( true );
+            getConfiguration().setNexusVersion( currentVersion );
+            storeConfiguration();
+            eventMulticaster.notifyEventListeners( new NexusVersionChangedEvent( this, previousVersion,
```
cstamas: Not sure i'd go with events (yet). Model upgrade was not emitting events either, and Status flags with this change would be properly set.

Let's go with baby steps, and not introduce events for these yet (unless some other good reason mentioned for it's need)

velo: Ok, will nuke the event

cstamas:
Even better with these changes :)

bdemers:
+1

PR #77, opened 2011-11-17
[NEXUS-4616] Send times as longs and display them based using user timezone

adreghiciu:


cstamas:
+1, nice

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus2797/Nexus2797Validate140TaskUpgradeIT.java
```diff
@@ -18,12 +18,14 @@
  */
 package org.sonatype.nexus.integrationtests.nexus2797;
 
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
```
velo, line 22: I don't have strong feelings about it, but we should be using Hamcrest assertThat() over regular assertions...

velo:
+1

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus2797/Nexus2797Validate140TaskUpgradeIT.java
adreghiciu, line 22: Right. Yet I did not interfere with test. Just adapted.

PR #78, opened 2011-11-17
[NEXUS-4617] Use user time zone to display the start/recurring time

adreghiciu:
Use the start date, that includes time, to determine the start / recurring time instead of using the value sent by server


cstamas:
+1

velo:
+1

PR #80, opened 2011-11-17
Fix for pull request #73: properly handle task states for "is done" detection

cstamas:
Fix for pull https://github.com/sonatype/nexus/pull/73 properly handle task states for "is done" detection

"is done" in this scope means "did run at least once and is not running now".

What happened with previous code, is that it  never returned true, hence the
waitForAllTasks was simply giving up after sleeping the given amount of time.
This was why ITs were running LONG.

### DIFF
--- a/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
+++ b/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/nexus/plugins/tasks/api/TaskHelperPlexusResource.java
@@ -38,6 +38,7 @@ import org.sonatype.plexus.rest.resource.PathProtectionDescriptor;
 import org.sonatype.plexus.rest.resource.PlexusResource;
 import org.sonatype.scheduling.ScheduledTask;
 import org.sonatype.scheduling.TaskState;
+import org.sonatype.scheduling.schedules.ManualRunSchedule;
 import org.sonatype.scheduling.schedules.RunNowSchedule;
 
 @Component( role = PlexusResource.class, hint = "TaskHelperResource" )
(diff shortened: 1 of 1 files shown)


peterlynch:
+1 looks sane and I like the comments

nabcos:
+1

cstamas:
Generally, the problem as I see is:
- we have two "cuckoos eggs" in tasks: "run now" and "manually" run, together with the rest of 6 schedules (once, hourly, daily, weekly, monthly, cron)
- "run now" is basically _misuse_ of "task scheduling", since they are actually plain "shoot and forget" background tasks
- "manually run" would fit somehow into rest of 6 of them

Facts:
- "run now" are usually spawned background actions (like "reindex" when a new repo is added, etc) and should not even be persisted (but they are, because they are "squeezed" into current scheduler
- "manual run"'s only similarity to the rest of 6 is that they are _user configured_ (added by user) and should be persisted. They are like "handy tools", present in the UI to be run then here and there by users (as needed).
- the rest 6 does make a consistent "whole": they are totally same (added by users, persisted to config, execute on regular basis), the only difference here is _when_ they run (better: their frequency).

Problem could be solved by:

adding low level "executor" component (managing only thread executions, not managing config, scheduling, etc), and a top of that a "scheduler" (managing persistent config, scheduling, and sending tasks to executor) component.
- "run now" would not even enter "scheduler", it should go directly to "executor" and once done, just be forgotten about (just like today but with a ton of extra work)
- "manual run" and the rest of 6 should go to "scheduler", and fixed to share same states as rest 6 of them
- their states should be cleaned up, these try types of them share different states/lifecycle projected onto _same bloated_ enum.

Etc.

velo:
+01

peterlynch:
>    "is done" in this scope means "did run at least once and is not running now"

Should have called the method `isRanAtLeastOnceAndNotRunningNow()`

bdemers:
+1 good comments

PR #83, opened 2011-11-17
NXCM-3427 - If login return 403 then automatically execute logout routine

velo:
If a user lacks "Login to UI" priv, they will essentially get a 403 from authentication/login resource when providing valid username and password.  But refreshing screen will present as logged in

To minimize risk:
If 403 received from authentication/login, keep all existing behaviour, but send a logout request to backend immediately and clear client session cookie in browser.


peterlynch:
+1

nabcos:
+1

PR #85, opened 2011-11-18
Get rid of manual mockery

cstamas:
We had few of these, just making our lives harder.

All of them were simple: getting it to return repo ID or URL or similar.


velo:
+1

nabcos:
+1

PR #87, opened 2011-11-18
NEXUS-4630, Sessions were still getting added to the cache, that is no longer the case

bdemers:


cstamas:
+1, but it'd like to see lower level tests to: like creating a session, and then _using actual EHCache instance_ for counting cached entries too.

While I am not questioning the test above (is correct as far as I see), the fact you would do assertions against EHCache.getSize() would make it more clear about the real intentions.

Naturally, with a big fat warning, that this UT tests SessionManager in a _given_ configuration only (since UT would depend on EHCache and specific back-end of session manager).

So maybe a smaller refactor, moving the "job" in abstract class and have the assertions in one UT class (these above), and have another class that does similar assertions but against EHCache -- that could be easily trashed IF we switch from EHCache to something else?

nabcos:
+1

PR #89, opened 2011-11-21
Nexus 4628 perftests

cstamas:
Based on Toby's branch, but rewritten to lessen code duplication.

Tests mostly untouched, just adapted to changes.
- merged original branch, but most of it thrown away
- reworked whole idea by introducing Marshaller (xstream, jackson x2)
- added delegating AS (to be able to perform upgrade as JD explained, "on the fly")
- adapted the four "perf" tests, but not sure what they test at all
- concentrated lastRequested "tweaks" to one single place
- other cleanups

### DIFF
--- a/nexus/nexus-proxy/pom.xml
+++ b/nexus/nexus-proxy/pom.xml
@@ -135,6 +135,14 @@
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-aether-provider</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.jackson</groupId>
+      <artifactId>jackson-mapper-asl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml</groupId>
+      <artifactId>jackson-xml-databind</artifactId>
+    </dependency>
     <!-- Sisu -->
     <dependency>
       <groupId>org.sonatype.sisu</groupId>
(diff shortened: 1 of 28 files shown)


nexus/nexus-proxy/pom.xml
```diff
@@ -135,6 +135,14 @@
...
       <artifactId>maven-aether-provider</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.jackson</groupId>
+      <artifactId>jackson-mapper-asl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml</groupId>
+      <artifactId>jackson-xml-databind</artifactId>
+    </dependency>
```
peterlynch, line 145: how are we preventing jackson from being included in webapp?

cstamas:
Closing this since there is more + will apply Alin's comments.

nexus/nexus-proxy/pom.xml
cstamas, line 145: I don't understand the question. These changes come from the branch that is made as "secondary goal" in NEXUS-4628, and according to issue, "(it) should merge attribute-performance-test branch to master".

Since that branch brings in some components _depending on_ Jackson, Jackson _is_ gonna end up in Core (since is needed by newly introduced core components).

peterlynch, line 145: to @adreghiciu point - I don't think we rely on jackson except in these alternate implementations. Perhaps the alternate implementations themselves should be in src/test/java and these dependencies scoped test

adreghiciu, line 145: Yep, that was my point also

bdemers:
This pull is already closed, but to continue the thread...
The original jira states: (The performance tests should be merged to master, the extra cruft for Jackson json/xml should be thrown out)
They were not ment to be merged as is, as we did not go down the path of replacing the attribute store.  So the tests need to be reevaluated

PR #91, opened 2011-11-22
NEXUS-4628: Perftests

cstamas:
Reworked the whole of changes from originating branch, but keeping all the "alternate" implementations under test scope for now.

Also, moved all the "lastRequested" tuning possibilities into one single place: the AttributeHandler class.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultAttributesHandler.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultAttributesHandler.java
@@ -309,11 +309,11 @@ public class DefaultAttributesHandler
         // Touch it only if this is user-originated request (request incoming over HTTP, not a plugin or "internal" one)
         // Currently, we test for IP address presence, since that makes sure it is user request (from REST API) and not
         // a request from "internals" (ie. a running task).
-        // TODO: maybe the other nexus.ls.file.touchLastRequested* props should be moved here
 
+        // we do this only for requests originating from REST API (user initiated)
         if ( request.getRequestContext().containsKey( AccessManager.REQUEST_REMOTE_ADDRESS ) )
         {
-            // if we need to do this at all...
+            // if we need to do this at all... user might turn this feature completely off
             if ( doTouchLastRequested( repository ) )
             {
                 final long diff = timestamp - storageItem.getLastRequested();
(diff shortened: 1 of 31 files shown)


cstamas:
Bah, closing it again, will reopen once tests pass with these mods

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultAttributesHandler.java
```diff
@@ -64,39 +71,72 @@ public class DefaultAttributesHandler
...
+    private static final boolean LAST_REQUEST_ATTRIBUTE_ENABLED_FOR_HOSTED = SystemPropertiesHelper.getBoolean(
+        "org.sonatype.nexus.proxy.attributes.DefaultAttributesHandler.lastRequested.enabled.hosted",
+        LAST_REQUEST_ATTRIBUTE_ENABLED );
+
+    /**
+     * Flag to completely enable/disable the lastRequested attribute maintenance for proxy repositories.
+     */
+    private static final boolean LAST_REQUEST_ATTRIBUTE_ENABLED_FOR_PROXY = SystemPropertiesHelper.getBoolean(
+        "org.sonatype.nexus.proxy.attributes.DefaultAttributesHandler.lastRequested.enabled.proxy",
+        LAST_REQUEST_ATTRIBUTE_ENABLED );
```
nabcos, line 96: The change and addition of system properties should be documented in confluence

https://docs.sonatype.com/display/Nexus/Configuration

cstamas:
Not sure what to do here, and actually completely unsure about the _value_ of these "perf tests" at all (and unsure why are we loosing time on this at all).

Here is the scoop: moving the tests and keeping test deps out of core is not enough: Toby did changes ("hack" as he noted too) on StorageItem implementations, to make them Jackson compat. Without those (I did remove them from nexus-api), Jackson is unable to cope with classes, rendering UTs that this pull req is about to add useless.

I am _against_ adding those hacks to core, only to be able to add these UTs that we actually don't even know what are testing...
- They test Xstream vs Jackson? It's done already here https://github.com/eishay/jvm-serializers/wiki
- The fact that we do not _invoke_ lastRequest update? It's done already here https://github.com/sonatype/nexus/blob/master/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/attributes/Nexus4628Test.java
- The fact that AttributeStorage perf will not downgrade in some future release? Why the Jackson implementations then?

bdemers:
I agree, I would NOT add those hacks to the core.  I don't think  there is any value in keeping the Jackson perf tests. (just keep the impl and tests we need)

PR #92, opened 2011-11-22
Passivate on nexus stop

adreghiciu:
Fix problem with passivation being called even if not activated.
Refactor to allow state of capability to be kept in registry.
Add more UTs.
Make IT just check a proper setup as UTs are already testing inner workings.

### DIFF
--- a/nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/api/CapabilityReference.java
+++ b/nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/api/CapabilityReference.java
@@ -16,27 +16,38 @@
  * Sonatype, Inc. Apache Maven is a trademark of the Apache Foundation. M2Eclipse is a trademark of the Eclipse Foundation.
  * All other trademarks are the property of their respective owners.
  */
-package org.sonatype.nexus.plugins.capabilities.internal.config.test;
+package org.sonatype.nexus.plugins.capabilities.api;
 
-import static org.mockito.Mockito.mock;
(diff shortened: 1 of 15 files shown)


nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/api/CapabilityReference.java
```diff
@@ -0,0 +1,53 @@
...
+ * http://www.gnu.org/licenses.
+ *
+ * Sonatype Nexus (TM) Open Source Version is available from Sonatype, Inc. Sonatype and Sonatype Nexus are trademarks of
+ * Sonatype, Inc. Apache Maven is a trademark of the Apache Foundation. M2Eclipse is a trademark of the Eclipse Foundation.
+ * All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.plugins.capabilities.api;
+
+/**
+ * Reference to a capability and its state.
```
bdemers, line 22: Can you add more to this doc.

nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/api/CapabilityReference.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @since 1.10.0
```
cstamas, line 24: Forgot to add `@author cstamas` ?

cstamas:
+1

jdillon:
+1 if this fixes the passivation on shutdown for capability which != enabled

PR #94, opened 2011-11-22
o Let nexus log at debug level with -Pdebug in ITs

nabcos:
The debug profile sets the root level of the nexus logger, but org.sonatype.nexus was still set to INFO.


velo:
+1

cstamas:
Should't the file be named `logback-test.xml`?

But in general, since this is IT, and it replaces the logback config coming from the bundle... +1

peterlynch:
+1 so magic in nexus-test-env-maven-plugin transforms this into logback.xml in nexus?

nabcos:
current logback.xml found in work/nexus/conf directory looks like this: 

```
<?xml version='1.0' encoding='UTF-8'?> 
<configuration scan='true'>
    <property file='${nexus.log-config-dir}/logback.properties'/>
    <include file='${nexus.log-config-dir}/logback-events.xml'/>
    <include file='${nexus.log-config-dir}/logback-default.xml'/>
</configuration>
```

PR #95, opened 2011-11-22
o Fix DefaultNotificationManager NPE when unconfigured

nabcos:
While debugging Nexus4593*IT, NPEs from DefaultNotificationManager#isEnabled cluttered the logs, caused by unset section in nexus configuration.

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/notification/DefaultNotificationManager.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/notification/DefaultNotificationManager.java
@@ -73,6 +73,12 @@ public class DefaultNotificationManager
     }
 
     @Override
+    protected boolean isConfigured()
+    {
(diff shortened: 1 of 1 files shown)


velo:
+1

cstamas:
Isn't this problem applicable to all components extending AbstractConfigurable? Why not address it there?

Also just look at the super.isConfigured() implementation... this looks like some concurrency problem...

nabcos:
why concurrency? as I understood it that NPE could come from a config like `<config/>`, where the notification manager expects `<config><$notificationName/></config>`. 

Maybe it should also be addressed in the abstract, but I don't know about side-effects of treating `<config/>` as unconfigured, are the Configurables expecting some behavior regarding missing config sections?

cstamas:
Ah, true. Sorry for the noise.

peterlynch:
+1

PR #96, opened 2011-11-23
Capabilities activation

adreghiciu:
The basic idea is that a capability should be able to express activation conditions. As in for example, a capability can be activated only when another capability is activated, or when a repository of a certain type exists. basically the implementation allows any types of activation conditions but for now we just implemented the activation when another capability is active.


cstamas:
:+1:

nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/internal/DefaultCapabilityRegistry.java
```diff
@@ -19,72 +19,251 @@
...
 import org.sonatype.nexus.plugins.capabilities.api.Capability;
 import org.sonatype.nexus.plugins.capabilities.api.CapabilityFactory;
 import org.sonatype.nexus.plugins.capabilities.api.CapabilityReference;
 import org.sonatype.nexus.plugins.capabilities.api.CapabilityRegistry;
+import org.sonatype.nexus.plugins.capabilities.api.activation.ActivationContext;
 
 /**
  * Default {@link CapabilityRegistry} implementation.
+ * <p/>
+ * TODO add locking
```
nabcos: is this TODO still valid? there are locks used in the code

nabcos:
+1

nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/plugins/capabilities/internal/DefaultCapabilityRegistry.java
adreghiciu: Nope. Removed.

PR #101, opened 2011-11-24
Nexus 4611 close stage repo

velo:
Although I do prefer this code to still handle Stream and not String, the stream error message do not give me much details:
Caused by: org.jdom.input.JDOMParseException: Error on line -1: Premature end of file.

So I changed it to String in hope to get more information on what is going wrong here.


nabcos:
+1

cstamas:
:+1: but still unsure what about the related issue is and where the problem is... i guess this is only to improve error reporting?

velo:
@cstamas
exactly. I still don't know what the problem is.  Last time I found we were swallowing an exception, so I fix that.
Now, I have the hole exception.  But, it doesn't give me much information, so I made it more verbose.

peterlynch:
+1

bdemers:
+1, 
Though I think the test failure is caused by the response body being an empty string.
Either way, this change will help debug

PR #104, opened 2011-11-25
Support filtering of repositories by user-agent

Vlatombe:
In some use cases I have, I may have concurrent deployment of 2 versions of the same g:a.
For instance, 1.1 and 2.0 can be staged at the same time.

To be sure that 2 different staging repositories get allocated, I use a specific user-agent for deployment that is unique to each build.

This change allows to filter all operations of the nexus-maven-plugin using a user-agent.

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
+++ b/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
@@ -43,7 +43,14 @@ public abstract class AbstractStagingMojo
 
     private StageClient client;
 
-    public AbstractStagingMojo()
+    /**
+     * Filter opened staging repositories using the provided user agent
+     *
+     * @parameter expression="${nexus.userAgent}"
+     */
+    private String userAgent;
+
+		public AbstractStagingMojo()
     {
         super();
     }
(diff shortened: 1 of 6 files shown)


cstamas:
Looks nice, :+1:

nabcos:
Nice, but needs some tests before going to master

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -249,4 +264,24 @@ public abstract class AbstractStagingMojo
         this.repositoryId = repositoryId;
     }
 
+    protected List<StageRepository> filterUserAgent(List<StageRepository> repos)
```
peterlynch, line 267: javadoc

peterlynch:
tests + plus I think Rich and Alin need to look this over.

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -43,7 +43,14 @@ public abstract class AbstractStagingMojo
 
     private StageClient client;
 
-    public AbstractStagingMojo()
+    /**
+     * Filter opened staging repositories using the provided user agent
```
peterlynch, line 47: I m not sure this comment by itself would provide enough information on when and how to use this filter

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/ListStageRepositoriesMojo.java
```diff
@@ -56,7 +56,7 @@ public class ListStageRepositoriesMojo
         if ( repos != null )
         {
             StringBuilder builder = new StringBuilder();
-            builder.append( "The following OPEN staging repositories were found: " );
+            builder.append( String.format( "The following OPEN staging repositories were found for user-agent: '%s'", getUserAgent() ) );
```
bdemers, line 59: I like this, but this message may be confusing when user-agent == null

bdemers:
@Vlatombe do you typically set the user-agent manually each deploy? Or are you always setting the user-agent to include the version ? ( just curious )

Vlatombe:
In our builds, we set the user-agent to %BUILD_TAG% which provides us the Jenkins build ID, that way it is easy to track down by which job a staging repository has been opened.

adreghiciu:
+1 on change. Just add unit tests

cstamas:
Closing this pull req as new one replaces this:
https://github.com/sonatype/nexus/pull/371

PR #105, opened 2011-11-28
[NEXUS-4657] Exclude abstract test and IT classes

nabcos:
Corresponding build:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/73/jdk=java-6x,label=linux/consoleText

### DIFF
--- a/nexus/pom.xml
+++ b/nexus/pom.xml
@@ -1543,7 +1543,7 @@
               <it.test.log.level>${it.test.log.level}</it.test.log.level>
               <it.nexus.log.level>${it.nexus.log.level}</it.nexus.log.level>
             </systemPropertyVariables>
-            <excludes>
+            <excludes combine.children="append">
               <exclude>**/Abstract*</exclude>
             </excludes>
           </configuration>
@@ -1565,7 +1565,7 @@
             </execution>
           </executions>
           <configuration>
-            <excludes>
+            <excludes combine.children="append">
               <exclude>**/Abstract*</exclude>
             </excludes>
             <properties>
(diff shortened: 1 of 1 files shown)


velo:
I always though the test runner automatically remove abstract classes... abstract like public abstract class...
+1

peterlynch:
-1
excludes are not additive - while your changes may work, they are not exhaustive.

Basically whereever there is surefire or failsafe plugin config specifying excludes, we need to exclude Abstract pattern.

Example: https://github.com/sonatype/nexus/blob/master/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/pom.xml#L157

nexus/pom.xml
```diff
@@ -1543,6 +1543,9 @@
               <it.test.log.level>${it.test.log.level}</it.test.log.level>
               <it.nexus.log.level>${it.nexus.log.level}</it.nexus.log.level>
             </systemPropertyVariables>
+            <excludes combine.children="append">
+              <exclude>**/Abstract*</exclude>
```
peterlynch, line 1547: well this is magical but seems harmless enough.

+1

can a child module override "combine.children" with a value that says " don't combine with parents"?

bdemers, line 1547: I'll always assumed that this is a junit/surefire bug, why would you run an abstract test class?

+1

peterlynch, line 1547: To answer my own question 

```
combine.self="override"
```

ignores the parent values.

See http://www.sonatype.com/people/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/

PR #108, opened 2011-11-28
ErrorReporting IT maintenance

nabcos:
Spotted this when looking into failure on grid:
- connections not released properly
- incorrectly ignored test methods
- one test method still disabled, does not work probably because of missing user in JIRA

I added NexusResponseMatchers#respondsWithStatus as a shorthand for using restlets status constants directly.

Reference build will be https://builds.sonatype.org/job/nexus-oss-its-feature/80


velo:
+1

cstamas:
:beer:

nabcos:
unrelated test failure made the build skip ITs
new try: https://builds.sonatype.org/job/nexus-oss-its-feature/83

PR #109, opened 2011-11-28
Fixing broken tests on windows (I think: https://builds.sonatype.org/job/nexus-oss-feature/jdk=java-6x,label=win/6/testReport/org.sonatype.nexus.proxy.storage.local.fs.perf/ )

bdemers:


bdemers:
I didn't mean 5867ef7 to be in this pull request, If interested, just look at 4f500cd

PR #111, opened 2011-11-29
Nexus 4611 close stage repo

velo:
Wanna expose the body and statusText to see if it reveals why this test fail sometimes.


velo:
CI votes blue 
https://builds.sonatype.org/job/nexus-oss-feature/12/

peterlynch:
+1
not that I think it has something to do with 4611 directly but it appears that method.releaseConnection may not get called in some rare case exception gets thrown outside try/finally

nabcos:
+1

PR #114, opened 2011-11-30
Event bus

adreghiciu:
Replace internal pub-subscribe with guava event bus.
We should consider getting rid of out messaging implementation in nexus and use the new NexusEventBus.


nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/eventbus/internal/DefaultNexusEventBus.java
```diff
@@ -0,0 +1,111 @@
...
+import org.sonatype.nexus.eventbus.NexusEventBus;
+import org.sonatype.nexus.eventbus.internal.guava.EventBus;
+import org.sonatype.nexus.eventbus.internal.guava.EventHandler;
+import org.sonatype.nexus.logging.AbstractLoggingComponent;
+
+/**
+ * {@link NexusEventBus} implementation using guava event bus.
+ *
+ * It differs from guava event bus by dispatching events as they appear (is re-entrant). Guava will queue up all event
+ * and dispatch them in the order they were posted, without re-entrance.
```
jdillon, line 36: I'm curious why this modification was needed.  Can you explain?

jdillon:
Generally -1 on copying the Guava EventBus code into the capabilities plugin.

cstamas:
Why is copying needed? Nexus Core uses Guava 10.0.1 which has the event bus...

jdillon:
I'm okay to remove my -1 on the pull, as long as there is a clear path to try and get the changes back to guava to avoid needing to fork these classes (assuming that its really a simple matter of access modifiers)

adreghiciu:
Feature build passed: https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/89/
The windows one failed to start (same as some other builds before so we need to check what is going one)

adreghiciu:
The copying is needed because guava classes are package protected to extension. All of this is explained in a README inside guava package

nexus/nexus-core-plugins/nexus-capabilities-plugin/src/main/java/org/sonatype/nexus/eventbus/internal/guava/README
```diff
@@ -0,0 +1,6 @@
+The classes in guava package are copied over from com.google.common.eventbus 1.10.1 in order to make then extensible.
+The annotations are not copied so handlers will not have to change if we later are able to drop this classes and
+extend the original.
+
+The changes done are:
+    * make some classes protected/public
```
bdemers, line 6: Maybe consider moving this to a package javadoc?  (though maybe people will actual read this if left as README )

jdillon:
I'm still concerned about the forked guava stuff in here, but I need that bug fixed so I'm +1

PR #116, opened 2011-11-30
Move Timeline and Feeds into a plugin

cstamas:
This change moves the "Spice Timeline" (this one here: https://github.com/sonatype/spice-timeline) and feeds based on it from core into a plugin together with its dependencies.

What does it buys for us? 

First, with this change, it is easy-peasy to remove completely the timeline plugin, and your Nexus will work just fine -- it's nice if you are not interested in feeds (you loose them without timeline), but you are interested in more performance.

Second, as spice-timeline was in core, it's dependencies was in core too, like Google Protobuf and Lucene. And this took us into impossible situation. For example: as Maven Indexer become independent project, we are tied and kept as "hostage" of it's Lucene version, and whenever it updates Lucene we have to bump Lucene version used in spice-timeline too to keep them aligned. Moreover, this makes Maven Indexer totally non-extensible (by adding IndexCreators in another plugin): the indexer plugin has it's own Lucene JARs, the core has another (same versioned) Lucene JARs and there will be linkage exceptions. If plugin includes it's Lucene, it still ends up with Linking errors. Having this situation, we were is check-mate position with extending anything that would use Lucene classes in API (Maven Indexer API has Lucene classes in it, that's another story).

Same of these stands for Google Protobuf (ie. if we had the ActiveMQ plugin that would for some reason wanted to share some protobuf generated classes, it's a no go).

Lucene and Protobuf are -- this is my own opinion -- too much "key technologies" to remain "pinned" by core. Core should not have any of those two.

Now, following deps are moved out from core into this new plugin (core has no more these in classpath):
- org.apache.lucene:lucene-core
- com.google.protobuf:protobuf-java
- jdom:jdom
- rome:rome

How it works?

All the direct couplings (in security, tasks and item content validation) are replaced with events. Simply, timeline plugin now carries a bunch of EventInspector implementations that listen to some events and create Timeline records (mostly as it was before, most of these classes are just moved). Also, the feed sources are moved here too.

One todo remains: modify the Feed UI to be more "smarter" with the lack of feed list resource. Since AFAIK the feed UI is generic ExtJS RSS component, and having all the timeline+feeds in this plugin, it allows us to ship feeds with completely different technology and backing source in future. So, IMO, leaving Feeds UI in core is fine as far:
- it would warn in case 404 is responded for it's "list feeds" request with message like "There are no feeds present on this Nexus instance" or alike
- if I am right, and the Feeds UI is a generic RSS reader


cstamas:
Just one remark: while two interfaces, the `NexusTimeline` and `FeedRecorder` did change, they are internal components, not marked with `@ExtensionPoint` and not directly used by (normal) plugins nor other core components.

As a proof, just check how none of the "core" feed implementations (in nexus-rest-api module) did _change_. Meaning, if a 3rd party plugin did implement a new feed, it's unlikely it will need a change.

nexus/nexus-core-plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/timeline/TimelineRecordWrapper.java
```diff
@@ -0,0 +1,48 @@
+package org.sonatype.nexus.timeline;
```
adreghiciu, line 19: Missing copyright (= build fails)

adreghiciu:
+2
I would even go further and move the whole timeline related things into a plugin as I think that no other core functionality is related to timeline

cstamas:
Yes, I started doing that but backed out, since I found two jungles, um, swamps better, from Jurassic period.

From Nexus Core (I use this term for everything except REST API and UI) perspective, all is cleanly separated, and the move is easily doable. We have few RSS feeds in core, moving them would be easy. They are decoupled from everything else, since eventing is used to capture feed entries.

But first swamp I stepped on was UI: the feed UI is part of the monolithic "main" UI.... And second swamp was security... And the I gave up.

I really, really really hate our UI and security...

They are just high walls needed to "jump over", and simply keeps everything in status-quo for too long (man just looses will to do something when he notices the hoops that has to be done).

cstamas:
Scheduled CI:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/90/

velo:
+1

bdemers:
-1
I'm not against the change, only that if this plugin fails to load (or is removed) the feeds will look like they should work.  I think if we go down this path the feed UI needs to be part of this plugin (before we merge to master)

cstamas:
CI voted -1, it seems some IT hold direct class reference to spice-timeline classes. Fix follows.

https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/90/

cstamas:
Moved out feeds too.

Now what core has is a "thin" delegating resource, that is able to detect presence of timeline plugin.

If present, the timeline based feeds are used and enlisted, otherwise UI simply reports _empty_ list of avaiable feeds.

Also, Core lost two more deps: rome and jdom, they are also now in timeline plugin.

cstamas:
Scheduled new CI job after these changes:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/93/

cstamas:
Moved feeds too, and core lost 2 more dependencies (in total 4 of them, pull comment updated, reread it please).

Scheduled job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/94/

I hope it will not be lost as others were...

bdemers:
An empty feed UI is going to be confusing.  I think these UI bits need to be part of the timeline plugin.  If not then hidden (if a feed can be added without the timeline plugin).  Or a simple case, show an error on the feed UI, telling the user how to resolve the problem.

cstamas:
Rebased against latest master (there was some minor conflicts).

nabcos:
Is this branch removing the whole feed UI now when the plugin is installed (as @bdemers requested)?

cstamas:
No, it does not touches UI at all, on server side all is done.

I'd prefer keeping the UI in Core (since AFAIK, it's a generic RSS reader implemented in ExtJS), and just one simple modification would be needed: if feed list returns 404 (the feed list queries to fill the grid is not present, like plugin is removed), UI should tell some "meaningful" message like "This instance of Nexus does not have feeds available".

And then (if my assumption about feed UI as RSS reader is true), nothing stops us to provide alternate feed plugins in the future, that might rely on some other technology (other from Timeline).

adreghiciu:
Make the feeds UI reader a plugin ;)

nabcos:
we can convert it to a capabilty as soon as we have the advanced UI =D

adreghiciu:
+1

cstamas:
Rebase against latest master. CI started:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/105/

cstamas:
Am closing this pull req as there is more to it.

PR #118, opened 2011-12-01
NEXUS-4668: Make Lucene shaded too (as Maven Indexer is already)

cstamas:
The nexus-indexer-lucene-plugin main artifact was already "shading" in Maven Indexer. This was needed to make actual components and API reachable to outer world.

This change does the same with Lucene Core, and allows the indexer plugin dependants to use Lucene classes,
since sadly, Maven Indexer API makes use of Lucene classes.


nabcos:
+1

velo:
+1

PR #119, opened 2011-12-01
NEXUS-4682: decouple shadows from master repositories.

cstamas:
Today, as many shadows (UI calls them "virtuals") you have against one master repository, the slow-down adds up to master repository slowdown, since all the content changes in shadows are done synchronously.

Before this change, shadow implementations were using the "low level" eventing in Nexus (plexus-app-events) that does not support async event handlers. This change simply "moves" shadow event processing into "high level" Nexus EventInspectors, and adds one new async EventInspector that replays the content change events asynchronously.

This way, the master repository performance is not affected (minimally is, but not in a way it was before: slowdown is not anymore linearly paired with number of shadows against it), or is minimally affected.

The fact that a M1 layouted artifact appears momentarily or few millis later when a M2 receives a deploy is not important at all.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/ShadowRepository.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/ShadowRepository.java
@@ -20,6 +20,7 @@ package org.sonatype.nexus.proxy.repository;
 
 import org.sonatype.nexus.plugins.RepositoryType;
 import org.sonatype.nexus.proxy.NoSuchRepositoryException;
+import org.sonatype.nexus.proxy.events.RepositoryItemEvent;
 import org.sonatype.nexus.proxy.registry.ContentClass;
 
 /**
@@ -88,4 +89,13 @@ public interface ShadowRepository
      * Triggers syncing with master repository.
      */
     void synchronizeWithMaster();
+
+    /**
+     * Performs some activity if the event is coming from it's master repository. Implementation should filter and take
+     * care what repository is the origin of the event, and simply discard event if not interested in it.
+     * 
+     * @param evt
+     * @since 1.10.0
+     */
+    void onRepositoryItemEvent( RepositoryItemEvent evt );
 }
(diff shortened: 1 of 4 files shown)


bdemers:
+1
But, verify that making these events async does not cause sporadic test failures.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/ShadowRepository.java
```diff
@@ -88,4 +89,13 @@ public interface ShadowRepository
      * Triggers syncing with master repository.
      */
     void synchronizeWithMaster();
+
+    /**
+     * Performs some activity if the event is coming from it's master repository. Implementation should filter and take
+     * care what repository is the origin of the event, and simply discard event if not interested in it.
+     * 
+     * @param evt
```
nabcos, line 97: `@param` should be dropped

nabcos:
is this thread-safe? in the async handler, you call `createLink(...)`, is this always a cheap operation? IIUC the next async event may overtake an earlier one here when create/deleteLink takes some time for an artifact.

cstamas:
Yes it is: createLink/removeLink builds upon Repository methods (it all boils down to storeItem/deleteItem), and they are threadsafe.

As for following scenario: somebody deploys then deletes immediately after it, yes, there is a chance that "deploy" comes after "delete" event. These cases were covered before (shadow will not fail master operation but will just report the error and go on). And if a link without master target is requested, it will be cleaned up as part of link resolve -- nothing of these changed.

nabcos:
+1

velo:
+1

peterlynch:
+1 but agree with @bdemers : are we sure adding this asyncness does not lead to some IT failures...

cstamas:
Since CI is queued up a bit, I did multiple local builds that passed on this change.

cstamas:
Rebased against latest master (to include the already merged Snap-Remover improvements this change relates to)

cstamas:
Am about to merge this into master to not have it rot too much, and if we detect test "instabilities", we will properly deal with those if needed, but I don't believe it should. Local builds are okay, resulting bundle too.

PR #120, opened 2011-12-02
o [NEXUS-4679] Fix validation message key for duplicate repository IDs

nabcos:
If no explicite key is used for validation errors, the REST reply will have numerical IDs that do not match field names and will not be displayed anywhere.


nabcos:
reference build will be https://builds.sonatype.org/job/nexus-oss-its-feature/96

velo:
+1

bdemers:
Can you add a test for these conditions (there should be existing similar tests)
Other then that it looks good

peterlynch:
+1 we really need javascript tests but that is a different story ;)

nabcos:
interesting. https://builds.sonatype.org/job/nexus-oss-its-feature/96 sat in queue for 3 days and started building just now, so the build will use latest test changes as well.

nabcos:
CI voted +1

PR #121, opened 2011-12-02
o [NEXUS-3292] Save LVO configuration on enable/disable

nabcos:


velo:
+1 and a extra +1 for using FileUtils =]

bdemers:
+1

peterlynch:
+1 oldie but a goodie...

cstamas:
-1

Core presently uses:
- PlexusUtils (it started with it, as it was Plexus IoC powered app)
- Commons-\* creeped in into test some time ago
- Guava being added recently (in 1.10.0)

So, we should at least strive to settle on one, and also keep in mind _migration_, but in either case, commons-\* artifacts are probably not the one we want to move to.

Facts:
- nexus-proxy declares dep commons-io, probably _by mistake_, since nothing uses (or should not use) it in there
- nothing in nexus core uses commons-io (while it was present in core classloader, see above)
- only ITs make (sadly) extensive use of commons-io (while almost all equiv methods exists in plexus-utils)

Note: see
org.codehaus.plexus.util.FileUtils.copyURLToFile(URL, File)

cstamas:
And for everyone else, I just spotted that some of the last written UTs (attribute perf tests merged by me) do use commons-io. Please take care which classes from which dependencies you use while writing UTs/ITs, since with those three overlapping libraries (not to mention jetty-utils and others present too) this is becoming messy and easy to "miss"!

nabcos:
hm, no reference build...
will be https://builds.sonatype.org/job/nexus-oss-its-feature/98

nabcos:
build passed on win, multiple IT failures on linux (like [that job in the main OSS build](https://builds.sonatype.org/job/nexus-oss-its/106/). The feature build was started around the same time and fails with a subset of the tests failing in the main build... coincidence? :(

I will treat that as "passed" and merge.

PR #123, opened 2011-12-02
Add a getter also for finding if local status of a repository changed

adreghiciu:
This is needed as the local status changed event is misleading. this because the event is fired before the change is committed so if for example I do repository.store() item in the moment that event is fired I will get an exception because even if the new status is IN_SERVICE the repository status is not actually in service (yet)


velo:
+1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/events/RepositoryConfigurationUpdatedEvent.java
```diff
@@ -56,6 +58,11 @@ public class RepositoryConfigurationUpdatedEvent
         return downloadRemoteIndexEnabled;
     }
 
+    public boolean isLocalStatusChanged()
```
velo, line 61: JDoc here wouldn't hurt...

cstamas:
:beer:

PR #124, opened 2011-12-02
Fixed a bug in the StatelessAndStatefulWebSessionManager that failed to allow non web logins and added a test for this condition.

bdemers:


velo:
+1

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/security/StatelessAndStatefulWebSessionManager.java
```diff
@@ -83,6 +84,11 @@ public class StatelessAndStatefulWebSessionManager
                 log.trace( "Session {} was created for User-Agent {}", session.getId(), getUserAgent( request ) );
             }
         }
+        else
+        {
+            log.trace( "Non http request, falling back to default implemenation." );
```
nabcos: typo

nabcos:
+1

PR #126, opened 2011-12-05
Removing Commons IO uses littered across.

cstamas:
We already have Plexus Utils, and just added Guava. There is no need
(and never was) to use IO related commons. In many classes that were changed
below, Plexus IOUtil was already imported! And many times, equiv
method with even same name was present.

So, in future please try to verify what you are using, and tend to use Plexus Utils OR Guava, but not both nor any third one.

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/timeline/DefaultNexusTimeline.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/timeline/DefaultNexusTimeline.java
@@ -24,7 +24,6 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Set;
 
-import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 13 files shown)


velo:
Shouldn't we go directly to guava (assuming there is guava-io)

velo:
+1

cstamas:
Probably yes, at least with classes like `Closeables` et al. But we should not pull in anything else aside those two: plexus utils or guava.

Sadly, as I said, commons-io is extensively used in UTs and ITs, so it would need some effort to clean those up. But with this change, Core will not contain commons-io... and this is something that _might_ break Pro plugins (OSS plugins like Migration plugin assuming commons-io is in core is fixed)

nabcos:
+1

adreghiciu:
+1

cstamas:
Locally built and validated during long phone calls :D

PR #130, opened 2011-12-07
o [NXCM-3543] pull up loadHandler into abstract repo editor UI

nabcos:
This small change fixes one UI issue:
The 'Download Remote Index' box will be set to false and disabled for every non-maven2 proxy repository.
That code was lying around but never called because the listener was not registered correctly.


cstamas:
+1

velo:
+1

adreghiciu:
+1

PR #131, opened 2011-12-07
NEXUS-4640: The Parent-O-Matic device

cstamas:
As Peter's testing against generated snapshot repository with 13000 artifacts revealed an interesting results (snapshot removal using 1.9.2.3 takes 3(!) minutes, but the "post processing" is 20 minutes, hence to complete the task you need to wait cca 23 minutes), the need for a special device emerged.

And here it is: Parent-O-Matic!

The problem with Snapshot remover was actually a lame "optimization" added in there, that seemed correct in "laboratory" circumstances (and ITs), but resulted in "anti-optimization" in real life: there was a Map that gathered which paths need to have "rebuild maven metadata" regeneration applied. On large repositories, this path ended up with all the existing paths in repository, and even OOMed Nexus. 

Finally, this change does not makes previous (already merged) pull request also related to NEXUS-4640 obsolete. While it also fixed major issues, it's fix covered so to tell just 10% of the problem, while this fix covers the rest of it, and even OOM related issue (the one linked to NEXUS-4640).

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
@@ -1,0 +1,409 @@
+/**
+ * Copyright (c) 2008-2011 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions
+ *
+ * This program is free software: you can redistribute it and/or modify it only under the terms of the GNU Affero General
+ * Public License Version 3 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License Version 3
+ * for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License Version 3 along with this program.  If not, see
+ * http://www.gnu.org/licenses.
+ *
+ * Sonatype Nexus (TM) Open Source Version is available from Sonatype, Inc. Sonatype and Sonatype Nexus are trademarks of
+ * Sonatype, Inc. Apache Maven is a trademark of the Apache Foundation. M2Eclipse is a trademark of the Eclipse Foundation.
+ * All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.proxy.walker;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.sonatype.nexus.proxy.item.RepositoryItemUid;
+import org.sonatype.nexus.util.Node;
+
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
+
+/**
+ * A helper class to "optimize" when some sort of "gathering for later processing" (most notable collection of
+ * directories to start Walks from) of repository paths happens, that are to be processed in some subsequent step by
+ * "walking" them (recurse them, a la "visitor" pattern). This utility class simply maintains a "tree" with
+ * <em>marked</em> nodes, while the meaning of "marked" is left to class user. One assumption does exists against
+ * "marked": some recursive processing is planned against it. It currently applies these simple rules to it.
+ * <p>
+ * <ul>
+ * <li>rule A: if parent node of the currently added node is already "marked", do not mark the currently added node
+ * (reason: it will be anyway processes when marked parent processing starts to recurse)</li>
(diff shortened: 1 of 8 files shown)


nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
```diff
@@ -0,0 +1,409 @@
...
+    {
+        // rule a: if parent is marked already, do not mark the child
+        if ( isParentMarked( changedNode ) )
+        {
+            changedNode.getPayload().setMarked( false );
+            return changedNode.getParent();
+        }
+
+        // rule b: if this parent's all children are marked, mark parent, unmark children
+        if ( isParentAllChildMarkedForRuleB( changedNode ) )
```
nabcos, line 264: this should be called 'siblings' IIUC?

cstamas:
CI build scheduled:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/103/

adreghiciu:
+1

cstamas:
CI voted :cake:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/103/

peterlynch:
+1 thx for ci build

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
```diff
@@ -0,0 +1,409 @@
...
+            final List<Node<Payload>> children = parent.getChildren();
+
+            if ( children.size() < 2 )
+            {
+                return false;
+            }
+
+            for ( Node<Payload> child : children )
+            {
+                if ( !child.getPayload().isMarked() )
```
nabcos, line 358: not sure I understand correctly:
- if `node` is the only child, it is treated as not marked and returns false
- if there are more children, it returns false if one of them is not marked

Doesn't that always return false, assuming that `node` has a non-marked payload as assumed in the first condition?

cstamas, line 358: The `node` parameter is always marked, just see how this method is invoked (entry point is addAndMarkPath()).

This code returns `true` in following case:

``` java
addAndMarkPath("/parent/sub1");
addAndMarkPath("/parent/sub2");
```

Second invocation will return `true` here.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    /**
+     * Returns true if parent exists (passed in node is not ROOT), and parent's all children are marked (their
+     * {@link Payload#isMarked()} is {@code true} for all of them.
+     * 
+     * @param node
+     * @return
+     */
+    protected boolean isParentAllChildMarkedForRuleB( final Node<Payload> node )
```
bdemers, line 343: confusing method name,  There is no matching "RuleA" method

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
```diff
@@ -0,0 +1,409 @@
...
+ * <p>
+ * This class also "optimizes" the tree size to lessen memory use. This "optimization" can be turned off, see
+ * constructors.
+ * <p>
+ * This class makes use of {@link Node} to implement the tree hierarchy.
+ * 
+ * @author cstamas
+ * @since 1.10.0
+ */
+public class ParentOMatic
```
bdemers, line 59: As much as i like anything named *-O-Matic,  I think this needs a more descriptive name.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
nabcos, line 358: But `if ( children.size() < 2 ) { return false }` that means you treat `node` as not marked in this condition?

nabcos:
Is there really no usable treenode class in the JRE other than the Swing one? otherwise okay aside of naming (although I think we should let all nexus classes end with -OMatic, just for style)

+1

cstamas:
Thanks Ben! It's going in.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/ParentOMatic.java
nabcos, line 358: I still think this is a bug:

```
cn.addPath( "/parent" );
cn.addAndMarkPath( "/parent/child" );
```

will return `false` for `isParentAllChildMarked...` here. Is that intended?

nabcos:
I still don't think ParentOMatic is a useful description of what this is. I will never find it again half a year from now, so we should really change that.

PR #132, opened 2011-12-07
o [NEXUS-4579] Option for snapshot removal task to delete immediately or move to trash

nabcos:
Backporting Snapshot Remover Tasks changes. Unchanged code but for one place regarding plexus/slf4j logger.

IT passes locally, I will try manually sometime tomorrow (separate +1 ;)


bdemers:
+1 if the only changed bits was the logger

cstamas:
What are those full file changes? Are we sneaking in bad line-endings?

nabcos:
that was pre- & post-'big EOL change commit'. The cherry-pick went without conflict (aside from logger) when I was converting the EOL for the touched files... you can see the detailed changes in the second commit, the first commit is only changing EOL.

cstamas:
+1

PS: (facepalm) i just realized this is merge into nexus-1.9.2.x, not master!

nabcos:
tested manually with 1.9.2.4-SNAPSHOT -> +1

PR #135, opened 2011-12-08
Remove key dependencies from Nexus Core

cstamas:
This is a follow-up of the "Timeline into plugin" pull request (https://github.com/sonatype/nexus/pull/116).

This pull request contains following changes:
- moving Timeline into a plugin (exactly same as in referenced pull request)
- removing Aether from Nexus Core (it's also "too important" stuff to pin it in core)
- removing Maven Indexer Artifact from core (more about this later)

The artifacts not in core classloader anymore (once this pull request applied):
- org.apache.lucene:lucene-core
- com.google.protobuf:protobuf-java
- jdom:jdom
- rome:rome
- org.sonatype.aether:\* (all 4 of them)
- org.apache.maven.indexer:indexer-artifact

All the "What does it buys for us?" explanation from referenced pull request still stands, and much more.

About Maven Indexer Artifact

This artifact contained classes that were originally part of Nexus. As we started Nexus indexer, those classes were also used in it. Later, as we moved Nexus Indexer into a plugin, we still left these classes in core. Meanwhile, many of the classes (well, some key classes at least) forms the basis of Nexus' Maven support too (just look at MavenRepository iface, and others). After donating the Indexer as "whole" to ASF, Nexus Maven support started relying on key classes not being present in Nexus core (like Gav, GavCalculator, etc). This led to similar "hostage" situation as we had with Timeline + Lucene (mentioned in previous pull request). Also, we had to make compromises wrt improvements we do, considering Maven Indexer other -- non Nexus -- integrations. All these just slowed down and complicated our advancement.

Maven Indexer _needs_ these classes, but does not use them in API. Nexus Maven support in contrary _needs_ them, but also _publishes them in API_ (again, see MavenRepository as example). Hence, they were "backforked", put in their proper packages and Core is changed to use them instead. Maven Indexer now has it's "own" indexer-artifact artifact as plugin dependency, while Nexus Core got back it's key classes needed for Maven support (with currently present compromises cleaned up).


nexus/nexus-api/src/main/java/org/sonatype/nexus/auth/AbstractSecurityEvent.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Sonatype Nexus (TM) Open Source Version is available from Sonatype, Inc. Sonatype and Sonatype Nexus are trademarks of
+ * Sonatype, Inc. Apache Maven is a trademark of the Apache Foundation. M2Eclipse is a trademark of the Eclipse Foundation.
+ * All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.auth;
+
+import org.sonatype.plexus.appevents.AbstractEvent;
+
+public class AbstractSecurityEvent
```
peterlynch: doc?

nexus/nexus-api/src/main/java/org/sonatype/nexus/auth/ClientInfo.java
```diff
@@ -18,21 +18,23 @@
  */
 package org.sonatype.nexus.auth;
 
-public class AuthenticationItem
+/**
+ * Client info about WHO is being doing something.
```
peterlynch: <del>being</del>

nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/java/org/sonatype/timeline/TimelineException.java
```diff
@@ -0,0 +1,59 @@
...
+ * (https://github.com/sonatype/spice-timeline) project. Since Nexus core is independent and decoupled from this
+ * project, and in one place there is only a class name check done against "org.sonatype.timeline.TimelineException"
+ * string, but due to how this helper works, it will try to instantiate the exception.
+ * 
+ * @author cstamas
+ * @since 1.10.0
+ * @see See {@link LogHelperPlexusResource} and {@link Nexus4427WarnErrorLogsToFeedsIT} IT class for how is it used.
+ * @deprecated This class has nothing to do with "real" Timeline exception. It is here only to serve the need of one IT:
+ *             Nexus4427WarnErrorLogsToFeedsIT
+ */
```
peterlynch, line 34: nice docs

peterlynch:
+1 other than a few class renames missing @since annos

peterlynch:
UM BTW what issue is this? - I want to make sure this gets tested and timeline gets a good once over

nexus/nexus-app/src/main/java/org/sonatype/nexus/DefaultNexus.java
```diff
@@ -387,7 +370,7 @@ public NexusStreamResponse getConfigurationAsStreamByKey( String key )
     // ----------------------------------------------------------------------------
 
     // creating
-
+/*
     @Deprecated
```
nabcos: /\* ? Dead code?

nexus/nexus-app/src/main/java/org/sonatype/nexus/Nexus.java
```diff
@@ -151,7 +144,7 @@ NexusStreamResponse getConfigurationAsStreamByKey( String key )
     // ----------------------------------------------------------------------------
 
     // creating
-
+/*
     @Deprecated
```
nabcos: Dead code?

nabcos:
+1

IIUC merging this in will generate a new issue regarding the Feed UI? There were some remarks in #116 about that.

cstamas:
Yes, Feed UI needs at least (this would be a "minimal requirement") a change, that would allow it to gracefully handle HTTP 404 response when filling up the Feeds grid.

Everything "more than that" (moving it into plugin, whatever) is just an extra that would not buy us a lot.

cstamas:
CI build voted +1 once:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/108/

bdemers:
-1 
Try not to lump unrelated changes together in one pull.  I think smaller more focused pull requests would be easier to deal with (review, revert, etc ).  This pull has changes related to a few different areas.

For example ccb1682e2730fefcbbae645d661dd039bc12fc1a, is about moving other libs out of nexus core 

-I'm still -1 on moving the timeline out until we figure out what we should do with the UI (or we decide an empty feed UI is a good thing if a user removes the timeline plugin)-
We talked about this after scrum the feed UI is NOT a concern.

Invalid license header for: https://github.com/sonatype/nexus/blob/ccb1682e2730fefcbbae645d661dd039bc12fc1a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/version/GenericVersion.java
This has commits from maven indexer, it needs to be Apache licensed

Verify any other code copied from indexer has valid licenses and have license header exclusions in the root pom.

I don't mean to be only negative,  I think the changes are good, there is a lot of good here too.

cstamas:
IMO, This pull contains related changes: they are all about _moving out key artifacts from core_, to not have them "pinned", and as we see today, make PITA and hell loose to bump their versions.

Aretta banned my IP again, but am willing to have skype (iChat or whatever else) about this (something that can conference too).

As for "backforked" code, and code copy-pasted from Aether (they are needed until Maven support itself is moved out from core into a plugin, then those should be dropped and switched back to Aether).

Finally, there are already two developer +1's and one CI +1 on this pull, so, I am not worried about fate of this pull ;)

After consulting about what license headers are needed for "backforked" Maven Indexer files, and what for copied Aether files, and -- if you guys want -- after having a chat, I will merge this in master (Pro part locally prepped), along with PluginManager fixes I did on other line (pull soon to be created).

cstamas:
CI:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/117/

PR #136, opened 2011-12-08
o [NXCM-3549] Add filtering of repo/group task setting combobox

nabcos:
Removal of repositories from the scheduled tasks UI was not supported. This change adds a filter based on the task id.

please check the filter expressions for sanity:
- DownloadIndexesTask : (repoType == 'proxy' && format == 'maven2') 
- OptimizeIndexTask : (format == 'maven2') 
- PublishIndexesTask : (format == 'maven2') 
- EvictUnusedProxiedItemsTask : (repoType == 'proxy')
- RebuildMavenMetadataTask : (format == 'maven2') 
- RepairIndexTask : (format == 'maven2') 
- UpdateIndexTask : (format == 'maven2') 
- SynchronizeShadowsTask : (repoType == 'virtual') 
- SnapshotRemoverTask : (format == 'maven2' && repoPolicy == 'SNAPSHOT')

### DIFF
--- a/nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js
+++ b/nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js
@@ -176,12 +176,25 @@ Sonatype.repoServer.SchedulesEditPanel = function(config) {
   };
 
   // A record to hold the name and id of a repository
-  this.repositoryRecordConstructor = Ext.data.Record.create([{
-        name : 'id'
-      }, {
-        name : 'name',
-        sortType : Ext.data.SortTypes.asUCString
-      }]);
+  this.repositoryRecordConstructor = Ext.data.Record.create(
+    [
+        {
+            name:'id'
+        },
+        {
+            name:'name',
+            sortType:Ext.data.SortTypes.asUCString
+        },
+        {
+            name:'repoType'
+        },
+        {
+            name:'format'
+        },
+        {
+            name:'repoPolicy'
+        }
+    ]);
 
   // A record to hold the name and id of a repository group
   this.repositoryGroupRecordConstructor = Ext.data.Record.create([{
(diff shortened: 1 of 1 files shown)


cstamas:
:beer:

nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js
```diff
@@ -176,12 +176,25 @@ Sonatype.repoServer.SchedulesEditPanel = function(config) {
...
+        {
+            name:'repoType'
+        },
+        {
+            name:'format'
+        },
+        {
+            name:'repoPolicy'
+        }
+    ]);
```
peterlynch, line 198: does maven java formatting have to come to js too?

peterlynch:
+1 like the idea and this solves immediate needs - am wondering if this filtering is handled on server side as well

nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js
nabcos, line 198: Please define the format style for javascript.

nabcos, line 198: Aside that IMO it's better than parts [like this](https://github.com/sonatype/nexus/blob/6d7582e521ffb328fc771a8ed7f770f5696b463c/nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js#L1484) where there is no recognizable indention at all...

nabcos:
@adreghiciu Agreed, but I'm not sure how to do that. IIUC using a 'remote' store might work that way, that (maybe ;) would query the repositories list resource with some parameters and get the filtered list directly from REST... but I think then we would have to mangle the list (add repo/group to name, add 'all repos' entry) also server-side... not sure how that would work.

(I did not investigate that because this was a one-beer-issue, and I was at about one and a half already because of ext js wrangling...)

nabcos:
@peterlynch WDYM with server-side filtering? the tasks will check if they support a repository and do nothing if they're don't.

nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.SchedulesEditPanel.js
peterlynch, line 198: no worries

peterlynch:
@nabcos sounds like the server side is sanely handing dealing with the different repos. good enuff

PR #143, opened 2011-12-12
o Turn off maven-specific fields in repo editor UI for non-maven-repositories

nabcos:
This change only enables repoPolicy, indexable and download remote index field for repositories with maven format. It also adds hooks before/after the repo editor contents change, so plugins do not have to handle load and repoProvider.select (for 'New Repository') differently.


peterlynch:
+1

velo:
+2

nabcos:
Adding some commits here:
- moving common code into abstract repo editor
- changing repository policy field to be enabled only for non-maven repositories

PR #146, opened 2011-12-13
The column model for the repo list grid is now built in array before pan...

dbradicich:
...el is created, then an event is spawned so plugins can mangle the column model (and toolbar) if necessary


peterlynch:
what issue is this related too and what impact will these changes have?

peterlynch:
+1

dbradicich:
ya sorry, didnt know the initial ticket number when i started the changes ;)  but just to add extension points for the health check plugin to have its way with the grid

nabcos:
+1

PR #149, opened 2011-12-14
Adding more features to Plugin Manager.

cstamas:
Extending abilities of plugin manager:
- ability to "enable"/"disable" a plugin in plugin repository
- spit warnings about deprecated plugins being present

### DIFF
--- a/nexus/nexus-plugins/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
+++ b/nexus/nexus-plugins/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
@@ -42,6 +42,7 @@ import org.sonatype.guice.nexus.binders.NexusAnnotatedBeanModule;
 import org.sonatype.guice.plexus.binders.PlexusXmlBeanModule;
 import org.sonatype.guice.plexus.config.PlexusBeanModule;
 import org.sonatype.inject.Parameters;
+import org.sonatype.nexus.logging.AbstractLoggingComponent;
 import org.sonatype.nexus.mime.MimeSupport;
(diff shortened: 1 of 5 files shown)


velo:
Don't we need to change plugin-console as well to support this?

BTW, is there a jira ticket related to this?

adreghiciu:
How do you enable/disable a plugin?
When enabling/disabling takes place? on next Nexus start?

cstamas:
Yes, this is checked only during boot. When you _disable_ a plugin, it will not deactivate it (if plugin activated). It will get applied on next reboot (since plugin manager, or if i am right, plexus-shim, cannot unload realms).

All this makes possible is to stop using "optional" folder for plugins, we can ship them simply "disabled". Also, this makes possible for some component to _enable_ a plugin (based on license?), do a reboot, and voila, you have it here.

cstamas:
On the contrary, plugin manager can "activate" at any time, so we could have UI for this, and let users click around to enable some optional plugins. But "disabling" a plugin would still need reboot to get applied.

velo:
Cool, well, create a ticket for the UI change then you get my plus =D

adreghiciu:
I do not think that Nexus is prepared to cope with a plugin being enabled. Some plugins may work some not. E.g. some plugins may be waiting for NexusInitialized or NexusStarted, some maybe do not recheck the List/Map/Set injected with contributed components from plugins, things like that. Beside the fact that I like the idea I think this will only get confusing

cstamas:
No, this will actually prevent them from loading at all ("activating" is actually loading them up and creating plexus injector for them). But you are right, this is incomplete. This was just an idea, how could we handle the oncoming requirement to do this. Will not merge but will leave it here for reference.

Alin, you comment made me realize, that plugin interdependencies are not covered here. Closing and will think about this more.

PR #151, opened 2011-12-14
nexus-test-environment-plugin throw error if any system dependency is present

velo:
When yourkit profile is active it includes system dependency, which cause errors if we try to resolve it, since system deps are on some know location on hard drive test can use it from there.

Found that while profiling NXCM-3572


adreghiciu:
+1

nabcos:
+1

PR #155, opened 2011-12-15
[NEXUS-4653] Show timezone after the time fields

adreghiciu:


adreghiciu:
As only JS changed there is no need for a feature build.

velo:
Screen shot would be sweet =D

+1

nabcos:
+1

peterlynch:
+1
agree with @velo screenshot would be called for. One of us will attach to issue.

PR #156, opened 2011-12-15
NEXUS-4715: fixed a major bug and improvements

cstamas:
Shadows were neglected a bit.

The bug

The call from master happened on wrong method, thus it was avoiding UID locking but also NFC management (and many other things). This way it not only let Nexus instance to access same files from multiple threads, but also generated unjustified traffic (if master is proxy).

The improvements

Methods that were doubling the master setting were deprecated (setMasterRepositoryId, and it's getter pair), since they just double the code and were not used (except few classes in REST API) with easy changes.

Also, source code "refreshed" by adding @Overrides, finals and fixing some (mainly) minor code issues.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/ShadowRepository.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/ShadowRepository.java
@@ -41,22 +41,6 @@ public interface ShadowRepository
     ContentClass getMasterRepositoryContentClass();
 
     /**
-     * Returns the master repository of this ShadowRepository.
-     * 
-     * @return
-     */
-    String getMasterRepositoryId();
-
-    /**
-     * Sets the master repository of this ShadowRepository.
-     * 
-     * @param masterRepository
-     * @throws IncompatibleMasterRepositoryException
-     */
-    void setMasterRepositoryId( String masterRepositoryId )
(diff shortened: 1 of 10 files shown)


cstamas:
Started CI job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/124/

velo:
I put CI to test if this fixes NEXUS-2692 tests
https://builds.sonatype.org/job/nexus-oss-its-debug/
let it run for a day or two...

nabcos:
+1, but this did not fix NEXUS-2692:

```
Failed tests:   testEvictPublicGroup(org.sonatype.nexus.integrationtests.nexus2692.Nexus2692EvictGroupTaskIT):     
The following files should have been deleted: central-m1/org.apache.maven.plugins/poms/maven-plugins-11.pom
```

https://builds.sonatype.org/job/nexus-oss-its-debug/jdk=java-6x,label=win/3469

cstamas:
Rebased against master, made shadows back to sync, CI started:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/126/

nabcos:
the link to build 3469 has disappeared, this one is kept for reference:
https://builds.sonatype.org/job/nexus-oss-its-debug/3479/

nabcos:
+1

peterlynch:
+1
When it gets moved to be tested please describe a scenario how it can be verified

PR #159, opened 2011-12-19
Make possible to set the default log level for nexus via a system property

adreghiciu:
This log back.xml is used by nexus for logging up till the moment that our log back configuration kicks in.
Without being able to change this logging as for example from Sisu it will not be shown.


cstamas:
:beer:

nabcos:
+1

adreghiciu:
Described the property in wiki: http://goo.gl/ZPqw8

PR #160, opened 2011-12-19
[NEXUS-4723[ Make REST test utility classes reusable via nexus-launcher

adreghiciu:


cstamas:
Looks fine, but a sidenote: we have similar case as with commons-io, is not a direct dependency of core (it gets in as transitive one), but tests (both UT/IT) are for some reason extensively using it (even if 99% of equivalent methods are present in plexus utils).

This same stands for indexer-artifact. Classes like Gav are not present in core anymore (the Maven Indexer Gav, while Nexus got it's own equiv class), but tests (mostly ITs) are still using it...

nabcos:
+1

adreghiciu:
Kicked off feature build: https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/128/

adreghiciu:
@cstamas I might miss something here, but I do not get the connection between your comment and this pull request

cstamas:
It was just a fyi "Maven Indexer Gav classes are used", just I realized while looking at the diffs. Not related to this pull req at all.

adreghiciu:
@cstamas Got it now. Well, I just used the current functionality so no change in that direction.

adreghiciu:
The build https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/128/ was passing with only one test failing which is also failing for master build.
I consider that CI voted +1

cstamas:
:kiss:

PR #162, opened 2011-12-20
NXCM-3600: exposed fix

cstamas:
CI job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/129/


nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
```diff
@@ -252,7 +255,35 @@ protected String parsePathFromUri( String parsedPath )
...
+        final ResourceStore resourceStore = getResourceStore( request );
+
+        // NXCM-3600: if this is not a readOnly action, and we deal with a Repository instance
+        // that is not exposed, we forbid access to it. And read operation will be handled by
+        // Repository level security anyway.
+        if ( !action.isReadAction() && resourceStore instanceof Repository )
+        {
+            if ( !( (Repository) resourceStore ).isExposed() )
+            {
+                throw new NoSuchRepositoryException( ( (Repository) resourceStore ).getId() );
```
bdemers, line 271: Not completely true, as the repo does exist. Though, I think this will do, as it is almost true.

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
```diff
@@ -252,7 +255,35 @@ protected String parsePathFromUri( String parsedPath )
      * @throws NoSuchRepositoryGroupException
      * @throws NoSuchRepositoryRouterException
      */
-    protected abstract ResourceStore getResourceStore( Request request )
+    protected ResourceStore getResourceStore( final Request request, final Action action )
```
bdemers, line 258: Verify this method is not overridden in a PRO plugin (we need to know what to back port 1.9.2.4)

bdemers:
Changes look good, but add a test ;)

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
cstamas, line 258: Read the diff better, the original method is intact, it is just a new method is added with different signature (but the diff probably mislead you), and it is called from abstract class instead of the abstract getResourceStore( req ).

Subclasses needs no change, unless they override one of the get(), upload() or delete() methods. In core, there is only the RRB plugins class. Will see what's in Pro.

cstamas:
@bdemers Sure, IT on the way, I just wanted to present needed changes for you to review them, since they are related to introduction of "view" perm made by you.

cstamas:
CI likes it:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/129/

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
cstamas, line 271: Very same things happen with "protected" repository registry and user not having "view" perm. For "surface", both of those end up with 404 Not found.

cstamas:
CI likes it even more:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/130/

Note: while both slaves failed the build (!), there was no UT/IT error, but it was the fact that very last "licence check" failed, since newly added IT had no license header.

This is already fixed in last commit (e2eca1a).

cstamas:
Added 2nd IT too.

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
bdemers, line 258: I see it now

nexus/nexus-test-harness/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nxcm3600/AbstractNxcm3600IntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+     * Sets the exposed flag of repository.
+     * 
+     * @param exposed
+     * @throws IOException
+     */
+    protected void setExposed( final boolean exposed )
+        throws IOException
+    {
+        TestContainer.getInstance().getTestContext().useAdminForRequests();
+        TestContainer.getInstance().getTestContext().setSecureTest( true );
```
bdemers, line 65: setSecureTest should already be called called by the parent class

I see now... sorry

bdemers:
+1
though, I'd like to see a PUT along with the GET and DELETE tests

cstamas:
One more CI +1:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/132/

nabcos:
+1

PR #165, opened 2011-12-23
o [NEXUS-4732] Use local jetty as endpoint for proxy repositories

nabcos:
The ResolvingLRTest was going remote and failed the build because of
hiccups on the remote side. It's now changing all proxy repo urls to point to a local jetty before starting resolving artifacts.


peterlynch:
+1 nice  - I was about to file a report for this cause I didn't notice a reference to the issue on the grid failure. The http test server fluent api looks helpful. Where does that live now?

nabcos:
it's currently living in https://github.com/sonatype/http-testing-harness/tree/master/server-provider

bdemers:
+1

cstamas:
:cake:

nabcos:
reference build was missing, started here: https://builds.sonatype.org/view/nexus/job/nexus-oss-feature/25

PR #168, opened 2011-12-28
NEXUS-4735: fix

cstamas:
This is not fixing the culprit -- which is still unknown to me (unable to reproduce), but I assume we will not see that problem again: it happened once on CI win slave.


nabcos:
+1

cstamas:
CI job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/140/

cstamas:
CI build failed with same error as issue this pull request should fix, so am closing this pull req since it obviously does not fix it.

PR #171, opened 2012-01-03
NEXUS-4735 as initial reason, but many IT fixes

cstamas:
The inital problem (log file kept open on windows making it impossible to delete) led me to believe there was a component in IT plexus container that kept the file open.

This led to realization that we "bounce" Plexus per-IT class, and this is not the case anymore. Plexus is brought up once (lazily) and kept the same instance over and over. Moreover, how security was managed in ITs was next to a disaster, and it turned out that many ITs did "interleave" (was relying on side effect!).

After fixing plexus and security, next was to remove the "feature" of ITs that it was bringin' up almost all of Nexus internals within IT (to "validate" configuration [why?], to mangle security configuration [why?] and many more). So, alternative implementations are added that does NOT bring up all of those components, except in few (2 i believe) ITs where they are a must to decipher some of the SMTP related passwords that would come in scrambled if using "low level" model loaded only.

Finally, it turned out that a few ITs (within same group) was responsible for more than 20minutes of runtime, and they were even not doing their job: they were silently timeouting (auto block related ITs).

### DIFF
--- a/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/utils/NexusConfigUtil.java
+++ b/nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/utils/NexusConfigUtil.java
@@ -19,11 +19,14 @@
 package org.sonatype.nexus.test.utils;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
 
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.sonatype.configuration.validation.InvalidConfigurationException;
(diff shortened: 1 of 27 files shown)


cstamas:
Started CI:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/141/

nabcos:
+1

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/utils/NexusConfigUtil.java
```diff
@@ -33,14 +36,28 @@
...
 
+import com.google.common.io.Closeables;
+import com.google.common.io.Flushables;
+
+/**
+ * This class tampers with nexus configuration files, not on the instance! It is meaningful to use it BEFORE nexus is
+ * booted (to set it up), or AFTER test is done (to verify config changes did happen), but during test the changes done
+ * by this class will be lost! (since nexus keeps config copy in memory, and on next save will simply overwrite it with
+ * own copy, or not be aware of the changes at all)
+ *
```
peterlynch, line 58: nice!

peterlynch:
+1

cstamas:
Closed due to CI failure, more on this on ML.

PR #172, opened 2012-01-03
NEXUS-4744: fix

cstamas:
The original NPE was not due to load, but due to config event and how workingFolder was
handled. As it turns out, having AttributeStorage components listen for config events
(and doing whatever they did) was totally nonsense, since there are no means that
a config event (related to nexus.xml changes) would change the nexus-work folder!
Doing the latter is possible only by bouncing Nexus. Hence, scrubbed.

Moreover, the way how Nexus decides is "transitioning" (for attributes) needed or not is
much smarter now with meaningful log messages. Before, we would waste CPU cycles (if not
actual IO) just to invoke methods on "legacy" storage that would simply return doing nothing.
Now, we detect the need for "transitioning", and on new instances we even don't install
any of those components, thus, not wasting any cycles on them.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/TransitioningAttributeStorage.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/TransitioningAttributeStorage.java
@@ -20,12 +20,6 @@ package org.sonatype.nexus.proxy.attributes;
 
 import java.io.IOException;
 
-import javax.enterprise.inject.Typed;
-import javax.inject.Inject;
-import javax.inject.Named;
-import javax.inject.Singleton;
-
-import org.sonatype.nexus.logging.AbstractLoggingComponent;
 import org.sonatype.nexus.proxy.access.Action;
 import org.sonatype.nexus.proxy.item.RepositoryItemUid;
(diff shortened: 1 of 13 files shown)


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/TransitioningAttributeStorage.java
```diff
@@ -91,8 +79,7 @@ public Attributes getAttributes( final RepositoryItemUid uid )
                         }
                         catch ( IOException e )
                         {
-                            // nag and ignore it
-                            getLogger().debug( "Problem during legacy attribute deletion!", e );
+                            // legacy swallows them, this is needed only to satisfy it's signature
```
bdemers, line 82: I still think this should be loged (maybe at trace?)

peterlynch:
+1 I need this fixed to make useful load tests for 2.0

nabcos:
+1

adreghiciu:
@bdemers I see you already are splitting the build ;)

bdemers:
@adreghiciu that change is unrelated to splitting the build, though splitting the build could supersede it.

cstamas:
Merged, but just for the record, CI job was started this morning:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/150/

PR #173, opened 2012-01-04
NEXUS-4735 as initial reason, but many (but less than previously) IT fixes

cstamas:
The inital problem (log file kept open on windows making it impossible to delete) led me to believe there was a component in IT plexus container that kept the file open.

Plexus is now bounced again per-IT-class, just as before.

Also, how security was managed in ITs was next to a disaster, and it turned out that many ITs did "interleave" (was relying on side effect!).

After fixing security, next was to remove the "feature" of ITs that it was bringin' up almost all of Nexus internals within IT (to "validate" configuration [why?], to mangle security configuration [why?] and many more). So, alternative implementations are added that does NOT bring up all of those components, except in few (2 i believe) ITs where they are a must to decipher some of the SMTP related passwords that would come in scrambled if using "low level" model loaded only.

Finally, it turned out that a few ITs (within same group) was responsible for more than 20minutes of runtime, and they were even not doing their job: they were silently timeouting (auto block related ITs).


nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/application/upgrade/Upgrade146to1100.java
```diff
@@ -102,6 +104,18 @@ public CRepository upgradeCRepository( org.sonatype.nexus.configuration.model.v1
         };
 
         org.sonatype.nexus.configuration.model.Configuration newc = versionConverter.upgradeConfiguration( oldc );
+        
+        // this should go into much "older" upgrader, this was a mistake!
```
bdemers, line 108: why not put it in an older upgrader ?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/threads/ThreadPoolManager.java
```diff
@@ -29,4 +29,6 @@
     void createPool( Repository repository );
 
     void removePool( Repository repository );
+    
+    // void shutdown();
```
bdemers, line 33: dead code

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/threads/ThreadPoolManagerEventInspector.java
```diff
@@ -37,12 +42,18 @@
     @Override
     public boolean accepts( Event<?> evt )
     {
-        return evt instanceof RepositoryRegistryRepositoryEvent;
+        return evt != null && evt instanceof RepositoryRegistryRepositoryEvent;
+        // return evt != null && ( evt instanceof RepositoryRegistryRepositoryEvent || evt instanceof NexusStoppedEvent );
```
bdemers, line 46: dead code

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/threads/ThreadPoolManagerEventInspector.java
```diff
@@ -52,5 +63,9 @@ else if ( evt instanceof RepositoryRegistryEventRemove )
         {
             poolManager.removePool( ( (RepositoryRegistryEventRemove) evt ).getRepository() );
         }
+        // else if ( evt instanceof NexusStoppedEvent )
+        // {
+        // poolManager.shutdown();
+        // }
```
bdemers, line 69: dead code

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
```diff
@@ -350,25 +341,26 @@ public void oncePerClassSetUp()
...
-                getNexusConfigUtil().enableSecurity(
+                final boolean testRequiresSecurityEnabled =
                     TestContainer.getInstance().getTestContext().isSecureTest()
-                        || Boolean.valueOf( System.getProperty( "secure.test" ) ) );
+                        || Boolean.valueOf( System.getProperty( "secure.test" ) );
 
-                // start nexus
-                startNexus();
+                // set security enabled/disabled as expected by current IT
+                TestContainer.getInstance().invokeAsAdministrator( new Callable<Object>()
```
bdemers, line 356: This is a little verbose.
But it is MUCH better then having the test deal with which user to use for requests!

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
```diff
@@ -930,24 +920,26 @@ private static void takeSnapshot()
         }
     }
 
+    /**
+     * Returns the basedir.
+     * 
+     * @return
+     * @deprecated Use {@link TestContainer#getBasedir()} instead.
+     */
+    @Deprecated
```
bdemers, line 933: why deprecate this?  this seems useful rather then calling TestContainer.getBasedir()
Why not just leave this method in place?

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
```diff
@@ -930,24 +920,26 @@ private static void takeSnapshot()
...
-        if ( basedir == null )
-        {
-            basedir = new File( "" ).getAbsolutePath();
-        }
-
-        return basedir;
+        return TestContainer.getBasedir();
     }
 
+    @Deprecated
```
bdemers, line 939: add @deprecated javadoc

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/utils/NexusConfigUtil.java
```diff
@@ -33,14 +36,28 @@
...
+
+/**
+ * This class tampers with nexus configuration files, not on the instance! It is meaningful to use it BEFORE nexus is
+ * booted (to set it up), or AFTER test is done (to verify config changes did happen), but during test the changes done
+ * by this class will be lost! (since nexus keeps config copy in memory, and on next save will simply overwrite it with
+ * own copy, or not be aware of the changes at all)
+ * 
+ * @author cstamas
+ */
 public class NexusConfigUtil
```
bdemers, line 61: Good comments

bdemers:
Other then some minor stuff +1

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/application/upgrade/Upgrade146to1100.java
cstamas, line 108: because older one will eventually never execute (think X-1 > X upgrade). Ultimately, an "upgrader" component that went out once with a release X should _never_ be modified in any of the Y (X < Y) release. Hence, it's always the _latest_ upgrader that is changed (even if we spot that some older upgrader is bad _after_ we released it(.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/threads/ThreadPoolManager.java
cstamas, line 33: true, i was toying with idea to drop the Plexus specific lifecycle (Disposable et al). But in last moment, I gave up, did not want to disturb it in this changeset.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/threads/ThreadPoolManagerEventInspector.java
cstamas, line 46: Left here for reference (when we leave plexus apis completely).

nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
cstamas, line 933: I like to have _one source of truth_. Having two public methods that returns basedir will confuse devs in future.

Having this non-deprecated will put us in that situation: two methods seemingly doing the same....

bdemers, line 933: I agree with the one source of true.  Its just that for the longest time people have been used to this method in various test support classes (PlexusTestCase.getBasedir() for example)

if you really feel strong about it, leave it deprecated

adreghiciu:
+1

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/application/upgrade/Upgrade146to1100.java
bdemers, line 108: can you add a note in the comment, about this.  So no one will read the comment and think it should be moved in the future.

cstamas:
CI +1:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/151/

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/source/FileConfigurationSource.java
```diff
@@ -195,6 +196,8 @@ public Configuration loadConfiguration()
 
         ValidationResponse vResponse =
             getConfigurationValidator().validateModel( new ValidationRequest( getConfiguration() ) );
+        
+        dumpValidationErrors( vResponse );
```
velo, line 200: AWESOME!!!

velo:
+1

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/source/FileConfigurationSource.java
velo, line 200: ow, I though there was no dump before, looks like there was =D

PR #178, opened 2012-01-05
NXCM-3564, adding nexus-proxy profile to run UT's against http client 4 (similar to the ahc tests).  Fixes a couple small problems that prevented the tests from passing.

bdemers:


nabcos:
+1

peterlynch:
+1

nexus/nexus-proxy/pom.xml
```diff
@@ -277,6 +277,32 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>apacheHttpClient4x-provider</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
```
velo, line 286: Do we really need to re-run ALL tests or should it be just a set of test?

velo:
+1

PR #180, opened 2012-01-08
NXCM-3651: fix for EI host

cstamas:
The fix for EventInspectorHost that caps the pool size and makes sure Nexus instance on heavy load
will not create cached threads in host pool endlessly (leading to OOM probably), but
will keep pool size capped. Also, when pool is full, caller threads will start to execute the
async inspectors.

This fix does NOT solve the indexer, when the threads were actually detected ("seen")
to stockpile up.

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/events/DefaultEventInspectorHost.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/events/DefaultEventInspectorHost.java
@@ -26,8 +26,6 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor.CallerRunsPolicy;
 import java.util.concurrent.TimeUnit;
 
-import javax.inject.Inject;
-
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Disposable;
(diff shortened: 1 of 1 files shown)


nabcos:
+1

nexus/nexus-app/src/main/java/org/sonatype/nexus/events/DefaultEventInspectorHost.java
```diff
@@ -21,120 +21,123 @@
...
-import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.ThreadPoolExecutor.CallerRunsPolicy;
+import java.util.concurrent.TimeUnit;
+
+import javax.inject.Inject;
 
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
```
velo, line 30: @Inject and @Component being used at same time....
is that valid?

peterlynch, line 30: Agree - shouldn't it use one or other?

peterlynch:
Document org.sonatype.nexus.events.DefaultEventInspectorHost.poolSize on wiki if this gets accepted?

+1 if tests pass and @Inject @Component gets addressed

bdemers:
+1 assuming we remove @inject (or remove @component/@Requirement)

nexus/nexus-app/src/main/java/org/sonatype/nexus/events/DefaultEventInspectorHost.java
cstamas, line 30: Still plexus, so @Inject is a mistake...

cstamas:
@peterlynch documentation already done https://docs.sonatype.com/display/Nexus/Nexus+Configuration+Properties#NexusConfigurationProperties-%7B%7Borg.sonatype.nexus.events.DefaultEventInspec...

PR #186, opened 2012-01-11
NEXUS-4468 - Wastebasket now delete empty directories

velo:
Wastebasket was not touching directories.

Now, if the dir is empty it will be removed.


velo:
Build launched
 https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/158/

PR #189, opened 2012-01-12
Drop capabilities from nexus core

adreghiciu:


nabcos:
+1

cstamas:
:beer:

But, none of the OSS bundle (nexus-core-webapp or nexus-core-oss-webapp) will include the new capabilities plugin? Asking, since it was included with previous versions

nabcos:
true, capabilities should be included in OSS bundles IMO.

velo:
+1

PR #190, opened 2012-01-12
Improve Empty trash task legacy support!

velo:
Basically I made 4 changes to EmptyTrash based on NEXUS-4642 and NEXUS-4468:
- trash root level now is holly and no longer deleted  (applied to new and legacy trash)
- using scanner when ALL trash should be purged (applied to new and legacy trash)
- legacy trash now is clean up when age is set (before it only ran with ALL was set)
- new EmptyTrashTask now removes empty directories.

### DIFF
--- a/nexus/nexus-proxy/pom.xml
+++ b/nexus/nexus-proxy/pom.xml
@@ -138,6 +138,11 @@
       <groupId>org.sonatype.sisu</groupId>
       <artifactId>sisu-charger</artifactId>
     </dependency>
+  <dependency>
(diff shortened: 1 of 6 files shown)


velo:
CI started.. https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/160/
+1

adreghiciu:
-1. We should use the file matchers for litmus not add a new class

velo:
good pointers, let's get trash and litmus...

PR #192, opened 2012-01-12
Making the uploaded file name unique.

cstamas:
This change still keeps the "first wins" (better last wins, as last overwriting wins) for
file uploads, but fixes the intermittent problems reported in logs as IOExceptions.


nabcos:
+1

velo:
+1

peterlynch:
NEXUS-4764 should be fixed by this.
+1

PR #194, opened 2012-01-12
Improve Empty trash task legacy support!

velo:
Basically I made 4 changes to EmptyTrash based on NEXUS-4642 and NEXUS-4468:
- trash root level now is holly and no longer deleted (applied to new and legacy trash)
- using scanner when ALL trash should be purged (applied to new and legacy trash)
- legacy trash now is clean up when age is set (before it only ran with ALL was set)
- new EmptyTrashTask now removes empty directories.


adreghiciu:
+1

nabcos:
+1

nexus/nexus-test/nexus-test-common/pom.xml
```diff
@@ -61,6 +61,17 @@
...
+      <groupId>org.sonatype.sisu.litmus</groupId>
+      <artifactId>litmus-testsupport</artifactId>
+      <version>1.1-SNAPSHOT</version>
+      <exclusions>
+        <exclusion>
+          <artifactId>mockito-all</artifactId>
+          <groupId>org.mockito</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
peterlynch, line 74: hmmm are  we sure this didn't just bump a bunch of dependencies? - I thought sisu-litmus depended on some newer stuff? Would like to see before and after of mvn dependency:tree

peterlynch, line 74: and now we need to release sisu-litmus - can we avoid this until post 2.0

velo, line 74: Either we avoid duplicated code or sisu-litmus...
https://github.com/sonatype/nexus/pull/194#commitcomment-861429

I'm fine with either way.

adreghiciu:
@velo What did you had to change in litmus?

velo:
I added this:
https://github.com/sonatype/sisu-litmus/commit/c8a06d1b9b00e6b586c09ad4423bb6035c31ca04

adreghiciu:
+1 on that and using litmus snapshot

velo:
Yeah I agree litmus is a better choice... so @peterlynch ?

peterlynch:
@velo Can you confirm that no existing nexus-test-common dependencies got version bumped by adding sisu-litmus?

I am all for sisu-litmus - it is the future and I will be first in line to add it after 2.0 - I am just looking at avoiding risk of ITs not passing because of some dependency change like this just for the sake of not temporarily duplicating a little bit of code.

Can you confirm that adding this dependency did not just bump other versions of dependencies. If it didn't I am +1. If it did, force them to previous versions and I am +0. If you can't do that or avoid bumping other versions -1.

peterlynch:
Also I think Rich needs to give approval to these changes to the trash task if not already since they may impact release notes.

velo:
@peterlynch I don't get it... isn't test-common a <scope>test</ dependency?  It shall not impact nexus...

velo:
I just checked using mvn dependency:tree -pl :nexus-proxy

It does add a bunch of dependencies, all scope test.
[INFO] |  +- org.sonatype.sisu.litmus:litmus-testsupport:jar:1.1-SNAPSHOT:test
[INFO] |  |  +- ch.qos.logback:logback-classic:jar:1.0.0:runtime (version managed from 0.9.28; scope managed from test)
[INFO] |  |  |  - ch.qos.logback:logback-core:jar:1.0.0:runtime
[INFO] |  |  +- org.sonatype.gossip:gossip-support:jar:1.8:test
[INFO] |  |  |  - org.sonatype.gossip:gossip-bootstrap:jar:1.8:test
[INFO] |  |  +- com.intellij:annotations:jar:9.0.4:test
[INFO] |  |  +- org.incava:java-diff:jar:1.1:test
[INFO] |  |  +- org.powermock:powermock-module-junit4:jar:1.4.9:test
[INFO] |  |  |  - org.powermock:powermock-module-junit4-common:jar:1.4.9:test
[INFO] |  |  |     - org.powermock:powermock-reflect:jar:1.4.9:test
[INFO] |  |  +- org.powermock:powermock-api-mockito:jar:1.4.9:test
[INFO] |  |  |  - org.powermock:powermock-api-support:jar:1.4.9:test
[INFO] |  |  +- org.powermock:powermock-module-junit4-rule:jar:1.4.9:test
[INFO] |  |  |  +- org.powermock:powermock-classloading-base:jar:1.4.9:test
[INFO] |  |  |  - org.powermock:powermock-core:jar:1.4.9:test
[INFO] |  |  |     - org.javassist:javassist:jar:3.14.0-GA:test
[INFO] |  |  +- org.powermock:powermock-classloading-xstream:jar:1.4.9:test
[INFO] |  |  - com.googlecode.junit-ext:junit-ext:jar:junit45:1.0-RC3:test

rseddon:
+1

PR #200, opened 2012-01-16
Nxcm 3642 fire expirecaches event on cache alteration

jdillon:
Potential fix for https://issues.sonatype.org/browse/NXCM-3642 and general optimization for how/when RepositoryEventExpireCaches fired.

This change will only fire this event if the cache was altered.

A few changes to the EhCache impl to expose meaningful results for removeWithChildren() invocation.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/cache/EhCachePathCache.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/cache/EhCachePathCache.java
@@ -109,14 +109,16 @@ public class EhCachePathCache
 
         String keyToRemove = makeKeyFromPath( path );
 
+        boolean removed = false;
         for ( String key : keys )
         {
             if ( key.startsWith( keyToRemove ) )
             {
-                getEHCache().remove( key );
+                boolean tmp = getEHCache().remove( key );
+                removed = removed || tmp;
             }
         }
-        return true;
+        return removed;
     }
 
     public void doPurge()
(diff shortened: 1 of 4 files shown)


adreghiciu:
+1

jdillon:
Updated to always fire new RepositoryEventNfcCleared event (subclass of RepositoryEventExpireCaches) which includes the cache alteration detail.

jdillon:
Latest change...

  Moved RepositoryEventExpireCaches firing to expireCaches() method, always fire this at the end for a hint to anything that might need it.
  Added separate RepositoryEventNfcCleared event which is only fired when the NFC has been altered

Some discussion w/Alin about how is the best way to fix this (to clean up the meaning of the events).  I'm gonna leave it here and let someone else figure out how to move forward.  As long as the extra detail is exposed about if the NFC is actually altered, then this should help fix infinite message cycle in smartproxy.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/cache/EhCachePathCache.java
```diff
@@ -109,14 +109,16 @@ public boolean removeWithChildren( String path )
...
         {
             if ( key.startsWith( keyToRemove ) )
             {
-                getEHCache().remove( key );
+                boolean tmp = getEHCache().remove( key );
+                removed = removed || tmp;
             }
         }
-        return true;
+        return removed;
```
mah0526, line 121: Are there potential upstream concerns as part of this since the method always (perhaps erroneously) returned true previously?

jdillon, line 121: This method AFACIT in nexus-oss is only used when its return value is ignored.  I didn't find any uses in nexus-pro.  Could potentially be other plugins using it.  Not sure that a constant return value of true would provide much meaning/value to an implementor, and if it was/is being used like it was intended then now that it has correct meaning usage should still be valid.

adreghiciu, line 121: I talked to Tamas about it and he has more insights. He will make a new pull request for this yet we will be able to know if something was actually cleared from NFC or not, and that is what we need.

jdillon:
Closing, replaced by 201

PR #201, opened 2012-01-17
NXCM-3642: expireCache related fixes.

cstamas:
Historically, the two methods (expireCaches and expireNotFoundCaches) were implemented
as one method, they were separated in Nexus 1.4, but the related events fired were never
separated. 

This change fixes that, while keeping backward compatibility completely.
New events introduced, that also carries useful information about actual action performed,
but does not change the "old" semantics, as they are always fired.

This change is about to supersede the change in:
https://github.com/sonatype/nexus/pull/200

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/events/RepositoryEventExpireCaches.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/events/RepositoryEventExpireCaches.java
@@ -18,24 +18,36 @@
  */
 package org.sonatype.nexus.proxy.events;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.sonatype.nexus.proxy.repository.Repository;
 
 /**
- * The event fired on Expiring cache.
+ * The event fired on Expiring Not Found cache.
+ * <p>
+ * Deprecation note: for historical reasons, before Nexus 1.4, there was only one method {@code expireCaches()}. After
+ * Nexus 1.4 we separated out the "purge NFC" part of it into a new method {@code expireNotFoundCaches()}, but the
+ * actual event being fired on those two were never split. Today, we are adding new event but also providing all the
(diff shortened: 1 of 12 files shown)


adreghiciu:
+1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/cache/PathCache.java
```diff
@@ -37,7 +37,7 @@
 
     boolean removeWithChildren( String path );
 
-    void purge();
+    boolean purge();
```
jdillon, line 40: This breaks binary compatibility, are we okay with this?  Probably fine for a 2.0.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/events/RepositoryEventExpireCaches.java
```diff
@@ -18,24 +18,36 @@
...
  */
-public class RepositoryEventExpireCaches
+public abstract class RepositoryEventExpireCaches
     extends RepositoryMaintenanceEvent
 {
     /** From where it happened */
     private final String path;
 
-    public RepositoryEventExpireCaches( final Repository repository, String path )
+    protected RepositoryEventExpireCaches( final Repository repository, final String path )
```
jdillon, line 47: There are also breaking changes, unless we assume that no one should ever be constructing one of these.  Which I think is probably safe to assume.

jdillon:
So I guess this is okay, though it is much bigger scope change than what is in pull/200 and so late in the release cycle, has me a bit uneasy.

You should include the UT improvements in pull/200 to validate the change to EhCachePathCache, and probably include a simply validation for new purge w/boolean return.

nabcos:
+1

velo:
+1

cstamas:
Did local full build (with ITs) and all went okay, merging in.

PR #202, opened 2012-01-17
Reconfigure logging when nexus is initialized as there can be plugins th...

adreghiciu:
...at can contribute configurations that are not initially visible

It did work before for migration plugin as migration plugin was doing its own reconfiguration call when initialized


adreghiciu:
Started build https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/168/ to check

cstamas:
:cake:

velo:
+1

jdillon:
Does this change allow LogConfigurationParticipant components in plugins to function?

adreghiciu:
yes

adreghiciu:
CI voted +1

PR #203, opened 2012-01-17
[NEXUS-4702] rename logback-default and add "Do not touch" headers

adreghiciu:


nabcos:
maybe also write 'edit logback-nexus.xml instead'?

+1 otherwise

adreghiciu:
That was the text suggested by rich. I copy pasted.

cstamas:
:kiss:

adreghiciu:
Started feature build https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/170/

adreghiciu:
The CI build gave some errors but they do not seem related to log names changes, so I consider it a +1

PR #208, opened 2012-01-19
NEXUS-4218: NXCM-3698: fix

cstamas:
The implementation of NEXUS-4218 was buggy, and was triggered by
userAgent-less invocation by smart proxy causing it to download (in case
of proxy repository) the file not being present in cache, ie. to pre-populate
the proxy cache.

This was never detected before SP, because invoking "expireCaches" method was
possible only over UI, and you could not invoke it against paths not present in
cache -- since UI would not represent them.

So, problems:
1) the "version hack" workaround was invoked in many more cases than needed (when client is old Maven)
2) the code was buggy causing to loose the "localOnly=true" flag from context, ending up with doing real remote fetches
3) code used too low level methods, circumventing UID locking completely

Unsure is deadlock fixed by this, I think not (the actual culprit), but the case that causes is will never happen (code was doing remote fetch and acquiring exclusive lock).


peterlynch:
are there unit tests to write?

adreghiciu:
+0 I cannot say if code is good or bad as I do not know what is supposed to do. looks okay :D

cstamas:
Locally validated, merging it in.

PR #209, opened 2012-01-19
Nexus 4611 it

velo:
Current logs saying plugin is trying to connect to localhost:80... this port is no available on linux.

Made sure port is over 1024

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
+++ b/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
@@ -59,10 +59,12 @@ public abstract class AbstractStagingMojo
         getLog().info( "User: " + getUsername() );
 
         ProxyConfig proxyConfig = null;
-        if ( getProxyHost() != null )
+        final String proxyHost = getProxyHost();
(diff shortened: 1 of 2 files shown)


adreghiciu:
+1

peterlynch:
+1 after log message and ProxyConfig constructor use same values.

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -61,6 +61,7 @@ protected final synchronized StageClient connect()
         ProxyConfig proxyConfig = null;
         if ( getProxyHost() != null )
         {
+            getLog().info( "Proxying nexus through: " + getProxyHost() + ":" + getProxyPort() );
             proxyConfig = new ProxyConfig( getProxyHost(), getProxyPort(), getProxyUsername(), getProxyPassword() );
```
peterlynch: Perhaps call getProxyHost and getProxyPort once only - else you risk the logged message masking a different version that is actually used. Potential future bug should someone change what getProxyPort and getProxyHost implementations.

velo: done

PR #212, opened 2012-01-20
o [NXCM-3688] Add config option for decoding html entities to TextField

nabcos:
Characters mapping to html entities are escaped server-side on incoming REST requests. The value received from server would render as expected, but the text field setting the value would contain the html escaped string instead of the 'literal' value, and would also send the html string on the next save - just to be processed again by the REST layer.

This change adds a config option to decode html entities for text field values, and applies this option to the repository name field.

### DIFF
--- a/nexus/nexus-webapp/src/main/webapp/js/extensions/Ext.form.js
+++ b/nexus/nexus-webapp/src/main/webapp/js/extensions/Ext.form.js
@@ -983,22 +983,59 @@ Ext.override(Ext.form.TextField, {
    * <tt>true</tt> to decode html entities in the value given to
    * Ext.form.ByteDisplayField.setValue and Ext.form.ByteDisplayField.setRawValue
    * before setting the actual value.
+   * <p/>
+   * This is needed for displaying the 'literal' value in the text field when it was received by the server,
+   * for example in the repository name. The REST layer will encode to html entities, which will be correct
+   * for html rendering, but text fields without this configuration will display '&quot;test&quot;' instead
+   * of the originally sent '"test"'.
    */
   htmlDecode : false,
 
+  /**
+   * @cfg {Boolean} htmlConvert
+   * <tt>true</tt> to decode html entities in the value given to
+   * Ext.form.TextField.set(Raw)Value
+   * before setting the actual value, and encode html entities again
+   * in the call to Ext.form.TextField.get(Raw)Value.
+   * <p/>
+   * This is needed for displaying the 'literal' value in the text field when it was received by the server
+   * (see htmlDecode configuration doc), and display to the user correctly before round-tripping to the server again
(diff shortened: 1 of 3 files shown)


velo:
+1

adreghiciu:
@nabcos This does not seem very related to NXCM-3688 beside the fact that we are talking about escape/decode, right?

+1 for the change but is strange that is associated with NXCM-3688

nabcos:
@adreghiciu Yes it's the foundation for the [change in pro](https://github.com/sonatype/nexus-enterprise/pull/103) for NXCM-3688, but is useful in it's own right (exactly for fields like repo name, that suffered from value change on consecutive save as well).

peterlynch:
@nabcos how backward compatible is this change for repository names saved in previous nexus versions

nexus/nexus-webapp/src/main/webapp/js/extensions/Ext.form.js
```diff
@@ -976,3 +976,67 @@ Ext.form.ByteDisplayField = Ext.extend(Ext.form.DisplayField, {
     });
 
 Ext.reg('byteDisplayField', Ext.form.ByteDisplayField);
+
+Ext.override(Ext.form.TextField, {
```
bdemers, line 980: Do we need to do something different for Tables? or are the cells rendered as TextFields ?

nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.RepoEditPanel.js
```diff
@@ -432,6 +432,7 @@ Sonatype.repoServer.HostedRepositoryEditor = function(config) {
               helpText : ht.name,
               name : 'name',
               width : 200,
+              htmlDecode : true,
```
bdemers, line 435: Why only this field ?

nabcos:
@peterlynch This pull does not change how data that is received from the server is presented. So if you use nexus 1.9 to set a repo name to `Kings&Things`, UI would receive and render the value `Kings&amp;Things` in the repository table (which displays the name correctly).

The difference is in the configuration form, in 1.9 you would now see `Kings&amp;Things` in the name TextField. When the repository is saved, nexus would encode this value again, and UI would receive and render the value `Kings&amp;amp;Things` - which is incorrect. With this pull request, you would see `Kings&Things` in the TextField, which would then be encoded on the server side and display correctly on the following saves.

nabcos:
@bdemers UI will be correct everywhere where the first save does the right thing. Special case for e.g. editable grids is when values are displayed without them having a round-trip to the server before, this would mean using htmlConvert and another htmlDecode before the values are sent.

bdemers:
AHH, I see, thanks for clearing that up
+1

PR #213, opened 2012-01-20
UT logging cleanup.

cstamas:
As nexus-test-common started bringing in logback, which is a SLF4J backend,
logging in UTs were borked as slf4j-simple was added too with test scope
and SLF4J complains about multiple backends present on classpath.

nexus-test-common brings in logback as transitive dep coming from litmus.

Now, the slf4j-simple is removed, only one module is left using it (that
rightfully does not use nexus-test-common): nexus-web-utils module.


adreghiciu:
+1

nabcos:
+1

PR #219, opened 2012-01-24
NEXUS-4786: Scheduled task logging fix

cstamas:
This is actually a followup of the "move timeline into plugin" change.
Before, while Timeline was in core, ScheduledTasks relied on it, and was coupled
to it to perform "task related logging" (report when task started, stopped, etc).
Today, Timeline plugin is demoted into a plugin, and is hence optional, this
change actually fixes the problem of potentially missing task related loglines,
but also corrects the misleading loglines coming from Feed recorder.

Changes:
- DefaultFeedRecorder task related logging demoted to DEBUG level (IMO, should be totally removed)
- AbstractNexusTask improved for small "bookkeepeing" about run duration and proper logging about it's
  execution status

This change actually "fixes" the problem forgot to be solved in the actual move of the Timeline into a plugin.


adreghiciu:
+1

nabcos:
+1

velo:
+1

PR #220, opened 2012-01-24
Improvements to repository base classes

cstamas:
Improvements wrt logging happening from Repository instances (three main abstract classes).

Also, Javadoced and un-deprecated the expireCaches() method while explained what those methods
(might) do.

After this change, expireCaches method nicely fits, as major entry point that follows OO
polymorphism correctly in all cases, without seeding confusion.


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
```diff
@@ -618,18 +632,6 @@ public void maintainNotFoundCache( ResourceStoreRequest request )
     }
 
     @Override
-    public void expireCaches( ResourceStoreRequest request )
```
velo, line 621: Any reason to move this up?
https://github.com/sonatype/nexus/pull/220/files#L2R140

Just wanna keep easier to review/blame =D

cstamas, line 621: Only reason of the move (method is unchanged) to logically group (make them closer to each other) some methods. Makes code reading easier, no need for lot of scrolling. Nothing else.

velo:
+1

nabcos:
+1

bdemers:
+1

PR #221, opened 2012-01-24
NEXUS-4770: Multiple fixes for Walker implementation.

cstamas:
Major issue reported in jira above was the presence of reportWalkEnd() method
in the catch of the inner block, that throws _wrapped_ exception that was catched
by the outer catch block. This was resulting in Walker doubling the reports of
aborting the walk. The irony is that in caller level (in this case, recreateAttributes method of AbstractRepository)
was properly handling this. It was only about improper logging -- spam.

Also, logging was improved, since by recent changes, this component got real Slf4j logger.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
@@ -53,9 +53,10 @@ public class DefaultWalker
         }
 
         // cannot walk out of service repos
-        if( LocalStatus.OUT_OF_SERVICE == context.getRepository().getLocalStatus() )
+        if ( LocalStatus.OUT_OF_SERVICE == context.getRepository().getLocalStatus() )
         {
-            getLogger().info( "Cannot walk, repository: '"+ context.getRepository().getId() + "' is out of service." );
+            getLogger().info( "Cannot walk, repository {} is out of service.",
+                RepositoryStringUtils.getHumanizedNameString( context.getRepository() ) );
         }
         else
         {
@@ -63,8 +64,8 @@ public class DefaultWalker
 
             if ( getLogger().isDebugEnabled() )
             {
-                getLogger().debug(
-                    "Start walking on ResourceStore " + context.getRepository().getId() + " from path '" + fromPath + "'." );
+                getLogger().debug( "Start walking on ResourceStore {} from path \"{}\".",
+                    RepositoryStringUtils.getHumanizedNameString( context.getRepository() ), fromPath );
             }
 
             try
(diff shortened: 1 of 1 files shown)


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
```diff
@@ -53,9 +53,10 @@ public void walk( WalkerContext context )
         }
 
         // cannot walk out of service repos
-        if( LocalStatus.OUT_OF_SERVICE == context.getRepository().getLocalStatus() )
+        if ( LocalStatus.OUT_OF_SERVICE == context.getRepository().getLocalStatus() )
         {
-            getLogger().info( "Cannot walk, repository: '"+ context.getRepository().getId() + "' is out of service." );
+            getLogger().info( "Cannot walk, repository {} is out of service.",
+                RepositoryStringUtils.getHumanizedNameString( context.getRepository() ) );
```
velo, line 59: I though there was a repository.getName()....

velo:
+1

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
```diff
@@ -137,29 +131,25 @@ protected void reportWalkEnd( WalkerContext context, String fromPath )
         {
             if ( context.getStopCause() == null )
             {
-                getLogger().debug( "Walker was stopped programatically, not because of error." );
+                if ( getLogger().isDebugEnabled() )
```
bdemers, line 134: no logic or method calls in body, consider dropping isDebugEnabled() check

bdemers:
+1

nabcos:
Hm, IIUC the other problem reported in NEXUS-4770 is still there, which is that the repository will be marked as `isUpgraded` for an ItemNotFoundException and thus will never be completely upgraded, because the upgrader will ignore it after that.

cstamas:
Yes, the actual culprit is not yet fixed, but issue NEXUS-4770 as far as I understand it, is questioning the log level and the doubling of the log lines too... that part is fixed by this pull req.

nabcos:
ok, I just wanted to make sure that this part does not get lost:
`I didn't notice the attributes upgrader ever try to rerun itself - is that expected? 
Can it be a little more forgiving?`

cstamas:
Wrt that quoted sentence.... it cannot be "forgiving", since ItemNotFoundEx is thrown exactly because there is NOTHING to upgrade... but this discussion should be continued on other (related) pull req :)

nabcos:
Well, the problem is that an INFE somewhere inside the repo will stop the whole upgrading process, so that one item should be forgiven ;)

cstamas:
Not quite: the point is that this INFE happens during upgrade of attributes running against a _group repository_, that in essence, (should) have no content whatsoever (yes, I know we violate this by saving checksums, but after having many phones about this, I am convinced checksums actually should NOT be saved either). But again, this pulls some other implications in the picture too...

So this knot still needs some love :D

PR #223, opened 2012-01-24
NEXUS-4774, reduce logging levels for walker, expire NFC and proxy cache

bdemers:


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
```diff
@@ -146,11 +146,22 @@ else if ( context.getStopCause() instanceof TaskInterruptedException )
...
+                            context.getStopCause().getMessage() } );
+                }
+                else
+                {
+                    getLogger().debug(
+                        "Aborted walking on repository {} from path \"{}\", cause: {}",
+                        new Object[] { RepositoryStringUtils.getHumanizedNameString( context.getRepository() ), fromPath,
+                            context.getStopCause().getMessage() } );
+                }
+
```
jdillon, line 164: I don't see why this branch on instanceof ItemNotFoundException is needed... both logs are the same.

cstamas:
Strange wording: "Attempting to expire proxy cache...". It does not attempt to do it, it does it, but it depends did it alter anything or not.

The older text replaced with "attempting" was better IMO: "Expiring proxy cache" and then "Proxy cache was expired" (if cache alteration happened) or "Proxy cache not altered" (no cache alteration happened).

ie. Google is not "attempting to search" either... it says "searching...".

Also, -1 to lowering the log level to debug on two "attempting" lines, it looks totally targeted at SP (and change inspired by it). We should not be "blinded" by SP, a user running a "standalone" (non-SP) instance of Nexus would be very glad to see these log lines, just as he did before IMO.

cstamas:
...or, as noted in NEXUS-4774, make the log that resulted in cache alteration INFO level at least.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
bdemers, line 164: your right, the else was suppose to be .info() thanks

cstamas:
:cake:

nabcos:
+1

bdemers:
@mhansen-sonatype the thinking is that most of the log spam is happening when the NFC or proxy cache is not cleared.  With this change only log info messages when the cache has changed

PR #226, opened 2012-01-25
Making TransitioningAttributeStorage "non-eager", throttling et al.

cstamas:
Attribute upgrade from now on happens only on attribute STORE operation,
while attribute READ operation just _reads_ them (from new or falls back to old
storage).

Just for review, not for vote yet. This change alone would not be sufficient for sure, coz some tests expect the "old" behavior.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/WalkerThrottleController.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/WalkerThrottleController.java
@@ -1,0 +1,53 @@
+package org.sonatype.nexus.proxy.walker;
+
+/**
+ * Interface to controle throttling (if desired) in executing of Walks.
+ * 
+ * @author cstamas
+ * @since 2.0
+ */
+public interface WalkerThrottleController
+{
+    /**
+     * Key used in RequestContext to pass it into Walker.
+     */
+    String CONTEXT_KEY = WalkerThrottleController.class.getName();
+
+    /**
+     * WalkerThrottleController that does not emply walk throttling.
+     */
+    WalkerThrottleController NO_THROTTLING = new WalkerThrottleController()
+    {
+        @Override
+        public boolean isThrottled()
+        {
+            return false;
+        }
+
+        @Override
+        public long throttleTime( long processItemSpentMillis )
+        {
+            return -1;
+        }
+    };
+
+    /**
+     * Returns true if the controllers wants to use "throttled walker" execution. Throttling in this case would mean
+     * intentionally slowing down the "walk" by inserting some amount (see {@link #throttleTime(long)} method) of
+     * Thread.sleep() invocations.
+     * 
+     * @return
+     */
+    boolean isThrottled();
+
+    /**
+     * Returns the next desired sleep time this context wants to have applied. It might be in some relation to the time
+     * spent in processItem() methods of registered WalkerProcessors, but does not have to be.
+     * 
+     * @param processItemSpentMillis time in millis WalkerProcessors spent in their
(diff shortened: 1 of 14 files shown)


velo:
+1

bdemers:
We should have tests for this
(not sure the best way to test throttling)
other then that it looks good

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/WalkerThrottleController.java
```diff
@@ -0,0 +1,77 @@
+package org.sonatype.nexus.proxy.walker;
+
+/**
+ * Interface to controle throttling (if desired) in executing of Walks.
+ *
```
peterlynch, line 5: sp.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,204 @@
...
+            getLogger().debug( "Legacy attribute directory not present, no need for attribute upgrade." );
+        }
+        else
+        {
+            if ( isUpgradeDone( getLegacyAttributesDirectory(), null ) )
+            {
+                // nag the user to remove the directory
+                getLogger().info(
+                    "Legacy attribute directory present, but is marked already as upgraded. Please delete, move or rename the \""
+                        + getLegacyAttributesDirectory().getAbsolutePath() + "\" folder." );
```
peterlynch, line 132: slf4j logger? could use {} templating

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,204 @@
...
+                final File target = new File( new File( attributesDirectory, repoId ), MARKER_FILENAME );
+                // this step is needed if new repo added while upgrade not done: it will NOT have legacy attributes
+                // as other reposes, that were present in old/upgraded instance
+                target.getParentFile().mkdirs();
+                FileUtils.fileWrite( target, MARKER_TEXT );
+            }
+        }
+        catch ( IOException e )
+        {
+            // hum?
```
peterlynch, line 201: finish that thought?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/UpgraderThread.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public void run()
+    {
+        try
+        {
+            Thread.sleep( 5000 );
+        }
+        catch ( InterruptedException e )
+        {
+            return;
```
peterlynch, line 76: Shouldn't Thread.currentThread().interrupt(); be called here?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/UpgraderThread.java
```diff
@@ -0,0 +1,175 @@
...
+    {
+        this.limiterUps = limiterUps;
+    }
+
+    @Override
+    public void run()
+    {
+        try
+        {
+            Thread.sleep( 5000 );
```
peterlynch, line 72: nit: personally prefer to see

```
TimeUnit.SECONDS.sleep(5);
```

cstamas:
Am closing this pull request for now, to separate it into two pulls, one about substantial fixes (that should go ASAP into master), and one about the "throttling" where we can continue this debate.

PR #233, opened 2012-01-27
Attribute related fixes

cstamas:
This pull request contains following substantial fixes:
- (F1) TransitioningAttributeStorage made less eager. Fix made attribute-related IO ops from "read, read, write, delete" to "read, read" (worst cases).
- (F2) RecreateAttributesWalker operated on groups too, hence, group member repositories was walked for upgrade multiple times.
- (F3) RecreateAttributesWalker was forcing re-checksumming.
- (F4) on upgrade, attribute save happens only when it originates from "legacy" store (ie. partially upgraded repository will perform saves only on non-upgraded attributes).

And some other minor improvements regarding overall IO load of upgrade.

This pull request contains following major improvements:
- Nexus Core Walker component supports walk-throttling, it is possible to "interfere" with walking speed from now on (by slowing it down, not by walking faster than your HW is capable of! ;)
- Two out-of-the-box throttle controllers implemented: "fixed rate" and "adaptive" (the 2nd is TBD, not yet implemented!).
- AttributeUpgrader component exposed over JMX, and properties like maxTPS are even settable live during upgrade. Also, the upgrade process itself can be invoked/started over JMX (ie. you defer upgrade auto-start by system property to be able to set up for proper JMX monitoring).
- NumberSequence got new member: "linear" number sequence, and it's API is enhanced (got prev() method for stepping back).

The "initial" maxTPS value is set to 100 upgrade/sec, which would upgrade Central repository hosted in Nexus in 5 hours. But, again, the maxTPS is settable over JMX.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
@@ -180,9 +180,9 @@ public class FixedRateWalkerThrottleController
     }
 
     @Override
-    public void walkEnded( final WalkerContext context )
+    public void walkEnded( final WalkerContext context, final ThrottleInfo info )
     {
-        // nop
+        calculateStats( info );
     }
 
     @Override
@@ -196,15 +196,7 @@ public class FixedRateWalkerThrottleController
     {
         if ( adjustmentNeeded() )
         {
-            // global average TPS since walk started
-            globalAverageTps =
-                (int) ( info.getTotalProcessItemInvocationCount() / ( info.getTotalTimeWalking() / 1000 ) );
-            // local TPS achieved in current "adjustment" slice
-            lastSliceTps =
-                (int) ( ( info.getTotalProcessItemInvocationCount() - lastAdjustmentProcessItemInvocationCount ) / ( lastSliceSize / 1000 ) );
-            lastAdjustmentProcessItemInvocationCount = info.getTotalProcessItemInvocationCount();
-            // the maximum TPS since the walk started
-            globalMaximumTps = Math.max( lastSliceTps, globalMaximumTps );
+            calculateStats( info );
 
             if ( lastSliceTps > limiterTps )
             {
@@ -238,7 +230,20 @@ public class FixedRateWalkerThrottleController
 
     // ==
 
-    protected boolean adjustmentNeeded( )
+    protected void calculateStats( final ThrottleInfo info )
+    {
+        // global average TPS since walk started
+        globalAverageTps = calculateCPS( info.getTotalProcessItemInvocationCount(), info.getTotalTimeWalking() );
+        // local TPS achieved in current "adjustment" slice
+        lastSliceTps =
+            calculateCPS( info.getTotalProcessItemInvocationCount() - lastAdjustmentProcessItemInvocationCount,
+                lastSliceSize );
+        lastAdjustmentProcessItemInvocationCount = info.getTotalProcessItemInvocationCount();
+        // the maximum TPS since the walk started
+        globalMaximumTps = Math.max( lastSliceTps, globalMaximumTps );
+    }
+
+    protected boolean adjustmentNeeded()
     {
         final long now = System.currentTimeMillis();
         lastSliceSize = now - lastAdjustmentTimestamp;
@@ -251,5 +256,22 @@ public class FixedRateWalkerThrottleController
         {
             return false;
         }
+    }
+
+    protected int calculateCPS( final long count, final long millis )
+    {
+        if ( count == 0 )
+        {
+            return 0;
+        }
+
+        if ( millis == 0 )
+        {
+            return -1; // um, would be division by zero, but since we deal with a measure that might be positive only
+                       // (count-per-seconds), returning -1 is good for now
+        }
+
+        return (int) ( ( (double) count / (double) millis ) * 1000.0d ); // we do this to make measure in seconds, but
+                                                                         // to retain as much precision as possible
     }
 }
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
@@ -82,7 +82,7 @@ public class FixedRateWalkerThrottleController
     /**
      * Internally maintained slice size (in milliseconds), milliseconds betwen two subsequent adjustments.
      */
-    private long lastAdjustmentSliceSize;
+    private long lastSliceSize;
 
     /**
      * Internally maintained processItem invocation count when last adjustment was made.
@@ -92,14 +92,14 @@ public class FixedRateWalkerThrottleController
     /**
      * The TPS measured in last adjustment "slice".
      */
-    private int lastAdjustmentTps;
+    private int lastSliceTps;
 
     /**
      * How often (in milliseconds) should adjustment to throttling be made. To small value might cause unneded overhead
      * and "oscillation", to big value might cause that throttling becomes not effective (ie. walk will spend too much
      * time in "overspeed" before adjustment kicks in).
      */
-    private long adjustmentFrequency;
+    private long sliceSize;
 
     /**
      * Creates a new instance of fixed rate throttle controller with some defaults (adjustment slice size is 2 seconds).
@@ -124,7 +124,7 @@ public class FixedRateWalkerThrottleController
         this.limiterTps = limiterTps;
         this.currentSleepTime = numberSequence;
         this.callback = callback;
-        this.adjustmentFrequency = TimeUnit.SECONDS.toMillis( 2 );
+        this.sliceSize = TimeUnit.SECONDS.toMillis( 2 );
     }
 
     public int getGlobalAverageTps()
@@ -137,9 +137,9 @@ public class FixedRateWalkerThrottleController
         return globalMaximumTps;
     }
 
-    public int getLastAdjustmentTps()
+    public int getLastSliceTps()
     {
-        return lastAdjustmentTps;
+        return lastSliceTps;
     }
 
     public long getCurrentSleepTime()
@@ -157,15 +157,15 @@ public class FixedRateWalkerThrottleController
         this.limiterTps = limiterTps;
     }
 
-    public long getAdjustmentFrequency()
+    public long getSliceSize()
     {
-        return adjustmentFrequency;
+        return sliceSize;
     }
 
-    public void setAdjustmentFrequency( final long adjustmentFrequency )
+    public void setSliceSize( final long sliceSize )
     {
-        Preconditions.checkArgument( adjustmentFrequency > 0 );
-        this.adjustmentFrequency = adjustmentFrequency;
+        Preconditions.checkArgument( sliceSize > 0 );
+        this.sliceSize = sliceSize;
     }
 
     // == WalkerThrottleController iface
@@ -174,7 +174,7 @@ public class FixedRateWalkerThrottleController
     {
         globalAverageTps = 0;
         globalMaximumTps = 0;
-        lastAdjustmentTps = 0;
+        lastSliceTps = 0;
         lastAdjustmentTimestamp = System.currentTimeMillis();
         lastAdjustmentProcessItemInvocationCount = 0;
     }
@@ -200,13 +200,13 @@ public class FixedRateWalkerThrottleController
             globalAverageTps =
                 (int) ( info.getTotalProcessItemInvocationCount() / ( info.getTotalTimeWalking() / 1000 ) );
             // local TPS achieved in current "adjustment" slice
-            lastAdjustmentTps =
-                (int) ( ( info.getTotalProcessItemInvocationCount() - lastAdjustmentProcessItemInvocationCount ) / ( lastAdjustmentSliceSize / 1000 ) );
+            lastSliceTps =
+                (int) ( ( info.getTotalProcessItemInvocationCount() - lastAdjustmentProcessItemInvocationCount ) / ( lastSliceSize / 1000 ) );
             lastAdjustmentProcessItemInvocationCount = info.getTotalProcessItemInvocationCount();
             // the maximum TPS since the walk started
-            globalMaximumTps = Math.max( lastAdjustmentTps, globalMaximumTps );
+            globalMaximumTps = Math.max( lastSliceTps, globalMaximumTps );
 
-            if ( lastAdjustmentTps > limiterTps )
+            if ( lastSliceTps > limiterTps )
             {
                 // hold down the horses, increase sleepTime
                 if ( currentSleepTime.peek() <= 0 )
@@ -241,9 +241,8 @@ public class FixedRateWalkerThrottleController
     protected boolean adjustmentNeeded( )
     {
         final long now = System.currentTimeMillis();
-        lastAdjustmentSliceSize = now - lastAdjustmentTimestamp;
-        // adjust every 5 second for now
-        if ( ( lastAdjustmentSliceSize ) > adjustmentFrequency )
+        lastSliceSize = now - lastAdjustmentTimestamp;
+        if ( ( lastSliceSize ) > sliceSize )
         {
             this.lastAdjustmentTimestamp = now;
             return true;
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
@@ -102,26 +102,39 @@ public class FixedRateWalkerThrottleController
     private long sliceSize;
 
     /**
-     * Creates a new instance of fixed rate throttle controller with some defaults (adjustment slice size is 2 seconds).
+     * Creates a new instance of fixed rate throttle controller without callback and some defaults (adjustment slice
+     * size is 2 seconds).
      * 
      * @param limiterTps
      * @param numberSequence
+     * @throws IllegalArgumentException If passed in numberSequence is not having needed characteristics.
      */
     public FixedRateWalkerThrottleController( final int limiterTps, final NumberSequence numberSequence )
+        throws IllegalArgumentException
     {
         this( limiterTps, numberSequence, null );
     }
 
     /**
-     * Creates a new instance of fixed rate throttle controller with some defaults (adjustment slice size is 2 seconds).
+     * Creates a new instance of fixed rate throttle controller with callback and some sensible defaults (adjustment
+     * slice size is 2 seconds).
      * 
-     * @param limiterTps
-     * @param numberSequence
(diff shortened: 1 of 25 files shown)


cstamas:
CI put it's +1 (feature IT build):
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/181/

The reason of build failure was actually missing headers from newly added sources, fixed in 9e0c13f439b3996372f5c3b69f73303a26c48709. Headers are checked after all the tests were passing.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
```diff
@@ -0,0 +1,277 @@
...
+    public long throttleTime( final ThrottleInfo info )
+    {
+        if ( adjustmentNeeded() )
+        {
+            calculateStats( info );
+
+            if ( lastSliceTps > limiterTps )
+            {
+                // hold down the horses, increase sleepTime
+                if ( currentSleepTime.peek() <= 0 )
```
mah0526: If I'm looking at this correctly, with a starting delay (NumberSequence) of 0, it seems you will never slow down -- i.e. you will never end up calling currentSleepTime.next().  If so, we should probably handle differently.

cstamas: :beer: X1

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
```diff
@@ -0,0 +1,171 @@
...
+            {
+                // mark instance as upgraded
+                DefaultAttributeUpgrader.markUpgradeDone( legacyAttributesDirectory, null );
+                logger.info(
+                    "Legacy attribute directory upgrade finished without any errors. Please delete, move or rename the \"{}\" folder.",
+                    legacyAttributesDirectory.getAbsolutePath() );
+            }
+            else
+            {
+                logger.info( "Legacy attribute directory upgrade PARTIALLY finished. Please fix the problems and retry the upgrade." );
```
mah0526: In the case of a problem with the upgrade where a WalkerException was thrown, that may be the case but the note about "fix the problems" should probably go in that log message (keeping the pertinent information for the user in one place), and perhaps logged as at ERROR (versus WARN).  Additionally, for upgrades on repos out of service that are not performed, it might be best just to tell the user that it will automatically be done at a later time (currently, on server restart I assume).  With those changes, then this message merely needs to note that the "...upgrade was partially completed.  Please see prior log messages for details."  There may not actually be any problems, and if there are, we should indicate to the user where they should start looking.  Of course, most users will probably never see this stuff anyway.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void onAdjustment( final FixedRateWalkerThrottleController controller )
+    {
+        logger.info(
```
mah0526: Probably better to have this at DEBUG level for final release.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    @Override
+    public void upgradeAttributes()
+    {
+        upgradeAttributes( false );
+    }
+
+    @Override
+    public synchronized void upgradeAttributes( final boolean force )
```
mah0526, line 206: What's the least amount of locking required for correctness?  Will this method acquire the dreaded configuration lock through the getLegacyAttributesDirectory() call?  If so and you need sync on entire method, you might want to make the external calls first and then lock.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,296 @@
...
+                final File target = new File( new File( attributesDirectory, repoId ), MARKER_FILENAME );
+                // this step is needed if new repo added while upgrade not done: it will NOT have legacy attributes
+                // as other reposes, that were present in old/upgraded instance
+                target.getParentFile().mkdirs();
+                FileUtils.fileWrite( target, MARKER_TEXT );
+            }
+        }
+        catch ( IOException e )
+        {
+            // hum?
```
mah0526: Seems like we ought to log something here, as unlikely as the case may be.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
cstamas, line 206: No, this is the least locking needed, to retain correctness only. This method is synchronized _only_ to ensure that it will not spawn multiple upgrade thread over each other.

Also, if you look at the method body, you will notice that the time spent here is minimal (best case logging something and getting out), worst case creating a thread and starting it.

This synchronization is needed because the method can be invoked over JMX. Actually, the "execution path" that brings you here to this method is either the NexusStartedEvent (guaranteed to happen only once, and it's not the reason of syncing), _or_ it's been fired over JMX. And the JMX path is the one needing sync (2nd incoming call will simply "bounce off", if the thread is be alive and running).

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
cstamas: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
cstamas: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
cstamas: Fixed in subsequent commit.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/FixedRateWalkerThrottleController.java
cstamas: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultThrottleInfo.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.sonatype.nexus.proxy.walker.WalkerThrottleController.ThrottleInfo;
+
+/**
+ * A simple single-threaded ThrottleInfo used in Walker implementation.
+ * 
+ * @author cstamas
+ * @since 2.0
+ */
+public class DefaultThrottleInfo
```
mah0526, line 23: No use for a lastProcessTimeSpent() call?  Not sure if it is being calculated externally.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,295 @@
...
+import org.sonatype.nexus.logging.AbstractLoggingComponent;
+import org.sonatype.nexus.proxy.registry.RepositoryRegistry;
+import org.sonatype.nexus.util.SystemPropertiesHelper;
+
+/**
+ * Component that handles upgrade of "legacy attribute storage". It does it by detecting it's presence, and firing the
+ * rebuild attributes background task if needed. Finally, it leaves "marker" file to mark the fact upgrade did happen,
+ * to not kick in on any subsequent reboot.
+ * 
+ * @since 1.10.0
```
peterlynch: wrong since

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
```diff
@@ -0,0 +1,205 @@
...
+import org.sonatype.nexus.proxy.repository.RecreateAttributesWalker;
+import org.sonatype.nexus.proxy.repository.Repository;
+import org.sonatype.nexus.proxy.utils.RepositoryStringUtils;
+import org.sonatype.nexus.proxy.walker.FixedRateWalkerThrottleController;
+import org.sonatype.nexus.proxy.walker.WalkerException;
+import org.sonatype.nexus.proxy.walker.WalkerThrottleController;
+import org.sonatype.nexus.proxy.walker.FixedRateWalkerThrottleController.FixedRateWalkerThrottleControllerCallback;
+import org.sonatype.nexus.util.FibonacciNumberSequence;
+
+public class AttributeUpgraderThread
```
velo, line 42: JDoc missing here

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgraderMBean.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.proxy.attributes.upgrade;
+
+import javax.management.StandardMBean;
+
+public class DefaultAttributeUpgraderMBean
```
velo, line 23: JDoc

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalker.java
```diff
@@ -315,6 +323,18 @@ protected void processItem( WalkerContext context, StorageItem item )
...
                 }
             }
+
+            info.exitProcessItem();
+
+            if ( !context.isStopped() && context.getThrottleController().isThrottled() )
+            {
+                final long throttleTime = context.getThrottleController().throttleTime( info );
+
+                if ( throttleTime > 0 )
```
mah0526, line 333: Only a thought... Worthwhile to encapsulate the Thread.sleep() call within the throttle controller classes?  This class probably need not care about those details (like having to get the time back and deal with it).  Perhaps refactor throttleTime( info ) to throttle( info ) and handle everything at a lower level of abstraction.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
```diff
@@ -0,0 +1,295 @@
...
+            // instance)
+            getLogger().debug( "Legacy attribute directory not present, no need for attribute upgrade." );
+        }
+        else
+        {
+            if ( isUpgradeFinished() )
+            {
+                // nag the user to remove the directory
+                getLogger().info(
+                    "Legacy attribute directory present, but is marked already as upgraded. Please delete, move or rename the \"{}\" folder.",
```
peterlynch: folder -> directory 

... use one term or the other

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/FibonacciNumberSequence.java
```diff
@@ -36,15 +38,20 @@ public FibonacciNumberSequence()
...
+    }
 
+    public FibonacciNumberSequence( long startA, long startB )
+    {
+        this.startA = startA;
+        this.startB = startB;
         reset();
     }
 
     public long next()
```
mah0526, line 51: This logic looks flawed (though not introduced by this commit).  A call to next() should (I think) return the next number in the sequence.  However, it returns the current value in the sequence and advances "a" to the next number in the sequence.  Only a subsequent call to peek() will get the real next value.

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/FibonacciNumberSequence.java
```diff
@@ -23,7 +23,9 @@
 public class FibonacciNumberSequence
     implements NumberSequence
 {
-    private final long start;
+    private final long startA;
```
mah0526, line 26: How about refactoring "startA" to "currentValue" and "startB" to "precedingValue"?  In the constructor, startA and startB could be replaced with startValue, precedingValue.  That would probably add a lot of clarity.

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LinearNumberSequence.java
```diff
@@ -0,0 +1,66 @@
...
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.util;
+
+/**
+ * A simple linear number sequence (linear equation).
+ * 
+ * @author cstamas
+ */
+public class LinearNumberSequence
```
mah0526, line 20: Wait, this is a discontinuous, non-linear function :)

cstamas, line 20: you mean "discrete function"? ;)

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
mah0526, line 206: You can't make the calls that call out of this class (i.e. into config) before acquiring the lock?  I'll submit that logging calls are not an issue, but even if this is not a problem with the current implementation, it leaves a potential lingering issue for someone (with potentially less knowledge) to contend with down the road.

bdemers:
should add a test for broken attributes durring upgrade, and the case you brought up about three consecutive IOExceptions

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
cstamas, line 206: This class is about attribute upgrade. It does not modify config. But it's possible I am missing your point...

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/walker/DefaultThrottleInfo.java
cstamas, line 23: No use, but it might get used in near future, or we can easily drop it if it proves un-useful (after this is merged but before release)

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
cstamas: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgraderMBean.java
cstamas, line 23: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/AttributeUpgraderThread.java
cstamas, line 42: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
cstamas: Fixed in subsequent commit.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/upgrade/DefaultAttributeUpgrader.java
mah0526, line 206: No problem.  For example, the calls to getLegacyAttributesDirectory() then call applicationConfiguration.getWorkingDirectory(), diving into the internals of Nexus configuration.

cstamas, line 206: Bad named class: almost all applicationConfiguration's methods are about "meta configuration" actually -- about various folders and their whereabout on disk, like nexus/conf folder is (this statement is a bit shaded with the presence of method saveConfiguration()). 

What you think is probably NexusConfiguration (that do extend ApplicationConfiguration)... that one DO expose things that could _modify_ configuration.

Again, getWorkingDir() method (or actually any other code within body of the synchronized method) does _not_ change configuration (nor does the spawned thread either).

cstamas:
By merging (and hence, closing) this pull request initial intention was to remove any obstacle in doing M4 internal release. It does NOT mean I closed my ears for receiving more scrutinizing remarks :)

Unresolved issues in this pull requests was all Mike's:
- potential factoring Thread.sleep() into "lower" abstraction levels in DefaultWalker -- worthwhile to consider
- locking/synchronized related questions -- non issue from my viewpoint
- logical inconsistencies in Fib number sequence, merely cosmetic (variable rename), the other one I'd skip (changing the next() method semantic, since other important part of Nexus (auto blocking) also relies on Fib sequence as is (even if it "looks" broken, this is how it was implemented).

cstamas:
... and the "adaptive controller" implementation.

mah0526:
+1,  oh wait, this looks closed.  go drink :beer:

PR #236, opened 2012-01-27
NEXUS-4622, adding css and javascript media types to the ContentPlexusResource.

bdemers:


peterlynch:
+1

cstamas:
This is clearly a hack, and is actually a deficiency of Restlet or our plexus-restlet integration (i'd put bets on the latter) that would allow proper content negotiation. With this change, we are saying _what the actual content could be_ in advance, clearly, it could be anything.

+1

bdemers:
@cstamas I agree

PR #243, opened 2012-01-30
o [NEXUS-4778] Wait for async events after reindex tasks

nabcos:
Nothing obvious wrong with the failing test, so this is just taking a guess.


nexus/nexus-test-harness/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/utils/SearchMessageUtil.java
```diff
@@ -654,6 +654,8 @@ public void reindexPath( String repo, String path )
 
         // let s w8 a few time for indexes
         TaskScheduleUtil.waitForAllTasksToStop();
+        // be safe and wait for async events as well
+        new EventInspectorsUtil( RequestFacade.getNexusRestClient() ).waitForCalmPeriod();
```
velo, line 658: IIRC we agreed on NOT doing... the general feeling would be that could hide some problem.  but I'm +1

peterlynch:
+1

bdemers:
+1

PR #245, opened 2012-01-30
Change @since 1.10.0 -> @since 2.0

jdillon:
There is no 1.10.0 release, update to use 2.0


nabcos:
+1

bdemers:
+1

bdemers:
there are also a bunch of unneeded @since tags in there... but that is a different story

PR #250, opened 2012-02-01
Boot simplification

cstamas:
Simplifies boot process, and also lessens memory consumption (by not creating a ClassWorld at place where it is unused anyway).

We should leave out Plexus Classworlds from the "top" of boot hierarchy,
as we do not gain anything with it, only complicate things.

This change is for eyes only, would break today's OSS ITs and Pro bundle!


velo:
Then -1 =D

peterlynch:
Put on whatever branch 2.0 is NOT built from.

adreghiciu:
+1 for the change. But I guess this will go in only after 2.0

cstamas:
Closing it for now...

PR #251, opened 2012-02-01
Expose AppContext as component

cstamas:
Expose AppContext as component, simple as that.

### DIFF
--- a/nexus/nexus-web-utils/src/main/java/org/sonatype/nexus/web/AppContextModule.java
+++ b/nexus/nexus-web-utils/src/main/java/org/sonatype/nexus/web/AppContextModule.java
@@ -1,3 +1,15 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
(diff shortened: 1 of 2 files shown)


cstamas:
Rationale: stop using System Properties directly!

As of Nexus 2.0, Core does not use System Properties directly anymore (except for tweaking). This makes "collocation" of multiple Nexus instances in same JVM possible (yummie).

We should stop using system properties, and delegate to AppContext instead, and in case of "collocation", it would enable to have "targeted tweaks". AppContext already implements various sources for accepting context values (properties file, system property prefixed with `contextId.`, environment variable prefixed with `CONTEXTID_` etc). This makes super-easy to affect temporarily or for good targeted instances in single JVM.

peterlynch:
Is this need to solve some issue affecting 2.0 or provide some other support for the 2.0 release that would be a bad idea to add later for some reason? No issue associated with this.

We already have at least one issue open against looking up components - I am hesitant to risk destabilizing things even more.

Maybe time to branch off Nexus 2.0...

nabcos:
I would just let the pull request sit until 2.0 is out the door and apply after that, instead of merging into a 2.1 branch.

PR #252, opened 2012-02-01
Updated about page to accept contributions from plugins

dbradicich:
Simply fires event to capture any changes to the about html that a plugin may want to make


velo:
+1

bdemers:
+1 the change looks fine, just make sure it works with PRO, a

dbradicich:
yes definitely validated with pro, pro simply overrides the getHelpText() method, so the existing use is covered fine

bdemers:
:+1:

On Wed, Feb 1, 2012 at 12:39 PM, Damian Bradicich
reply@reply.github.com
wrote:

PR #258, opened 2012-02-02
o [NEXUS-4809] Consume released jetty-testsuite

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/189/
https://builds.sonatype.org/job/nexus-oss-its-feature/192/


adreghiciu:
Just remind you that you will not go through with a SNAPSHOT dependency. So, you will have to make a release to be able to consume it and merge to master.
I do not agree with that, but seems that is the policy lately. In my view, as Tamas says for the projects we control we should be able to use SNAPSHOTS for a finite amount of time so we do not release until we validate it.

bdemers:
+1 pending using a  released version

nabcos:
CI voted sort-of +1 (intermittent failure on lnx, win built ok)

velo:
+1

PR #268, opened 2012-02-02
Update to app-lifecycle 1.6

jdillon:


cstamas:
:beer:

nabcos:
+1

velo:
Are we sure that is safe just a week before release?

+1 after CI says it is all OK

bdemers:
+1

jdillon:
@velo This is a fairly safe change.  Currently nexus-oss uses version 1.5, and nexus-pro uses 1.4, 1.6 only adds additional configuration for pluginName, pluginDescription and pluginSiteUrl.

PR #269, opened 2012-02-02
Improved logging of task failures.

cstamas:
- Raised level to WARN as we deal with an unexpected failure
- by default, we show exception message only, but if logger in DEBUG mode, we spit all of the stack trace


cstamas:
Related to:
https://issues.sonatype.org/browse/INSIGHT-2188

bentmann:
+1, thanks

mah0526:
+1

nabcos:
+1

PR #271, opened 2012-02-03
Throttling enhancements.

cstamas:
Implemented Mike's comments. This time, he _did not_ create an issue for these :P
- exposed all throttling variables with fixed names
- implemented lower limiting number sequence
- added possibility to "tune" minimum sleep time
- Added total runtime to log when upgrade ends

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/attributes/internal/DefaultAttributes.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/attributes/internal/DefaultAttributes.java
@@ -73,7 +73,18 @@ public class DefaultAttributes
     @Override
     public String remove( String key )
     {
-        return values.remove( key );
+        // key needs to be removed from both maps
+        final String valuesPrev = values.remove( key );
+        final String defaultsPrev = defaults.remove( key );
+        // but what we "saw" before depends where was the removed value
+        if ( valuesPrev == null )
+        {
+            return defaultsPrev;
+        }
+        else
+        {
+            return valuesPrev;
+        }
     }
 
     @Override
--- a/nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LowerLimitNumberSequence.java
+++ b/nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LowerLimitNumberSequence.java
@@ -1,0 +1,75 @@
+package org.sonatype.nexus.util;
+
+/**
+ * A NumberSequence implementation that wraps another sequence and imposes a lower limit to it.
+ * 
+ * @author cstamas
+ * @since 2.0
+ */
+public class LowerLimitNumberSequence
+    extends NumberSequenceWrapper
+{
+    private long lowerLimit;
+
+    public LowerLimitNumberSequence( final NumberSequence numberSequence, final long lowerLimit )
+    {
+        super( numberSequence );
+        this.lowerLimit = lowerLimit;
+    }
+
+    public long getLowerLimit()
+    {
+        return lowerLimit;
+    }
+
+    public void setLowerLimit( long lowerLimit )
+    {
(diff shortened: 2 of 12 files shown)


nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LowerLimitNumberSequence.java
```diff
@@ -0,0 +1,75 @@
+package org.sonatype.nexus.util;
```
velo, line 13: License header and UT missing

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/NumberSequenceWrapper.java
```diff
@@ -0,0 +1,49 @@
+package org.sonatype.nexus.util;
```
velo, line 13: License header and Junit

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LowerLimitNumberSequence.java
cstamas, line 13: Fixed.

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/NumberSequenceWrapper.java
cstamas, line 13: License headers fixed, and for the other... "eh?" :)

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/LowerLimitNumberSequence.java
```diff
@@ -0,0 +1,87 @@
...
+
+        if ( wrapped < lowerLimit )
+        {
+            // we are already "one step" below
+            return lowerLimit;
+        }
+        else
+        {
+            // perform the state change
+            final long wrappedPrev = super.prev();
```
mah0526, line 58: EDIT:  I see the error in my ways.  the prior conditional check handles this correctly, though you will still sit at the lower limit for one extra call to next(), which is trivial.

This looks like it will move the sequence to a lower number, even if you are already at the low point and will continue to do so potentially indefinitely.  The implication seems to be (if I have this right), then you could have to advance forward potentially many times before you get up above the lowerLimit value -- i.e. you could return the value of lowerLimit repeatedly following calls to .next(), kind of like a Toyota Prius with the gas pedal stuck to the floor :)

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/attributes/internal/DefaultAttributes.java
```diff
@@ -73,7 +73,18 @@ public String put( String key, String value )
     @Override
     public String remove( String key )
     {
-        return values.remove( key );
+        // key needs to be removed from both maps
+        final String valuesPrev = values.remove( key );
+        final String defaultsPrev = defaults.remove( key );
+        // but what we "saw" before depends where was the removed value
```
nabcos, line 79: needs a brief description of what was seen before IMO (or issue ref, ...)

cstamas, line 79: In short (I spotted this on RSO), the "marker" key was left in and persisted:
https://github.com/sonatype/nexus/blob/master/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/TransitioningAttributeStorage.java#L37

Because, item is "lifted", then "attributes", and attributes are _overlaid_ to the item's attributes... meaning, they went in into "defaults" map.

And the transitioning attr storage did call "remove" but it was actually not removing it...

PR #276, opened 2012-02-06
Fixed gitattributes for text files of test repositories which are to be ...

bentmann:
...treated as binary as otherwise checksum errors will occur (on Windows)


velo:
+1

nabcos:
+1

PR #281, opened 2012-02-07
Enhancing attribute upgrade

cstamas:
- applied Ben's comments, using linear seq with small steps to avoid "big step oscillation" (slice of 2sec still provides good adaption)
- added Disposable lifecycle to component to make it unregister from JMX. Not a big deal, but is correct thing to do, and also helps getting rid a lot of noise in IT logs (when 2nd registration tried without unregistering it before).


nabcos:
+1

adreghiciu:
+1

velo:
+1

PR #283, opened 2012-02-07
NEXUS-4826 - removing org.jsecurity entries

velo:
org.jsecurity date from a legacy entry onto log4j.properties that was there since the file was create https://github.com/sonatype/nexus/commit/20814f6b64e56198fee366e15ec2f62d31ef5898#L2R11


nabcos:
+1

bdemers:
+1

PR #285, opened 2012-02-08
Making FixedRateThrottleController UT more reliable.

cstamas:
The original UT was too unreliable, mostly because it was "mixing" millis and nanos (and in some cases it did fail the build without any reason).

Now, the test is simplified, and on my mac returns (if you uncomment the System.out lines) nice values (around 15% of limit TPS)


velo:
+1

nabcos:
+1

adreghiciu:
+1

PR #287, opened 2012-02-08
NEXUS-4789: Fixes

cstamas:
Culprit was HttpClientProxyUtil that was "touching" RemoteStorageContext on every
invocation. This was not an issue so far, since these calls happened only from
RemoteRepositoryStorages. But in a moment any "third party" component -- not
participating in RRS -- does the same, it causes all of RSS to detect as
"settings changed" and reconfigure their underlying transport.

Also, AHC provider modified to make use of newly introduced events
(thanks to Bentmann)

Added Javadocs to the two related events.

---

Note: all this happens due to lack of NEXUS-4689, that would enable all interested "parties" to simply grab a HTTPClient (an abstraction, not an actual 3rd party http client implementation that would be hidden beneath it) from a provider. Without this, the 3rd parties are forced to "tamper" with lower levels of Nexus... doing stuff they should actually never do.


nabcos:
+1

adreghiciu:
+1

PR #288, opened 2012-02-08
NEXUS-4845: Consolidate logging.

cstamas:
Improving logging for some important events and cases.

See related issue for context.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
@@ -17,7 +17,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
@@ -65,8 +64,8 @@ public abstract class AbstractGroupRepository
     implements GroupRepository, CallableExecutor
 {
     /** Secret switch that allows disabling use of Charger if needed */
-    private static final boolean USE_CHARGER_FOR_GROUP_REQUESTS = SystemPropertiesHelper.getBoolean(
-        "nexus.useParallelGroupRequests", false );
+    private final boolean USE_CHARGER_FOR_GROUP_REQUESTS = SystemPropertiesHelper.getBoolean( getClass().getName()
+        + ".useParallelGroupRequests", false );
 
     @Requirement
     private RepositoryRegistry repoRegistry;
@@ -173,7 +172,7 @@ public abstract class AbstractGroupRepository
 
         return result;
     }
-    
+
     // ==
 
     @Override
@@ -226,16 +225,11 @@ public abstract class AbstractGroupRepository
                 }
                 else
                 {
-                    if ( getLogger().isDebugEnabled() )
-                    {
-                        getLogger().debug(
-                            "Repository ID='"
-                                + repo.getId()
-                                + "' in group ID='"
-                                + this.getId()
-                                + "' was already processed during this request! This repository is skipped from processing. Request: "
-                                + request.toString() );
-                    }
+                    getLogger().info(
+                        String.format(
+                            "Repository %s member of group %s was already processed during this request! Skipping it from processing. Request: %s",
+                            RepositoryStringUtils.getHumanizedNameString( repo ),
+                            RepositoryStringUtils.getHumanizedNameString( this ), request.toString() ) );
                 }
             }
         }
@@ -302,16 +296,11 @@ public abstract class AbstractGroupRepository
                         }
                         else
                         {
-                            if ( getLogger().isDebugEnabled() )
-                            {
-                                getLogger().debug(
-                                    "Repository ID='"
-                                        + repository.getId()
-                                        + "' in group ID='"
-                                        + this.getId()
-                                        + "' was already processed during this request! This repository is skipped from processing. Request: "
-                                        + request.toString() );
-                            }
+                            getLogger().info(
+                                String.format(
+                                    "Repository %s member of group %s was already processed during this request! Skipping it from processing. Request: %s",
+                                    RepositoryStringUtils.getHumanizedNameString( repository ),
+                                    RepositoryStringUtils.getHumanizedNameString( this ), request.toString() ) );
                         }
                     }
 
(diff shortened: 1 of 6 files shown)


adreghiciu:
+1

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
```diff
@@ -302,16 +296,11 @@ protected StorageItem doRetrieveItem( ResourceStoreRequest request )
...
-                            {
-                                getLogger().debug(
-                                    "Repository ID='"
-                                        + repository.getId()
-                                        + "' in group ID='"
-                                        + this.getId()
-                                        + "' was already processed during this request! This repository is skipped from processing. Request: "
-                                        + request.toString() );
-                            }
+                            getLogger().info(
```
bdemers: was is this info now? seems like this will create log spam

cstamas: It was debug by mistake (please not that this branch actually never executed so far, since this is the "charger" branch).

But to return to your main point: could you elaborate why you expect cycles should be "normal" in repository configuration?

bdemers:
-1 for 2.0
+1 for 2.1
For 2.0 we should have very select changes at this point

cstamas:
I wonder why you put -1 on this pull.

The point is, that this code change:
- removes multiple existing logging (so this change does not introduce potential NPE, danger or whatnot as Peter mentioned he's afraid of)
- replicates the line already present in NEXUS-4843 (hence, tested for NPE as you were seeing it), it is simply replicated to unify them (copy paste one log invocation does introduce some "error prone" level I believe)
- the two added logging (about transport init and update) are easily testable (and tested by me, as I quoted them in NEXUS-4845 from console): by opening Reposes tab on UI, and doing a change and refreshing on same tab. Something we would do anyway. Or, if not we, almost all of the ITs does it (drives proxies, so NPE would be materialized as IT failure)
- implements NEXUS-4845 and improves overall quality of logging

O_o

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
```diff
@@ -65,8 +64,8 @@
     implements GroupRepository, CallableExecutor
 {
     /** Secret switch that allows disabling use of Charger if needed */
-    private static final boolean USE_CHARGER_FOR_GROUP_REQUESTS = SystemPropertiesHelper.getBoolean(
-        "nexus.useParallelGroupRequests", false );
+    private final boolean USE_CHARGER_FOR_GROUP_REQUESTS = SystemPropertiesHelper.getBoolean( getClass().getName()
+        + ".useParallelGroupRequests", false );
```
peterlynch, line 69: just curious your motivation for removing static - binds too early?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractGroupRepository.java
```diff
@@ -229,12 +228,10 @@ public void expireCaches( ResourceStoreRequest request )
...
-                                + repo.getId()
-                                + "' in group ID='"
-                                + this.getId()
-                                + "' was already processed during this request! This repository is skipped from processing. Request: "
-                                + request.toString() );
+                            String.format(
+                                "Repository %s member of group %s was already processed during this request! Skipping it from processing. Request: %s",
+                                RepositoryStringUtils.getHumanizedNameString( repo ),
+                                RepositoryStringUtils.getHumanizedNameString( this ), request.toString() ) );
                     }
```
peterlynch, line 235: Is there a guarantee that request will never be null? If so why not add a Precondition. If not - time bomb. Does request.toString()  act sanely? What does it print and what type of memory impact is that?

This is an example im my mind that this seems to be a very simple change that is good - but in practice may have a hidden bomb I'd hate to see at the stage of RC.

Yes admittedly if the bomb was there before this change does not make it any worse.

Worth a null check on request? Assuming getHumanizedString handles nulls sanely

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/AbstractRemoteRepositoryStorage.java
```diff
@@ -121,15 +122,28 @@ protected synchronized RemoteStorageContext getRemoteStorageContext( ProxyReposi
...
-                                           new Object[]{
-                                               ctx,
-                                               new Date( ctx.getLastChanged() ),
-                                               new Date( repositoryContexts.get( repository.getId() ) )
-                                           }
-                        );
+                        getLogger().debug(
+                            "Remote context {} has been changed at {}. Previous change {}",
+                            new Object[] { ctx, new Date( ctx.getLastChanged() ),
+                                new Date( repositoryContexts.get( repository.getId() ) ) } );
```
peterlynch, line 128: Seems fragile logging like this. Something you did not introduce but if we are cleaning it up shouldn't we add guarantee these embedded methods calls are not going to blow up new Date()?

peterlynch:
I like where you are going with this but I fear this change or at least parts of this change for 2.0.

String.format is evil for logging and a time bomb waiting for happen. Especially since no one ever writes tests for logging statements.

Methods calls via String format arguments is also evil when inside methods that do not guarantee those objects will not be null.

For 2.0 I would only submit a pull changing log levels and not using String.format. Otherwise -1.

Oh did I mention plexus logger accessed via getLogger method is evil for pull requests diffing - but that is just another rant for another day.

mah0526:
We are effectively under a code freeze at this point with only highly selective changes being incorporated into the release.  While I have no problem with the content of the changes above, there seem to be elements of the change that are not absolutely critical to meeting the minimal objectives for the release.  If correct, then we are incurring risk that we could likely defer to post release -- as well applying time and energy to the review process that might otherwise be spent elsewhere.

Perhaps I sound overly conservative here because these are clearly not significant architectural changes.  Why the fuss?  I believe it is good practice at this point to be excessively cautious as well as judicious in the use of everyone's time.

The above said, rather than make more changes -- potentially incurring more risk -- let's get this the thorough review it needs ASAP and move forward.

bdemers:
I think we should just merge the needed bits from:
https://github.com/sonatype/nexus/pull/288/files#L1R157
https://github.com/sonatype/nexus/pull/288/files#L0R371

This in combination with https://github.com/sonatype/nexus/pull/287
would solve the log spam listed in: https://issues.sonatype.org/browse/NEXUS-4843

mah0526:
I'm fine with that approach as well (merging the needed bits).

cstamas:
I'm really amused seeing "Law of Triviality" in action (see http://en.wikipedia.org/wiki/Parkinson's_Law_of_Triviality but the whole book http://en.wikipedia.org/wiki/Parkinson%27s_law is warmly recommended too).

Everybody trying to explain _so hard_, by _breaking one's neck_, how these logging related changes are _dangerous_, _risky_ and _fragile_, while in the same time having changes happening like 4f567546ad528f5a6897a588b54a31c1c1a7702e. True, is trivial, is _only one liner_ :D

:v: :peace:

PR #292, opened 2012-02-10
o [NEXUS-4848] Make throttle controller UT more robust

nabcos:
Tweaked the UT to get an already throttled baseline, to minimize the risk of underperforming because of other load and thus invalidating the assertions.

([Test failed on the grid](https://builds.sonatype.org/job/nexus-oss-its-external-developer/jdk=java-6x,label=win/4/testReport/org.sonatype.nexus.proxy.walker/FixedRateWalkerThrottleControllerTest/testDoesItHelpAtAll/) because the 'unlimited' run had less TPS than the throttled one.)


cstamas:
:beer:

nabcos:
feature build was https://builds.sonatype.org/job/nexus-oss-feature/30/

PR #300, opened 2012-02-13
[NEXUS-4465] Subtract 1 second from scheduled next run

adreghiciu:
On Nexus initialization of tasks use a 1 second before the actual already scheduled next run so next run calculation (that adds a second) will not skip over the already scheduled time.

Alternatively we could change Sonatype scheduler to add a param or a new method that will also consider the passed in time for calculating resetFrom(), but this is more intrusive.

The disadvantage of proposed solution is that if Sonatype scheduler changes (let's say it adds 2 seconds for calculation) then we have to change this. But this is covered by a test now so we should not miss such a change.


cstamas:
:tea: (alin has a flue)

nabcos:
+1

PR #302, opened 2012-02-13
NXCM-3892, adding UT for dead lock

bdemers:
I'm looking for suggestions (better alternatives) for the SleepingServlet and multiple request threads from a UT.
SleepingServlet is just delays for a number of ms before servicing the request.

### DIFF
--- a/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/maven2/M2RepositoryTest.java
+++ b/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/maven2/M2RepositoryTest.java
@@ -1,0 +1,133 @@
+/*
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+
+package org.sonatype.nexus.proxy.maven.maven2;
+
+import junit.framework.Assert;
+import org.codehaus.plexus.util.IOUtil;
+import org.junit.Test;
+import org.sonatype.jettytestsuite.ServletServer;
+import org.sonatype.nexus.proxy.AbstractProxyTestEnvironment;
+import org.sonatype.nexus.proxy.EnvironmentBuilder;
+import org.sonatype.nexus.proxy.M2TestsuiteEnvironmentBuilder;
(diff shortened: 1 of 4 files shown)


jdillon:
Why is this a UT and not a IT?  UT w/sleep behavior to validate tests are poor tests at best, they also slow down CI builds.

cstamas:
Agreeing with JD:
- https://github.com/sonatype/http-testing-harness has "sleeping servlet" already, no need to reinvent it
- this should be IT not UT
- we are about to switch to Jetty 7/8 (so no org.mortbay imports anymore).

bdemers:
Most of the nexus-proxy tests should be IT's as the load much of nexus.

Should I make this test an *LRTest similar to: DefaultRepositoryItemUidLRTest

What about the test threads question?  Do we have some helper class or anything to deal with this? There are a few tests that create Executers, but that isn't very reusable.

cstamas:
You should make this test an IT. 

The fact that we have many old UTs that you think they should be an IT (which, frankly, I don't understand, since those do tamper with internal components of Nexus, something impossible to do from an IT, that may "speak" with Nexus over REST API only and is isolated from Nexus internals) does not mean you should add newly written test that should be an ITs as UT.

bdemers:
Let me clarify, by: "Most of the nexus-proxy tests should be IT's as the load much of nexus."
This may have better understood as: "Most of the nexus-proxy tests should be *PRTest's as they load much of nexus."

cstamas:
Sure, I might agree with that statement. But still not getting what it has to do with the test we discuss here? Is that the reason to keep this UT?

The reproduce case of NXCM-3892 issue done with wgets clearly calls for an IT... just do two HTTP GETs.

bdemers:
Personally, I do not see the need to start up a full nexus to test this, when it can be done in the nexus-proxy module, which is were the effected code is.

I'm not completely against the IT idea,  any other thoughts ?

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/maven2/M2RepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+
+package org.sonatype.nexus.proxy.maven.maven2;
+
+import junit.framework.Assert;
```
peterlynch, line 16: do not use this package

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/maven2/M2RepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Tests for M2 Repository.
+ */
+public class M2RepositoryTest
+    extends AbstractProxyTestEnvironment
```
peterlynch, line 38: Prefer to not name hog here and call this M2RepositoryTest, which should be held for basic unit test class. Since the setup is more involved, suggest name the test class around what is being tested.

and add the issue number in the test javadoc.

M2RepositoryMetadataJavaUserAgentMTTest

'MT' for multithreaded?

jdillon:
screw it this is old, no one is going to do anything... so killing it.

PR #307, opened 2012-02-16
NXCM-3900: nexus-maven-plugin depends on artifacts not in Central

cstamas:
Hence, it's unusable unless you access RSO too to grab those ("patched" mainly) artifacts.

Most notably, Apache HttpClient 3.1 is patched and is used "almost everywhere", even if that patch is actually needed in Nexus CommonsHttpClient RRS implementation only.

Commons Beanutils Core (and -bean-collection) was also "patched" ones, due to [BEANUTILS-379](https://issues.apache.org/jira/browse/BEANUTILS-379) but in this pull it is reverted to "vanilla" ones (available on Central) and tuned Tattletale configuration instead.

This should make nexus-maven-plugin (and actually any Nexus Restlight Client user) able to resolve dependencies from Central.

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/pom.xml
+++ b/nexus-maven-plugins/nexus-maven-plugin/pom.xml
@@ -219,6 +219,18 @@
       <groupId>org.sonatype.nexus.restlight</groupId>
       <artifactId>nexus-restlight-client-common</artifactId>
     </dependency>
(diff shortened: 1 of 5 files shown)


bdemers:
+1

velo:
Not really a big fan of excluding things from tatletale

Still +1

cstamas:
Agreed, but see comment in tattletale POM...

velo:
That is why I +1 =D

nabcos:
+1

PR #308, opened 2012-02-16
NEXUS-4862 - add zero lenght handler to attribute files

velo:
Fixed attribute storage to discard attribute files that are empty (ZERO length)

Also add test condition for this.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
@@ -226,6 +226,12 @@ public class DefaultFSAttributeStorage
 
                         marshaller.marshal( attributes, fos );
                     }
+                    catch ( IOException ex )
+                    {
+                        // NEXUS-4862 prevent zero length/corrupt files
+                        target.delete();
+                        throw ex;
+                    }
                     finally
                     {
                         Closeables.closeQuietly( fos );
@@ -306,6 +312,12 @@ public class DefaultFSAttributeStorage
 
             try
             {
+                if ( target.length() == 0 )
+                {
+                    // NEXUS-4862
+                    throw new InvalidInputException( "Attribute of " + uid + " is empty!" );
+                }
+
                 fis = new FileInputStream( target );
 
                 result = marshaller.unmarshal( fis );
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
@@ -228,7 +228,7 @@ public class DefaultFSAttributeStorage
                     }
                     catch ( IOException ex )
                     {
-                        // NEXUS-4862 prevent zero length/corrupt files
+                        // NEXUS-4871 prevent zero length/corrupt files
                         if ( target.length() == 0 )
                         {
                             target.delete();
@@ -317,7 +317,7 @@ public class DefaultFSAttributeStorage
             {
                 if ( target.length() == 0 )
                 {
-                    // NEXUS-4862
+                    // NEXUS-4871
                     throw new InvalidInputException( "Attribute of " + uid + " is empty!" );
                 }
 
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
@@ -229,7 +229,10 @@ public class DefaultFSAttributeStorage
                     catch ( IOException ex )
                     {
                         // NEXUS-4862 prevent zero length/corrupt files
-                        target.delete();
+                        if ( target.length() == 0 )
+                        {
+                            target.delete();
+                        }
                         throw ex;
                     }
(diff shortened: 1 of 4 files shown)


velo:
The real ticket would be https://issues.sonatype.org/browse/NEXUS-4871

CI scheduled as well
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/235/

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
```diff
@@ -226,6 +226,12 @@ public void putAttributes( final RepositoryItemUid uid, Attributes attributes )
 
                         marshaller.marshal( attributes, fos );
                     }
+                    catch ( IOException ex )
+                    {
+                        // NEXUS-4862 prevent zero length/corrupt files
```
bdemers: We may want to be more specific and check the length here, there are other reasons why an IOE could be thrown.

cstamas: +1

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
```diff
@@ -226,6 +226,12 @@ public void putAttributes( final RepositoryItemUid uid, Attributes attributes )
...
+                        target.delete();
```
bdemers: If the file is empty just delete it.  If not should we stick it in the trash? or rename it?  (just thoughts/questions)

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java
velo: @bdemers well, if we keep this future read will fail, then this metadata will be deleted and recreated....

And if we are on a disk full situation it many not be a great idea to rename files....

bdemers: true, but nexus will fall down at that point.  
We want to make sure we are not deleting good data.  

What happens if the existing attribute file is valid, the nexus users runs out of file handles, and nexus cannot overwrite the existing file.  We would not want to delete the file in this case, as the old data would be beter then no data.

velo: But if we have an IOException while recording I don't think it is gonna be valid... but sure, I added this check

velo:
new CI build https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/237/

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/attributes/DefaultAttributeStorageTest.java
```diff
@@ -182,4 +191,97 @@ public void testSimplePutDelete()
...
+        long creationTime = new Date().getTime();
+        assertThat( attFile, FileMatchers.sized( 0L ) );
+
+        att = file.getRepositoryItemAttributes();
+        lsStorage.putAttributes( uid, att );
+
+        // put shall create the file
+        assertThat( attFile, FileMatchers.exists() );
+        // must be newer then the zero length file
+        assertThat( attFile.lastModified(), greaterThan( creationTime ) );
```
nabcos, line 279: This may fail depending on the filesystem timestamp resolution (ext3 uses 1s IIRC)

peterlynch:
-1 You've added additional throws of exceptions but have not javadoced those specific exceptions with @throws or explained in javadoc when they will get thrown. First javadoc all those changes.

velo:
@peterlynch but I do not throw any new exception on this methods

https://github.com/sonatype/nexus/blob/949bce9a0c231999e2d65fb8336e180a2fa51caa/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java#L236
The catch block did not existed before, so the IOException was threw cause I caught it do check if file should or shouldn't be delete.  A few lines later (#L250) this exception is logged and not threw any further

https://github.com/sonatype/nexus/blob/949bce9a0c231999e2d65fb8336e180a2fa51caa/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultFSAttributeStorage.java#L321
It is catch at #L345, logged, metadata is set as corrupted and not threw any further

https://github.com/sonatype/nexus/blob/949bce9a0c231999e2d65fb8336e180a2fa51caa/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/attributes/DefaultLSAttributeStorage.java#L243
Same as above.

So the 3 new throw statements I added are self contained in the method body.

velo:
CI voted +1 on windows
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/237/

Failing linux UTs shall run shortly.
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/243/

velo:
UT on linux +1
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/243/jdk=java-6x,label=linux/testReport/org.sonatype.nexus.proxy.attributes/DefaultAttributeStorageTest/

peterlynch:
+1 pending CI - sorry I missed that you eventually caught thrown exceptions.

PR #313, opened 2012-02-17
o [NEXUS-4736] Set up jira mock for error reporting IT

nabcos:
lnx succeeded, win slave is stuck:
https://builds.sonatype.org/job/nexus-oss-its-feature/247/

(I'll release snapshot dependencies before merge)


velo:
This mocking api is awesome... but, I think on this case we should try to reach jira itself.  In the past it did caught problems caused due to new jira version....

cstamas:
:beer:

peterlynch:
@velo I like the comment about testing real jira - here is not the best place for a real test against our real production jira though - we need an internal monitoring job or make it part of our jira upgrade process ) I think it already is.

+1

nabcos:
I think the same as peter, this is not the right place for tests against JIRA. The sisu-problem-reporting project runs tests against the production server and actually checks whether the server-side values match and an attachment exists on the created issue etc.

(sisu-pr is not wired yet because it causes feature builds to go OOM - that's next on my list)

nabcos:
With released PR testutil:
https://builds.sonatype.org/job/nexus-oss-its-feature/261/

PR #314, opened 2012-02-18
NEXUS-4787: Upgrade EHCache to latest

cstamas:
Currently it is 2.5.1. 

This change bumps EHCache version to latest,
but is not trivial one, especially since we have a legacy plexus
component used by nexus core and security. 

As the sisu-ehcache was cleaned, GAV changed to reflect these changes and is made a
drop-in replacement for the old plexus one. But, the new
pure SISU ehcache component depends on new AppContext, this
is bumped too (anyway needed for Jetty8 incoming changes).

### DIFF
--- a/nexus/pom.xml
+++ b/nexus/pom.xml
@@ -1424,6 +1424,13 @@
                   </includes>
                   <message>Use the Eclipse/Jetty version defined in nexus parent. See jetty.version property</message>
                 </bannedDependencies>
+                <bannedDependencies>
+                  <searchTransitive>true</searchTransitive>
+                  <excludes>
+                    <exclude>org.sonatype.spice:plexus-ehcache</exclude>
+                  </excludes>
+                  <message>Use the org.sonatype.sisu:sisu-ehcache instead of plexus-ehcache!</message>
+                </bannedDependencies>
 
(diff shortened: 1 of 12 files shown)


cstamas:
CI queued (not started yet):
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/250/

velo:
CI it then =D

cstamas:
First CI failed for real reasons. Fixed them (in sisu-ehcache but in Nexus too) and scheduled new CI

nexus/pom.xml
```diff
@@ -170,7 +170,7 @@
     <plexus-app-events.version>1.0.1</plexus-app-events.version>
     <plexus-cli.version>1.2</plexus-cli.version>
     <plexus-digest.version>1.1</plexus-digest.version>
-    <plexus-ehcache.version>1.3</plexus-ehcache.version>
+    <sisu-ehcache.version>1.0-SNAPSHOT</sisu-ehcache.version>
```
velo: Add an enforcer rule to make sure we won't drag plexus-ehcache into nexus.

velo:
+1 after enforcer

cstamas:
CI voted +1
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/257/

peterlynch:
+1

adreghiciu:
+1

PR #315, opened 2012-02-22
NXCM-3747: Delete related events made fine grained.

cstamas:
Delete actions now fire different events, making them SP friendly.


cstamas:
CI job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/259/

cstamas:
CI voted +1
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/259/

jdillon:
Not sure I understand RepositoryItemEventDeleteRoot and RepositoryItemEventDeleteRecursively really mean or the frequency they will be emitted when a delete operation occurs.

From reading the javadocs, looks like for a delete op, RepositoryItemEventDeleteRoot would fire, then for each item in a collection that is deleted RepositoryItemEventDeleteRecursively would be fired for each item?

If that is the case, then the name of RepositoryItemEventDeleteRecursively seems misleading.  Just reading the name I'd almost think the *Root and *Recursively events would be equivalent, both saying delete me + anything under me.

cstamas:
True, classes are probably badly named, that's my trademark :)
Please suggest then something for the names, as I guess the solution in general is okay for you and would work for you, and you have problems with class names only...

peterlynch:
-1 no tests
guess need to nudge @jdillon for better names

velo:
+1 after UTs....

But the names are kinda odd indeed....

nabcos:
maybe `RepositoryItemEventDeleteOperation` for the root of the delete operation and `RepositoryItemEventDelete` for every item actually removed (including the root)?

nabcos:
Or even `RepositoryOperationEventDelete` to separate more from the item events?

cstamas:
Two things:
- to keep backward compatibility, both classes should subclass RepositoryItemEventDelete class (and that one is made abstract for this reason). I'd keep this pattern as it is same as we made for create/update stuff. I'd like to keep solutions same, not have different patterns for every of these content change related events.
- the "operated" item will not emit event when recursively walking... simply, since it's a _collection_ (otherwise no walk would happen at all), hence, the two places, where we fire the "operation" and where we fire the "recursive" events are two distinct places.

The current solution fulfills both of these requirements and would work, only problem with this is bad naming. But I am fine with 2nd event (that emitted from recursive walk) to something like RepositoryItemEventDeleteRecursivelyVisited, RepositoryItemEventDeletePing or whatever, just come up with some name and push the button finally.

jdillon:
RepositoryItemEventDeleteRoot (fired at the root of a collection to be deleted) and RepositoryItemEventDeleteItem (fired for each item in a collection being deleted).

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/DeleteEventsTest.java
```diff
@@ -0,0 +1,149 @@
...
+public class DeleteEventsTest
+    extends AbstractProxyTestEnvironment
+{
+    private M2TestsuiteEnvironmentBuilder jettyTestsuiteEnvironmentBuilder;
+
+    @Override
+    protected EnvironmentBuilder getEnvironmentBuilder()
+        throws Exception
+    {
+        ServletServer ss = (ServletServer) lookup( ServletServer.ROLE );
```
bdemers, line 47: Can this test use a Hosted repository (or mocks) to avoid starting up jetty?

cstamas, line 47: It can use hosted, yes, but jetty will be still started up (see JettyEnvironmentBuilder). This is as it was since day 0 of this UT helper class, and to be able to implement what you want (and without manually tamper with repository creation but have the UT do it for you), it would be needed to have deeper changes in AbstractProxyTestEnvironment and/or in JettyEnv builder....

Again, today, even if this UT would use hosted instead of proxy repository, it would not buy much, as Jetty would be still started as part of "env creation".

bdemers, line 47: Yes, that is what I was getting at, the changes to AbstractProxyTestEnvironment (or a different parent test class).
I'm not saying we should hold up this pull, but we need to start doing this soon.

PR #318, opened 2012-02-22
NXCM-3133 - Made it possible to have multiple PathProtectionDescriptor for one PlexusResource

velo:
Quick way to allow multiple PathProtectionDescriptor  w/o breaking current API.

This can be moved outside to plexus-restlet-brige.

### DIFF
--- a/nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/NexusApplication.java
+++ b/nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/NexusApplication.java
@@ -218,15 +218,34 @@ public class NexusApplication
     @Override
     protected void handlePlexusResourceSecurity( PlexusResource resource )
     {
-        PathProtectionDescriptor descriptor = resource.getResourceProtection();
+        PathProtectionDescriptor[] descriptors;
+        if ( resource instanceof AdvancedPlexusResourceSecurity )
+        {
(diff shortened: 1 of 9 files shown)


velo:
FWIW, this is just a initial implementation mean to discuss with @bdemers, but feel free to drop comments

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/NexusApplication.java
```diff
@@ -218,15 +218,34 @@ protected void doCreateRoot( Router root, boolean isStarted )
...
 
-        if ( descriptor == null )
+        if ( descriptors == null )
         {
             return;
         }
 
-        this.protectedPathManager.addProtectedResource( "/service/*"
-                                                        + descriptor.getPathPattern(), descriptor.getFilterExpression() );
+        for ( PathProtectionDescriptor descriptor : descriptors )
```
bdemers, line 245: add a test for this to make sure a user with one of the privileges has access. e.g. make sure a user with the second permission in the list can still access the resource.

velo, line 245: ok, I can add an IT for that on OSS, FWIW, staging test that and it passed!

velo:
Ok, this is no good.

PR #321, opened 2012-02-27
NEXUS-4894 - Add some locks for read/write operations

velo:
Originally I made my changes on at DefaultTargetRegistry.

But when multiple threads tried to create, delete and check if targets was sane concurrency exceptions started to pop everywhere.

This changes look ugly and verbose on my opinion.  And also other parts of nexus would need this "same" change.

May be we could come up with some annotations like @ReadLock, @WriteLock or something.  But I think that needs some discussion first.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/configuration/RevertableConfiguration.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/configuration/RevertableConfiguration.java
@@ -12,6 +12,8 @@
  */
 package org.sonatype.nexus.configuration;
 
+import java.util.concurrent.locks.Lock;
+
 import org.sonatype.configuration.ConfigurationException;
 
 /**
@@ -49,4 +51,19 @@ public interface RevertableConfiguration
     // ==
     
     Object getConfiguration( boolean forWrite );
+
+    /**
+     * Returns the configuration lock used on write operations (like {@link #commitChanges()} and
+     * {@link #rollbackChanges()})
+     * 
+     * @return
+     */
+    Lock getConfigurationWriteLock();
+
+    /**
+     * Returns the configuration lock used on read operations (like {@link #validateChanges()})
+     * 
+     * @return
+     */
+    Lock getConfigurationReadLock();
 }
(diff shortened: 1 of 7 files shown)


velo:
For now -1

nexus/nexus-api/src/main/java/org/sonatype/nexus/configuration/RevertableConfiguration.java
```diff
@@ -12,6 +12,8 @@
  */
 package org.sonatype.nexus.configuration;
 
+import java.util.concurrent.locks.Lock;
```
cstamas, line 15: Change to SISU locks, for configuration we should not use  j.u.c locks -- but this statement is not true in general (or the other way around). so SISU locks are not to be used _everywhere_.

nexus/nexus-api/src/main/java/org/sonatype/nexus/configuration/RevertableConfiguration.java
```diff
@@ -49,4 +51,19 @@ void commitChanges()
...
     
     Object getConfiguration( boolean forWrite );
+
+    /**
+     * Returns the configuration lock used on write operations (like {@link #commitChanges()} and
+     * {@link #rollbackChanges()})
+     * 
+     * @return
+     */
+    Lock getConfigurationWriteLock();
```
cstamas, line 61: Why exposing the locks here? As I see, no implementor of RevertableConfiguration uses the lock, meaning, the locks being here is wrong decision.

velo, line 61: https://github.com/sonatype/nexus/pull/321/files#L3R305

it does try to change the configuration, 
so either we expose add/remove/update like operation on CRepositoryTargetConfiguration....
or does like this.

mah0526:
Perhaps this has already been considered... As part of this, the paths for lock acquisition need to be evaluated.  If there was lower level locking happening previously (assume there was since the configuration was being rebuilt out side of the DefaultTargetRepository class) we need to know how and where that was being called.

PR #323, opened 2012-02-28
o [NEXUS-4754] Don't display validation errors for nexus managed roles

nabcos:


nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.RoleEditPanel.js
```diff
@@ -463,7 +463,11 @@ Ext.extend(Sonatype.repoServer.DefaultRoleEditor, Sonatype.ext.FormPanel, {
           this.find('name', 'id')[0].disable();
           this.find('name', 'name')[0].disable();
           this.find('name', 'description')[0].disable();
-          this.find('name', 'roleManager')[0].disable();
+
+          var roleManager = this.find('name', 'roleManager')[0];
+          roleManager.disable();
+          roleManager.clearValidation();
```
velo, line 469: Won't this show validation and remove right after? Kinda making the UI to blink?

No idea if that is gonna happen just asking =D

velo:
+1 if UI not blinking
if it is, we may should consider a different approach.

nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.RoleEditPanel.js
nabcos, line 469: The change is in the `loadHandler`, so I hope the field is not rendered before ;)
FWIW, I did not notice a flicker.

bdemers:
+1 if no flicker

PR #327, opened 2012-03-01
o [NXCM-3914] Defer loading of predefined repos until Mirrors tab is activated

nabcos:
Eagerly loading the predefined repos will cause the p2 proxy to hit all mirrors with a request for '/.meta/repository-metadata.xml'. This will still happen, but only when the mirros tab is actually selected. There is some underlying bug that needs to get fixed, but we could 'cure' the symptom with this pull request.

### DIFF
--- a/nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.MirrorConfigPanel.js
+++ b/nexus/nexus-webapp/src/main/webapp/js/repoServer/repoServer.MirrorConfigPanel.js
@@ -273,6 +273,8 @@ Ext.extend(Sonatype.repoServer.AbstractMirrorPanel, Sonatype.ext.FormPanel, {
       },
 
       activateHandler : function(panel) {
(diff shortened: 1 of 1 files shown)


velo:
+1

cstamas:
The related issue mentioned by Ben
https://issues.sonatype.org/browse/NXCM-3751

nabcos:
Not sure that this is the 'real' issue for  P2. IIUC mirrors should not be asked for stuff in .meta at all, so the P2 repos (or abstract repo in nexus, or wherever) make a mistake in relaying these requests to their mirrors(?)

cstamas:
THat is the real one, reason is this class "NexusRawTransport" (https://github.com/sonatype/nexus/blob/master/nexus/nexus-app/src/main/java/org/sonatype/nexus/repositories/metadata/NexusRawTransport.java) that delegates to actual repository.

PR #330, opened 2012-03-01
NEXUS-4815 - minor changes to log configuration UI

velo:


nabcos:
why exactly are we not removing the read only fields?

velo:
It is doable, but with read only fields it is prettier
http://screencast.com/t/cfnyibbrb

Anyway, I can remove....

nabcos:
it's only I remember @rseddon saying the fields should be removed.
But the change is okay -> +1

cstamas:
:cake:

PR #331, opened 2012-03-02
NEXUS-4894: Fixes for TargetRegistry thread safety.

cstamas:
Multiple problems, as this is one of the oldest unchanged class in core it seems.
- looking for target by ID is very, very, VERY often done by Security, but it was done
  in a "linear search" manner, by iterating over a list and constructing object(s) on the fly
- a "cache view" (list) was present (but unused in case above, unsure why) but was only
  nullified in case of configuration change, and was "lazily" rebuilt on lookups/gets. This was not only
  thread unsafe (see related issue), but also caused slow-downs for target getter (most certainly
  security). Now, the "cache view" is a map, keyed by Target ID and eagerly rebuilt on config
  changes.
- other minor improvements

Now we ensure that the "cache view" is rebuilt on config change (in config changing thread,
where through-output is not a concern). Also, this excludes the thread safety problem
of lazily creating the "cache view" and stealing CPU cycles from security.

Also, this makes a get-target-by-ID faster by orders of magnitude, as the "cache view"
is now a map that already contains pre created "live" object (no object creation on the fly)
keyed by target id (no linear search).

CI build:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/272/

### DIFF
--- a/nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
+++ b/nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
@@ -13,6 +13,7 @@
 package org.sonatype.scheduling;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 
 import java.text.ParseException;
@@ -179,16 +180,10 @@ public class DefaultTaskConfigManagerTest
             defaultManager.addTask( task );
 
             // loadConfig();
-
-            assertEquals( 1, getTaskConfiguration().size() );
-
-            assertEquals( TaskState.SUBMITTED,
-                TaskState.valueOf( ( (CScheduledTask) getTaskConfiguration().get( 0 ) ).getStatus() ) );
-
-            assertEquals( TASK_NAME, ( (CScheduledTask) getTaskConfiguration().get( 0 ) ).getName() );
-
-            // assertTrue( typeClassMap.get( scheduleType ).isAssignableFrom(
-            // ( (CScheduledTask) getTaskConfiguration().get( 0 ) ).getSchedule().getClass() ) );
+            
+            assertThat( getTaskConfiguration().size(), equalTo( 1 ) );
+            assertThat( TaskState.valueOf( ( (CScheduledTask) getTaskConfiguration().get( 0 ) ).getStatus() ), equalTo( TaskState.SUBMITTED ) );
+            assertThat( ( (CScheduledTask) getTaskConfiguration().get( 0 ) ).getName(), equalTo( TASK_NAME ) );
 
             defaultManager.removeTask( task );
 
--- a/nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
+++ b/nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
@@ -155,14 +155,14 @@ public class DefaultTaskConfigManagerTest
         cst.setType( TestNexusTask.class.getName() );
         cst.setNextRun( new SimpleDateFormat( "yyyy-MM-DD hh:mm:ss" ).parse( "2099-01-01 20:00:00" ).getTime() );
 
-        System.out.println(new Date(cst.getNextRun()));
+        System.out.println( new Date( cst.getNextRun() ) );
 
         final CScheduleConfig csc = new CScheduleConfig();
         csc.setType( SCHEDULE_TYPE_ADVANCED );
         csc.setCronCommand( "0 0 20 ? * TUE,THU,SAT" );
         cst.setSchedule( csc );
-        
-        defaultManager.initializeTasks( defaultScheduler, Arrays.asList( cst ));
+
+        defaultManager.initializeTasks( defaultScheduler, Arrays.asList( cst ) );
 
         final ScheduledTask<?> task = defaultScheduler.getTaskById( cst.getId() );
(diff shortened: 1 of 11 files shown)


velo:
Missing UT for the problem.

cstamas:
@velo If you read related issue description, then you will see it's very involved. If by "UT for the problem" you mean a unti test that would _fail_ with old code and _pass_ with this change, I am unsure it's easily doable, especially since the "configuration framework" is spread across multiple modules (from nexus-configuration to nexus-app, with this last one having the majority of the stuff for config framework).

But will take a peek your UT you had in prev pull req...

velo:
FWIW https://github.com/sonatype/nexus/pull/321/files#diff-5

cstamas:
Another CI
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/274/

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/model/CRepositoryCoreConfiguration.java
```diff
@@ -134,6 +134,8 @@ else if ( !cfg.getId().matches( REPOSITORY_ID_PATTERN ) )
         }
 
         // ID uniqueness
+        /* This is totally flawed, but this is NOT THE ONLY ONE of the places we did uniqueness check!
+         * But this one in certainly not the place to do this.
         List<CRepository> repositories = getApplicationConfiguration().getConfigurationModel().getRepositories();
```
nabcos, line 139: Rather remove the code and leave the comment only? Is there a reason to leave it around?

nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
```diff
@@ -180,12 +180,12 @@ public void genericTestStore( String scheduleType, HashMap<String, Object> sched
 
             // loadConfig();
 
-            assertTrue( getTaskConfiguration().size() == 1 );
+            assertEquals( 1, getTaskConfiguration().size() );
```
nabcos: you touched that but did not change to assertThat? ;)

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/target/DefaultTargetRegistryLRTest.java
```diff
@@ -0,0 +1,138 @@
...
+        {
+            Target t = targetRegistry.getRepositoryTarget( id );
+            assertThat( "ref: " + ref + " ids: " + ids, t, nullValue() );
+            assertThat( "ref: " + ref + " ids: " + ids, targets, not( containsInAnyOrder( id ) ) );
+        }
+    }
+
+    @Test
+    @Ignore
+    public void testConcurrency()
```
nabcos, line 110: the only test method in this class is ignored...

nabcos, line 110: ... and that fact is noted in the commit description. Please add some javadoc here for clarification.

nexus/nexus-app/src/test/java/org/sonatype/scheduling/DefaultTaskConfigManagerTest.java
cstamas: Yes, since as I mentioned on scrum yesterday, I tried fixes "here and there" and ended up with a disaster. This change was made by me only to get better failure message (a bit, like "size is not 1, but still, what is the size?").

This change did not serve any other purpose. But will fix it.

nexus/nexus-configuration/src/main/java/org/sonatype/nexus/configuration/model/CRepositoryCoreConfiguration.java
cstamas, line 139: Left only for reference. We have a LOT of duplicated (and inconsistently duplicated) validation code in Nexus core, and we are about to clean this up. The code is left here (should be left here) only up to the point where we consolidate validation.

I think this commented code does not hurt a lot for now.

cstamas:
Applied Ben's comments.

nabcos:
+1

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/target/DefaultTargetRegistryLRTest.java
```diff
@@ -0,0 +1,143 @@
...
+        for ( String id : ids )
+        {
+            Target t = targetRegistry.getRepositoryTarget( id );
+            assertThat( "ref: " + ref + " ids: " + ids, t, nullValue() );
+            assertThat( "ref: " + ref + " ids: " + ids, targets, not( containsInAnyOrder( id ) ) );
+        }
+    }
+
+    @Test
+    @Ignore
```
velo, line 109: IMHO, an @Ignore'd test is as good as no test.

We need to come with a Test that does test the user scenario (even if that mean an IT).  Otherwise there is no way of telling if the problem got fixed or not and it will likely rise again.

cstamas, line 109: The user scenario is "configuration modification from multiple threads". The current nexus configuration is unable to handle that, as I wrote in the comment of this commit and in class javadoc.

Simply put: we know something is not working, and we also know, we cannot fix it without complete overhaul of configuration in core (see NEXUS-4909). So, this problem is not about this test, but more about config framework and it's deficiency.

adreghiciu:
+1

PR #333, opened 2012-03-06
NEXUS-4933 - Using version to decide if SnapshotVersion should be updated or not

velo:
When metadata is rebuild from zero the 2 different SNAPSHOT may end with the same  timestamp and the SnapshotVersions end wrong.

Using timestamped version should handle this a bit better.


nabcos:
+1

adreghiciu:
+1

PR #334, opened 2012-03-06
o [NEXUS-4815] Let log appender pattern field have a default value

nabcos:
As commented on the issue by @bdemers, this commit makes the log pattern be optional with the default value used for an 'empty' field.


bdemers:
+1

cstamas:
:beer:

PR #335, opened 2012-03-07
NEXUS-4901: Undoing some changes making impossible Shadows to be "lazy"

cstamas:
Change in 64fb67dc16ef2e18300d3df3eb4f87d419b584cc was too "eager"
and introduced some changes that made shadows unable to perform "lazy",
and be pointed at groups for example.


nabcos:
+1

PR #336, opened 2012-03-07
[NEXUS-4905] record events about all files in timeline

adreghiciu:


adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/275/

nexus/nexus-core-plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/rest/feeds/sources/NexusFileEventEntryBuilder.java
```diff
@@ -63,4 +64,32 @@ protected String buildDescriptionMsgItem( NexusArtifactEvent event )
...
 
+    @Override
+    public boolean shouldBuildEntry( final NexusArtifactEvent event )
+    {
+        if ( !super.shouldBuildEntry( event ) )
+        {
+            return false;
+        }
+
+        final Gav gav = buildGAV( event );
```
cstamas, line 75: IMO, the "file" entries are not about artifacts ("artifacts" implied here as "maven artifacts"), hence this GAV check makes the content almost the same as the artifacts feed (class above).

Here, IMO, no GAV check should happen and should just pour _all files_ it has.

Again, the main distinction about *artifact and *files feeds (those that are doubled) is that former shows only _files obeying maven2 layout_ while latter shows all (obviously, latter is superset of former).

Still, "general" filtering rules do apply (ie. content generated items are not shown, but that is handled in EventInspector, and are not even recorded)

cstamas:
:beer:

nexus/nexus-core-plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/rest/feeds/sources/NexusFileEventEntryBuilder.java
nabcos, line 75: IIUC the artifact feed shows only 'interesting' files from m2-repositories, and the file feed shows only 'interesting' files from all repository types. (With 'interesting' meaning not generated and not a m2 signature or hash.)

Seems fine for me, not need to bomb any feed with 'uninteresting' updates.

nabcos:
+1

nexus/nexus-core-plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/rest/feeds/sources/NexusFileEventEntryBuilder.java
cstamas, line 75: Sure, that's why I put "IMO" (i did not refer to International Meteor Organization, see http://vmo.imo.net/observer/code/CSETA/view :)

This is fine, but is Maven specific....

adreghiciu, line 75: I could drop the GAV class from here, yet we should still have to make this class filter out md5/sha1 maven files.
So I could just check that paths end with .m5/.sha1. If you think is less tight to maven I can do that.

Another. more complex alternative is the way we discuss by phone Tamas to have an IsVisibleInFeeds attribute that divers repos could implement specific to their content and do filtering baed on that.

Shall I go that route? I just did not do it to not spend more time on it. Timeline is anyhow tiden to maven repo format already and we may delay this till we re-implement timeline as a journal.

adreghiciu:
Build #275 failed to run. Started https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/277/

adreghiciu:
CI voted +1

nexus/nexus-core-plugins/nexus-timeline-plugin/src/test/java/org/sonatype/nexus/feeds/record/ItemChangesFeedEventInspectorTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonatype.nexus.proxy.item.AbstractRepositoryItemUidFactory;
+import org.sonatype.nexus.proxy.item.DefaultRepositoryItemUid;
+import org.sonatype.nexus.proxy.item.DefaultStorageFileItem;
+import org.sonatype.nexus.proxy.item.RepositoryItemUid;
+import org.sonatype.nexus.proxy.item.StorageFileItem;
+import org.sonatype.nexus.proxy.item.uid.IsHiddenAttribute;
+import org.sonatype.nexus.proxy.repository.Repository;
+import org.sonatype.sisu.litmus.testsupport.TestSupport;
+
+public class ItemChangesFeedEventInspectorTest
```
velo, line 39: missing jdoc

velo:
+1

nexus/nexus-core-plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/rest/feeds/sources/NexusFileEventEntryBuilder.java
cstamas, line 75: This is fine for now, just push this as is, and we will solve things as we progress with journal.

PR #337, opened 2012-03-08
o [NXCM-3929] Display failed staging rules on close/promote

nabcos:
Depends on sonatype/nexus-enterprise#165

https://builds.sonatype.org/job/nexus-oss-its-feature/276/

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
+++ b/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
@@ -16,8 +16,11 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.codehaus.plexus.components.interactivity.PrompterException;
+import org.jdom.Document;
+import org.jdom.Element;
 import org.sonatype.nexus.restlight.common.ProxyConfig;
 import org.sonatype.nexus.restlight.common.RESTLightClientException;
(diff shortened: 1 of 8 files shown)


nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/CloseStageRepositoryMojo.java
```diff
@@ -70,6 +73,13 @@
     @Deprecated
     private boolean auto;
 
+    /**
+     * If true, the mojo will try to remove the staging repository it was not able to close.
+     *
+     * @parameter expression="${nexus.removeOnFailure}" default-value="false"
```
velo: I would use DROP instead of 'remove'. Just to use same terminology as UI.

velo:
+1

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -246,4 +253,49 @@ public void setRepositoryId( final String repositoryId )
         this.repositoryId = repositoryId;
     }
 
+    protected String ruleFailureMessage( final Document document )
```
peterlynch, line 260: Should indicate @throws NullPointerException if document null ( javadoc)

cstamas:
:beer:

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/CloseStageRepositoryMojo.java
peterlynch: I agree use drop to keep terms the same. I guess this means rename remove to drop in the various helper methods.

peterlynch:
other than the things mentioned +1

PR #338, opened 2012-03-09
Use the eventually GA found plugin not the passed in GAV

adreghiciu:
The bug related to booting in case that a plugin dependeing on another version of another plugin then the one in plugin-repository seems as not fixed by the previous try. This because the version that we returned and used later on to create the injector was the one expressed in plugin dependency and not the one founded by GA.


adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/279/

cstamas:
:beer:

nabcos:
+1

peterlynch:
-1 why not add a unit test that breaks before this change and passes after? there is already DefaultNexusPluginManagerTest.java to start with.

adreghiciu:
@peterlynch The problem with test is that the error is only manifested when the plugin injector is created. And in the UT we do not wanna do that. The existing test was covering the use case only that in the UT we did override the container creating so, we did not get the exception.

adreghiciu:
CI voted +1 (one test failing on Windows but unrelated)

PR #339, opened 2012-03-13
NEXUS-4893: Make Jetty use real logger

cstamas:
SLF4J and it's logback "backend" (is not backend but direct implementation)
moved into bundle. But, to avoid problems already described, logging
related packages are made "server classes", classes hidden from WAR,
as Nexus deliver the same stuff, and extends those by adding custom
appender implementations.

This change does/fixes:
- the original issue, as at "server level" it is not the SrdErrLogger trivial implementation used by Jetty to log (server logs and servlet logs) but logback (with Jetty's preferred native SLF4J support).
- the Jetty -- based on that item above -- now uses Slf4j and does not break the log "layout" (there was a similar issue, as pattern Jetty's logger used was different than Nexus pattern, making logs hard to read and parse)
- logging related classes are made "server classes" to not impose any requirements against Nexus WAR, hence, Nexus WAR wrt it's dependencies it ships in /lib folder are unchanged, it's still "fully functional" (and will not log to "noop" logger).

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/pom.xml
+++ b/nexus-maven-plugins/nexus-maven-plugin/pom.xml
@@ -298,6 +298,13 @@
     </dependency>
     <dependency>
       <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+      <version>${logback.version}</version>
+      <type>jar</type>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <version>${logback.version}</version>
       <type>jar</type>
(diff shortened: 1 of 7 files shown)


jdillon:
Cursory inspection looks good.  What does ci say?

velo:
+1
@jdillon I don't think this will affect CI since CI doesn't test the bundle itself.

@cstamas no changes need to nexus-bundles??

jdillon:
ci says this doesn't compile:

https://builds.sonatype.org/job/nexus-oss-feature/jdk=java-6x,label=linux/lastBuild/console

velo:
I could be wrong =D

cstamas:
Started CI job:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/285/

cstamas:
CI linux voted +1

peterlynch:
:beer:

cstamas:
Both slaves (lin and win) says +1

PR #341, opened 2012-03-14
o [NXCM-3953] Add capability to sort items in a collection before processing

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/281/

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalkerContext.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/walker/DefaultWalkerContext.java
@@ -13,11 +13,13 @@
 package org.sonatype.nexus.proxy.walker;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import org.sonatype.nexus.proxy.ResourceStoreRequest;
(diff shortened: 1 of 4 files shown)


jdillon:
How does sorting avoid problem captured in NXCM-3953?

velo:
+1
but wanna hear the answer for jason's question too.

nabcos:
Preparation for the real fix in Pro. From the description of the pull request referencing this one (sonatype/nexus-enterprise#175):

This is important for staging snapshots, when you have multiple unique
snapshot versions in the staging repository. The items were processed in
file system order before, so the snapshot metadata wasn't always set to
the latest snapshot in the staging repo.

jdillon:
mkay, cursory look at the issue shows the test depends on timing information which seemed fishy compared to the fix here that it is for.   Though generally looks like this change is fine, looks like its optional feature to use when needed and not always going to be sorting which was what I had thought originally (which would be bad imo to always force to sort when walking).

+1

nabcos:
yes, the comparator has to be explicitely set in the WalkerContext. If not set, behavior is as before.

The test stages multiple unique snapshot versions and expects the newest one to 'win'. That's the behavior for the 'Recreate Maven Metdata' task as well (but according to Tamas it's actually a bug for the recreation, IIUC it should actually be 'last modified wins'). So the only place using a sorted walk so far is staging when promoting repositories.

PR #345, opened 2012-03-19
Removed url from plugins that don't really have a site

velo:
UI already handle this just fine.


peterlynch:
:beer:

velo:
https://github.com/sonatype/nexus-enterprise/pull/180

jdillon:
Is simply to remove the default generated url from parent no?  Perhaps the parent should omit its url so that plugins don't have to put this empty tag in to override it?

I admit I tend to remove these as they don't clearly provide any reason to exist...

velo:
No, if you remove from top level maven add a /${lower-lever-artifactId}....

jdillon:
I've no clue what your comment means @velo

velo:
Top level pom
<url></url>

Plugin pom
no url entry

Effective plugin pom (what maven sees after all CoC is applied)
<url>/plugin</url>

PR #348, opened 2012-03-21
Sharing some debug utilities I had locally.

cstamas:
DebugEventInspector simply "spits" out the Nexus internal events
for easier debugging. It can be enabled via system properties
or JMX.

LogManager also exposed over JMX for easier management.

By having JConsole (or any JMX client), you can now easily tune
the DEBUGness of running nexus instance without even bouncing it,
and without even touching the UI (for logging level).


nabcos:
+1

nabcos:
if it's easy to do adding a xml conf snippet to the log configuration over jmx would be useful as well =)
(similar to what @jdillon did with the capability in sp)

cstamas:
Sure thing. 
This pull request contains what I had locally, but nothing stops us extending what we expose over JMX.

adreghiciu:
+1

PR #349, opened 2012-03-21
NEXUS-4955 - expand the prefix when it is too short

velo:
This issue was reported on user list.

Basically File.createTempFile() fail if prefix is too short, what may happen when the file name we are uploading is too short.

I just append System.nanotime() to it, but we can append anything else.

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/RepositoryDependentException.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/RepositoryDependentException.java
@@ -1,3 +1,15 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
 package org.sonatype.nexus.configuration.application;
 
 import static java.lang.String.format;
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
@@ -56,6 +56,8 @@ public class DefaultFSPeer
 {
     private static final String HIDDEN_TARGET_SUFFIX = ".nx-upload";
 
+    private static final String APPENDIX = "nx-tmp";
+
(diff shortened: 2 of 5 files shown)


nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/RepositoryDependentException.java
```diff
@@ -1,3 +1,15 @@
...
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
 package org.sonatype.nexus.configuration.application;
```
cstamas, line 13: ? What's this?

velo, line 13: A missing header...

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
```diff
@@ -328,6 +333,18 @@ protected File getHiddenTarget( final File target, final StorageItem item )
         }
     }
 
+    private String expandPrefix( String prefix )
```
cstamas: Never leave _any chance_ for shit to hit the fan.

Here, you would have it, if -- even if there is a minimal chance, but there IS a chance -- to deploy file "id" and have nanoTime as 12345... and then in very next moment deploy of file "id12345..."... kaboom, collision might happen.

Why not _always prefix with same content_? 

Even something like `return "aaa" + prefix;` would work, as with this you effectively "translate" the space (which is a "space of possible file names" here) to have it's border above minimum required (minimum name length of 3 characters), but since translation is "rigid motion", you _preserved uniqueness too_, unlike the solution in this pull request (and hence, the minimum chance to something smelly hit the fan is present).

velo: So you are saying collision will happen 100% of the times name is bigger then 3 chars?  It doesn't make sense.

I will use the constant instead of nanoTime.

cstamas: No, I am saying there is room for collision.

One deploys file "id", and nano at the moment is "12345". So, the prefix you create is "id12345".
In same moment, _someone else_ deploys file named "id12345". Here, you do not prefix, since name is longer than 3.

But the strings returned by new expandPrefix method are actually the same.

velo: So when there is no nano on the mix there will be collision....

cstamas: No, it is the condition that makes it possible. If you append unconditionally with a constant, you make it impossible to produce a collision.

velo: ok, then I will remove the if.

cstamas:
:beer:

adreghiciu:
+1

PR #350, opened 2012-03-21
o [NXCM-4274] Avoid UI credential timeouts

jdillon:
I've been testing this for a week or so now, and the NX instance I'm running has yet to prompt me for basic auth muck.

I think we should consider merging this fix.


bentmann:
The RHC plugin initially had a similar JavaScript task polling the Nexus server for status info which eventually was considered bad due to the traffic.

jdillon:
Perhaps this should only be installed once authenticated?

jdillon:
BTW, I think its worse to show a browser auth panel than to hit this url periodically, since the former actually breaks the UI.

The alternative would be some less often check to see if auth is broken, and then force the UI to log out.

jdillon:
... and/or make a capability to turn this on/off

nexus/nexus-webapp/src/main/webapp/js/Sonatype.utils.js
```diff
@@ -776,6 +777,29 @@
...
+                {
+                  Ext.Ajax.request({
+                    method : 'GET',
+                    options : {
+                      ignore401 : true
+                    },
+                    url : Sonatype.config.repos.urls.status,
+                    callback : function(options, success, response) {
+                      // do nothing, we only request to refresh session
+                    }
```
peterlynch: Does the UI lock up, wait, throw error dialog on any http error doing this refresh? Sounds like it was only tested when server was running?

Expecting that nothing noticeable in the UI if Nexus is unreachable or connection is slow.

^ bring this up because I think I have seen the status resource ping throw error dialogs on connection failure in the past.

peterlynch:
hesitant to merge this until after Shiro 1.2 upgrade is merged.

nexus/nexus-webapp/src/main/webapp/js/Sonatype.utils.js
nabcos: did not try, it will probably show the generic error message box.

peterlynch: ok would like to avoid

adreghiciu:
+1

cstamas:
+1

What's happening on this front?

peterlynch:
just don't have much faith in this since corner cases are not being tested

jdillon:
Ya, I think there are corners too.  Like when I have security turned off, then enable it, I'm prompted for auth via the browser mechanism twice in a row before it gives up (w/o this change) no clue how it would behave with this change.

peterlynch:
We decided testing this further would be ok for 2.1 since if it works reasonably well then it is better than user experience we currently have. Will add a jira for testing this so that we can merge

peterlynch:
https://issues.sonatype.org/browse/NXCM-4274

nabcos:
Tested and refactored a bit to make this more robust. Should yield no visible UI on http errors - and the task will stop itself on an error. It will start on (UI) login or 'load with credentials' (like reloading page while logged in), stop on (UI) logout.

peterlynch:
+1 lets give a try

PR #352, opened 2012-03-22
NEXUS-4943: Generalize metadata handling

cstamas:
The existing "maven metadata manager" concept pulled up a level, with "defaulted" NOOP implementation for now, except in case of M2Repository where a real implementation (merely just delegation to existing calls) added.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultStorageNotFoundItem.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultStorageNotFoundItem.java
@@ -1,0 +1,32 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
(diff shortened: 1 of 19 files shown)


jdillon:
Missing tests?

I took a quick look and nothing negative to say about it ;-)  A few empty/superfluous javadocs (like \@return w/o any text).

What is the example use for implementing the metadata related bits for the SP spec ( https://docs.google.com/a/sonatype.com/spreadsheet/ccc?key=0An3vtHmXvlwXdDUzVG5aMkFCMXE3TmVKd2tEUkUya1E#gid=0 )?

cstamas:
```
  Repository repository = ...
  repository.getRepositoryMetadataManager().expireMetadataCaches( new ResourceStoreRequest("/") );
```

This could be done against any repository, but currently only M2 repositories will do anything (others not implementing this feature will fall back to NOOP manager doing nothing).

jdillon:
Okay, I asked Alin to look this over for sanity.  I think this change is going to be enough to generically cope with metadata muck... 

+1

adreghiciu:
+1

cstamas:
Rebased against latest master.
Added tests, and those tests resulted in a bug discovery and it's fix is in place. 

This also means, that JD owes beer to two of us, as he actually made a good deed by asking for UTs. And people doing good deeds usually give out some beer too. At least, it stems from their "good" nature. They like to do good things to others. So, conclusion is that JD owes 2 beers: one for @adreghiciu and one for me :D

cstamas:
Before anyone starts panicking: it's friday past 6PM here :P

PR #354, opened 2012-03-23
Dependency cleanup

cstamas:
Some cleanup wrt variety of versions for same artifacts we use
across Nexus.

Is good to be merged, but eyeballs needed.

Got inspired by
http://www.jasonwhaley.com/blog/2012/03/21/dependency-convergence-in-maven/

Modules up to "lightweight" client is cleaned up (hence, Core is cleaned).

The enforcer is not quite working (light client deps), so did not include it here. Just the results it "caught".

### DIFF
--- a/nexus-runtime-platform/pom.xml
+++ b/nexus-runtime-platform/pom.xml
@@ -121,7 +121,7 @@
       <dependency>
         <groupId>org.codehaus.plexus</groupId>
         <artifactId>plexus-utils</artifactId>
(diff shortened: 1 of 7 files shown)


adreghiciu:
+1 CI?

cstamas:
Started CI:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/293/

nabcos:
+1

PR #355, opened 2012-03-23
NEXUS-4986: Get rid of duplicated classes for parsing Artifact Versions

cstamas:
Metadata VersionComparator "redirected" to proper class, other one
deprecated.

Some context: the deprecated class came along with other Mercury (pre-Aether library) classes, as they were lifted into Nexus core.

The proper class came into core with Nexus 2.0, and is copied from Aether (see Javadoc) and is present as long as Maven support gets out of Core (Maven support should be in plugin just like any other repo implementations, keeping Core thin).


adreghiciu:
+1

nabcos:
+1

PR #359, opened 2012-03-30
NEXUS-4985: AbstractGroupRepository#doRetrieveItem(ResourceStoreRequest) hides actual reason why something is not found in it's member

cstamas:
Changes:
- introduced "bookkeeping" for groups, made them not "swallow" member exceptions anymore
- REST API ?describe method changed to make use of this new information
- improvements for some exception messages
- improvements for some toString() methods (simply for easier debug)

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
@@ -13,6 +13,7 @@
 package org.sonatype.nexus.proxy;
 
 import org.sonatype.nexus.proxy.repository.Repository;
+import org.sonatype.nexus.proxy.utils.RepositoryStringUtils;
 
 /**
  * Thrown if the requested item is not found.
@@ -34,6 +35,7 @@ public class ItemNotFoundException
      * @param path
      * @deprecated use a constructor that accepts a request!
      */
+    @Deprecated
     public ItemNotFoundException( String path )
     {
         this( path, null );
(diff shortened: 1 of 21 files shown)


nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -46,51 +48,93 @@ public ItemNotFoundException( String path )
...
     {
         this( request, null, null );
     }
 
-    public ItemNotFoundException( ResourceStoreRequest request, Repository repository )
+    /**
+     * Constructor. To be used in places where no Repository exists yet in context (like in a Router).
+     * 
+     * @param request
+     * @param cause
```
peterlynch, line 73: Would have been nice to mention what calling these constructors would have been equivalent too ( null repo for example) and fill in param javadoc.

peterlynch:
can we get some UTs that verify the output of multiple failure reasons?

cstamas:
Added UT. CI started:
https://builds.sonatype.org/view/nexus/job/nexus-oss-feature/38/

cstamas:
Bah, CI succeed, but file RemoteItemNotFoundException misses a valid file header.
So, fixing the missing header but this is a +1 from CI

peterlynch:
+1 thx for tests

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/ArtifactStoreRequest.java
```diff
@@ -88,7 +89,9 @@ public String getExtension()
     @Override
     public String toString()
     {
-        StringBuffer sb = new StringBuffer( getGroupId() );
+        StringBuffer sb = new StringBuffer( super.toString() );
```
peterlynch: why not use StringBuilder?

cstamas: Bah. a mistake, fixing it.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/charger/AbstractRetrieveCallable.java
```diff
@@ -0,0 +1,78 @@
...
+    private final Repository repository;
+
+    private final ResourceStoreRequest request;
+
+    private Throwable processingException;
+
+    public AbstractRetrieveCallable( final Logger logger, final Repository repository,
+                                     final ResourceStoreRequest request )
+    {
+        this.logger = Preconditions.checkNotNull( logger );
```
jdillon, line 46: Using standard static factory for loggers here can simplify this code and its usage.

```
protected final Logger logger = LoggerFactory.getLogger(getClass());
```

IMO there is only a very limited scope when its sane to passing in a Logger object to another object to use.

jdillon:
+1 ... other that the bits about passing Logger instances around, which is minor, but IMO should be considered for clean up to simplify this change.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/charger/AbstractRetrieveCallable.java
cstamas, line 46: This code is used in conjunction with [Charger](https://github.com/sonatype/sisu-charger) in Nexus Core that enables to "parallelize", end hence, greatly improve response times when Groups are involved in requests, see [NEXUS-2592](https://issues.sonatype.org/browse/NEXUS-2592).

This part was "done" but never tested and let into production (by default this feature is disabled, dormant), but you can enable it with "hidden" (read undocumented) switch. I did some preliminary tests, and stopped when I discovered similar bugs in HttpClient, that was since fixed (it was depleting sockets, as it was using pre-1.4 API as HttpClient 3.x we use is still 1.3 compatible).

This code above does the "parallel request" on behalf of given group, and if log would be inited as in your code, we could not differentiate what group was it, etc (all the requests would log to same log). But in general, I do agree with your statements, but in this very place, I do not feel it as "very problematic".

PR #361, opened 2012-04-02
NXCM-3965: Adding new constructor to NoSuchRepositoryException

cstamas:
There are cases when we do want to throw NoSuchRepositoryException
but the actual repository ID for some reason is unknown (and is
irrelevant actually).


nabcos:
+1

peterlynch:
Do we really want exception messages that may read "Repository with ID="null" not found!"?

cstamas:
@peterlynch What would give you that kind of message?

peterlynch:
https://github.com/sonatype/nexus/pull/361/files#L0R32 was my concern but this is nothing new
+1

PR #362, opened 2012-04-03
NEXUS-4968: Simple fix.

cstamas:
It seems Java running on VM Windows platform is known to
"over sleep" but also "under sleep". Also, in links below
it is mentioned that "slice size on Windows" is 10ms, but might
be worse when running in Hypervisor.

This fix simply raises the sleep time over the "slice size" to
make it more plausible to "over sleep" instead to "under sleep".

See issue comments and links in them:
https://issues.sonatype.org/browse/NEXUS-4968


nabcos:
+1

jdillon:
Shows exactly why time-based test mechanisms are faulty and should be avoided.

BTW, not suggesting not to merge this... since this test already relies on timing.  But I mentioned it to help point out that timing-based tests are generally poor and we should avoid them if possible in the future.  And/if time permits (ya right) rewrite timing-based tests that exist.

peterlynch:
+1

PR #363, opened 2012-04-03
NEXUS-4968: Proper fix.

cstamas:
In short, not using "millisecond timestamp" for change detection.

In old code, the timestamp was not used at all to get the date
(except in one DEBUG log, but that was irrelevant, as change
date will be same as timestamp of logline it sits on!). The pure
reason of the timestamp was to support "change detection", but
it was wrong, as millisecond resolution is too coarse: think
concurrent changes, or just like in related issue, UT issues
with Windows running on Hypervisor and sleeping less than OS
slice and having JRE "under sleeping". See links in the
issue comments.

Also, there is a "funkiness" of the DefaultFSLocalRepositoryStorage class
with regard of mixing Plexus and jsr330 (constructor injection in plexus
was never supported), I did not touch that, but did remove the wrong
Map injection as that map is not a component have to be modifiable.


nexus/nexus-configuration/src/test/java/org/sonatype/nexus/configuration/application/SimpleRemoteStorageContext.java
```diff
@@ -26,11 +26,20 @@
...
+    @Override
     public long getLastChanged()
     {
         return 0;
     }
 
+    @Override
+    public int getGeneration()
+    {
+        // TODO Auto-generated method stub
```
jdillon: What is all this "TODO Auto-generated method stub" for?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
```diff
@@ -60,7 +87,9 @@ public StorageContext getParentStorageContext()
 
     public void setParentStorageContext( StorageContext parent )
     {
-        this.parent = parent;
+        // noop
+        getLogger().warn(
+            "Class {} uses illegal method invocation, org.sonatype.nexus.proxy.storage.AbstractStorageContext.setParentStorageContext( StorageContext ), please update the code!" );
```
jdillon: Use of {} but no argument

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/storage/remote/commonshttpclient/HttpClientProxyUtilTest.java
```diff
@@ -153,7 +180,7 @@ public String getHostname()
         // get last changed
         long lastChanged = ctx.getLastChanged();
         // making sure we will have subtraction result grater than 0
-        Thread.sleep( 5 );
+        Thread.sleep( 15 );
```
jdillon, line 183: Tests should strive to avoid use of time based mechanism to validate functionality.

cstamas, line 183: Agreed, but this is the same test that exists in master (also, see [Simple fix](https://github.com/sonatype/nexus/pull/362) with same change). I just left the test in place -- just like the method call but made it deprecated.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
cstamas: Fixed, thanks.

nexus/nexus-configuration/src/test/java/org/sonatype/nexus/configuration/application/SimpleRemoteStorageContext.java
cstamas: The "Simple..." muck is removed.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
```diff
@@ -25,32 +26,58 @@
...
+            {
+                this.generation.set( parentGeneration );
+            }
         }
-        else
+
+        return generation.get();
+    }
+
+    @Deprecated
```
peterlynch, line 61: Why?

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
```diff
@@ -25,32 +26,58 @@
...
+    private final HashMap<String, Object> context;
 
-    private long lastChanged = System.currentTimeMillis();
+    private final StorageContext parent;
 
-    private StorageContext parent;
+    private AtomicInteger generation;
 
-    public AbstractStorageContext( StorageContext parent )
+    @Deprecated
```
peterlynch, line 35: Why? strange to deprecate a private member so you probably have a good reason ;)

cstamas, line 35: To have a visual feedback in IDE, that for change detection this field should not be used.

Change detection based on milliseconds are bad. Thats why we need to sleep in UTs, but also think concurrent config changes. Using millisecond based change detection omits to "detect" changes happened in same moment and/or concurrently.

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
cstamas, line 61: Left only to be backward compatible (Nexus Core is not using this field, but some plugin might).

Change detection should be done based on "generation" not on "change timestamp" (this is millisecond resolution)

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/AbstractRemoteRepositoryStorage.java
```diff
@@ -122,10 +122,8 @@ protected synchronized RemoteStorageContext getRemoteStorageContext( ProxyReposi
                     }
                     else
                     {
-                        getLogger().debug(
-                            "Remote context {} has been changed at {}. Previous change {}",
-                            new Object[] { ctx, new Date( ctx.getLastChanged() ),
-                                new Date( repositoryContexts.get( repository.getId() ) ) } );
+                        getLogger().debug( "Remote context {} has been changed. Previous generation {}",
+                            new Object[] { ctx, repositoryContexts.get( repository.getId() ) } );
```
jdillon, line 126: Should use 2 argument method now that only 2 {} exist (damn slf4j for not making these varargs already).

jdillon:
What does CI say?

cstamas:
CI scheduled:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/297/

cstamas:
CI voted +1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/storage/AbstractStorageContext.java
peterlynch, line 61: useful to add as a comment in source code?

peterlynch:
+1

adreghiciu:
+1

PR #365, opened 2012-04-05
NXCM-3986: Improve logs

cstamas:
There is a complex set of conditions when to log what, when to log
stack traces and when not to log stack traces. This would need a fix/cleanup
but for now, this change simply makes to log all HTTP 500 Internal
error responses _with_ stack trace, as previously they were logged
without it, and it did not make any sense and caused information loss.

This seems a long outstanding issue, as original commit implementing
this was unchanged for a long time, and it was causing that all
StorageException (that subclass IOException) gets logged _without_
stack trace, that is bad:

Change introduced in:
fcd4ccf26ad3e963d62e6e45ef06b8428f0d7608

We do have server side logs of the NXCM-3986, but as the cause was
not logged (no stack trace), we lost information why the rename
failed (clearly, it was some sort of FS related problem)

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
@@ -453,7 +453,7 @@ public class DefaultFSPeer
                         + " ms intervals " + hiddenTarget.getAbsolutePath() + " --> " + target.getAbsolutePath() );
 
                 throw new IOException( "Cannot rename file \"" + hiddenTarget.getAbsolutePath() + "\" to \""
-                    + target.getAbsolutePath() + "\"! " + e.getMessage() );
+                    + target.getAbsolutePath() + "\"! " + e.getMessage(), e );
             }
         }
     }
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
@@ -39,6 +39,7 @@ import org.sonatype.nexus.proxy.item.StorageItem;
 import org.sonatype.nexus.proxy.item.uid.IsItemAttributeMetacontentAttribute;
 import org.sonatype.nexus.proxy.repository.Repository;
 import org.sonatype.nexus.proxy.storage.UnsupportedStorageOperationException;
+import org.sonatype.nexus.proxy.utils.RepositoryStringUtils;
 import org.sonatype.nexus.util.ItemPathUtils;
 import org.sonatype.nexus.util.SystemPropertiesHelper;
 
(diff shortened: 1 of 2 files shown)


peterlynch:
AbstractResourceStoreContentPlexusResource#handleException - why not fix the javadoc on the way through(params)?
Also I was hoping that method at least had some unit tests, especially now we are changing what it catches

cstamas:
We are not changing what it throws, I merely just moved the catch block (literally), and am catching IOEx instead of StorageEx, that is a subclass of StorageEx...

cstamas:
ah... catches. Well, `StorageException` was already catch, and if you read whole method. `IOException` was going under the "else" branch. With this change, in case of IOException same happens, as it was before: `throw new ResourceException( Status.SERVER_ERROR_INTERNAL, t );` is executed... hm, this might need a cleanup too.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
```diff
@@ -453,7 +453,7 @@ protected void handleRenameOperation( File hiddenTarget, File target )
                         + " ms intervals " + hiddenTarget.getAbsolutePath() + " --> " + target.getAbsolutePath() );
 
                 throw new IOException( "Cannot rename file \"" + hiddenTarget.getAbsolutePath() + "\" to \""
-                    + target.getAbsolutePath() + "\"! " + e.getMessage() );
+                    + target.getAbsolutePath() + "\"! " + e.getMessage(), e );
```
jdillon: String.format() probably a little better here?

```
throw new IOException( String.format("Cannot rename file '%s' to '%s': %s",
    hiddenTarget.getAbsolutePath(), target.getAbsolutePath(), e.getMessage()), e );
```

cstamas: Fixed

cstamas:
Rebased against latest master.

cstamas:
Removed the duplicated logic of StorageException handling. Explained in commit message, but quoting it here too:

Before this fix, there was a separate handling of StorageException's
that was loosing stack trace. Also, superclass of StorageException
is IOException, and it was logged _with_ stack trace (in the else
branch). But both were creating a response of HTTP 500 internal error
(so, ultimately the only difference was the lack of the stack trace
in case of StorageException).

After this change, the StorageExceptions are handled in same way
as their superclass IOException, and will end up in logs _with_
stack trace too and unmodified HTTP 500 response code.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
peterlynch: -1 
String.format is evil sauce in logging - if you are trying to make it look like what slf4j does, suggest you write a wrapper string formatter that never could throw an exception if you mistyped a token.

peterlynch:
+1 other than direct use of String.format

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
jdillon: This isn't logging this is building an exception message.

peterlynch: well duh :) even worse then

jdillon: You are objecting to printf bitching if you gave it %something it didn't comprehend only?  Could either use the slf4j {} thingy or IMO better to use guava-like thingy that only knows about %s, though I don't think that is public we'd have to dup that code.

cstamas:
CI scheduled:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/299/

cstamas:
CI voted +1, with one intermittent (and unrelated) failure on Win slave

adreghiciu:
+1

PR #366, opened 2012-04-05
Read locking code moved to better place.

cstamas:
"Closer" to local storage. Also, we had issues regarding "wrong method invoked, hence
no locking in place" in relation to shadows and P2 repositories already. Moreover,
reading the code reveals same problems affects shadow, group and webSite repositories
too.

Hence, the whole block doing the locator swap is "moved down" to this method.

This pull is related to [NXCM-3986](https://issues.sonatype.org/browse/NXCM-3986) and it's pull request (that does not fixes anything but will give us more information on next failure) [#365](https://github.com/sonatype/nexus/pull/365), as it makes obvious some file locking problems are present (visible on Windows platform only, and the test did fail consistently on Win slave only), and code reading revealed the problem fixed in this pull: a "read" access to file contents was easily possible while UID locks were avoided (hence, UID locking in FSPeer does not help, as "counterpart", while reading did not acquire shared lock).

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
@@ -849,17 +849,6 @@ public abstract class AbstractRepository
                     throw new ItemNotFoundException( request, this );
                 }
             }
-            // plain file? wrap it
-            else if ( item instanceof StorageFileItem )
-            {
-                StorageFileItem file = (StorageFileItem) item;
-
-                // wrap the content locator if needed
-                if ( !( file.getContentLocator() instanceof ReadLockingContentLocator ) )
-                {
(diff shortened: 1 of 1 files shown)


nabcos:
CI?

cstamas:
Forgot it.
CI started:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/295/

cstamas:
CI voted +1

peterlynch:
nervous because I find no unit tests for AbstractRepository directly or javadoc on a lot of its methods

jdillon:
looked it over, but +0 since I don't know this subsystem well enough to know if this is a good change or not.

cstamas:
There are no UTs for abstract AbstractRepository class, but all the classes extending `org.sonatype.nexus.proxy.AbstractProxyTestEnvironment` UT class in Nexus Core are actually (implicitly) testing this class.

peterlynch:
+1 but for the record I am not going to try to make a habit out of liking classes with no direct UTs or javadoc

adreghiciu:
+1

PR #367, opened 2012-04-07
[NXCM-4046] Remove logback.xml included as resource, this is included in...

jdillon:
... the conf/ dir in the assembly and causes logback to complain.

https://issues.sonatype.org/browse/NXCM-4046


jdillon:
build passed https://builds.sonatype.org/job/nexus-oss-its-feature/296/

cstamas:
What happens now with the logging when Nexus deployed as WAR?

jdillon:
depends I suppose on where its deployed, if logback is already initialized, etc.

peterlynch:
guess we don't understand why it was added in first place @cstamas

jdillon:
IIUC it was added because the logback jars were exposed in the boot lib so that jetty can use the same logging.  And it needed a configuration file to control how it was initialized.

Removing it from the war avoids WARN and thus logback dumping its bootstrap logging details when Nexus boots.  When users are embedding via war they can sort out how to deal with whatever logging system the container they deploy into is using to get any bootstrap messages they want/need out.

cstamas:
This is wrong change. Read my full comment on ML.

jdillon:
well then fix it, I don't have time to read books you write in emails.  just fix it

cstamas:
Fix in [pull request 370](https://github.com/sonatype/nexus/pull/370)

Closing this pull.

PR #370, opened 2012-04-13
NXCM-4046: Fixing the logback warning at Nexus bundle boot

cstamas:
Nexus WAR contains the logback.xml, but also our "container"
has one too (on server classpath) as Jetty uses logback for
logging too.

The solution is not to take out the logback.xml from WAR,
as in that case we screw users of WAR, but instead, when
we _embed_ our WAR into our container (and have logging
already present on it), we simply leave out the WAR logback
config from WAR as logback is set up already at container
level, and during boot, Nexus will set up logging with it's
own means anyway.

This change modifies the module creating the _bundle_, by simply
leaving out the logback.xml from WAR when assembling. 
WAR remains intact and same as before.


peterlynch:
+1 looks sane - great

jdillon:
+1 just fix it already :-P

PR #371, opened 2012-04-13
Support filtering of repositories by user-agent

cstamas:
Replaces incomplete pull request:
https://github.com/sonatype/nexus/pull/104

In some use cases I have, I may have concurrent deployment of 2 versions of the same g:a.
For instance, 1.1 and 2.0 can be staged at the same time.

To be sure that 2 different staging repositories get allocated, I use a specific user-agent for deployment that is unique to each build.

This change allows to filter all operations of the nexus-maven-plugin using a user-agent.

US: https://issues.sonatype.org/browse/NEXUS-5017

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
+++ b/nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
@@ -37,7 +37,7 @@ public abstract class AbstractStagingMojo
 
     /**
      * If provided, and this repository is available for selection, use it.
-     *
+     * 
      * @parameter expression="${nexus.repositoryId}"
      */
     private String repositoryId;
@@ -46,12 +46,12 @@ public abstract class AbstractStagingMojo
 
     /**
      * Filter opened staging repositories using the provided user agent
-     *
+     * 
      * @parameter expression="${nexus.userAgent}"
      */
     private String userAgent;
 
-		public AbstractStagingMojo()
+    public AbstractStagingMojo()
     {
         super();
     }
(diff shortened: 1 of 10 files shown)


nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -102,12 +109,15 @@ protected void listRepos( final String groupId, final String artifactId, final S
             if ( groupId != null )
             {
                 repos = getClient().getClosedStageRepositoriesForUser( groupId, artifactId, version );
-                builder.append( String.format( " for: '%s:%s:%s'", groupId, artifactId, version ) );
+                builder.append( String.format( " for: '%s:%s:%s', user-agent: '%s'", groupId, artifactId, version,
+                    getUserAgent() ) );
```
peterlynch, line 113: fyi https://github.com/sonatype/sisu-goodies/blob/master/common/src/main/java/org/sonatype/sisu/goodies/common/SimpleFormat.java

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
```diff
@@ -340,4 +360,33 @@ protected MojoExecutionException logErrorDetailAndCreateException( final RESTLig
...
         return new MojoExecutionException( msg + ": " + e.getMessage(), e );
     }
+
+    /**
+     * Returns the list of staging repositories filtered by user agent. It returns the same list of no userAgent
+     * provided.
+     * 
+     * @param repos the list of staging repositories to be filtered.
+     * @return the filtered list of staging repositories or the passed in list if no userAgent explicitly defined.
+     */
```
peterlynch, line 370: javadoc could be cleaned up. There is no user agent 'passed in' for example, it comes from getUserAgent

peterlynch:
+1 in general

adreghiciu:
+1

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
cstamas, line 113: That's added 3 days ago? ;)

Sure, will convert to that or another "template" like this one:
https://github.com/cstamas/sisu-goodies/commit/abd24693a27295d61bdf51d5fa2c81876f0c3779

But right now, it would need sisu-goodies release and bump on Nexus, so will defer for now.

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/AbstractStagingMojo.java
cstamas, line 370: Um, the javadoc does not says user agent is "passed in". The list is "passed in", and userAgent is explicitly defined or not, but it does not implies it is a method input parameter or what.

PR #375, opened 2012-04-17
Smaller fixes

cstamas:
found during usual code reading.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeEventInspector.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeEventInspector.java
@@ -17,7 +17,7 @@ import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.util.StringUtils;
 import org.sonatype.nexus.proxy.events.AbstractEventInspector;
 import org.sonatype.nexus.proxy.events.EventInspector;
-import org.sonatype.nexus.proxy.events.NexusStateChangeEvent;
+import org.sonatype.nexus.proxy.events.NexusStartedEvent;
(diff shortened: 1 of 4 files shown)


nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeEventInspector.java
```diff
@@ -34,7 +34,7 @@ public boolean accepts( Event<?> evt )
         final String simpleName = evt.getClass().getName();
 
         // TODO: nexus-proxy module does not reference plugin manager, so this is a quick'n'dirty workaround for now
-        return evt instanceof NexusStateChangeEvent
+        return evt instanceof NexusStartedEvent
```
peterlynch, line 37: Why did this change?

This is the type of thing that should have a UT that self documents that this inspector responds only to _aaa_ events and does _xyz_ for such and such reason.

cstamas, line 37: See commit message of corresponding commit:
44474dc405dceeb661abac26d9c93d517de3c773

jdillon:
+something, okay one... though I don't understand the implications of the event change.

cstamas:
Wrt event change: the event inspector was listening to superclass of "nexus state event changed" events. When this was implemented, it was fine, as it had only 2 extending classes (NexusStarted, NexusStopped).

But since then, we added 2 more: NexusInited and NexusStopping..... and it just doubled the invocation count on the component, while it always needed only one: NexusStarted (and plugin loaded and unloaded, but it's unrelated to this change).

In these 3 cases the UID attributes registry "updates" itself, as plugins might contribute attributes, and NexusStarted is needed to load the "core" attributes. That's all folks.

PR #376, opened 2012-04-17
NEXUS-4922: Increasing timeouts.

cstamas:
The SO timeout ("socket timeout") is the one that made the regression
in the first place. It was set too low, for 5 seconds, while the
request on server side was processed without any error, but the processing
took 6 seconds and client timeouted with socket timeout (the setSoTimeout
for client).


peterlynch:
+1

nabcos:
+1

PR #377, opened 2012-04-20
o [NEXUS-5024] Set system properties on proxy settings change and fix problem reporting to use those.

nabcos:
That this ever worked was a side effect of the p2 plugin. p2 runtime sets the well-known http.proxy\* system properties, which got picked up by the xmlrpc transport. A change in that transport (newest swizzle uses async-http-client, which does not use those properties by default) broke proxy settings even when p2 plugin was present.

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/error/reporting/SetProxyPropertiesInspector.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/error/reporting/SetProxyPropertiesInspector.java
@@ -1,0 +1,151 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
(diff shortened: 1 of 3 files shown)


cstamas:
+1

peterlynch:
+1

nexus/nexus-app/src/main/java/org/sonatype/nexus/error/reporting/SetProxyPropertiesInspector.java
```diff
@@ -0,0 +1,151 @@
...
+            || evt instanceof NexusStartedEvent;
+    }
+
+    @Override
+    public void inspect( final Event<?> evt )
+    {
+        if ( !accepts( evt ) )
+        {
+            return;
+        }
```
jdillon, line 65: Why duplicate accepts() here?  In the flow of an accepted event accepts() will be called 2x instead of only one time.  I don't understand what the reason is folks guard inspect again with accepts().. I've seen this in a few places.

PR #378, opened 2012-04-23
NEXUS-4985: Added missing repoId

cstamas:


adreghiciu:
+1
Just checking: shouldn't we also return the repo id when the request is not for a group?

nabcos:
+1

PR #381, opened 2012-04-24
[NEXUS-4970] Skip merge of invalid metadata from group member repositories

adreghiciu:


adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/308/

cstamas:
This change is not enough. Take a look at org.sonatype.nexus.proxy.maven.metadata.operations.NexusMergeOperation class (the original reason of this issue). My point is, that exceptions are thrown _after_ partial execution (and some setter invocation are done) on targetMetadata.

Other point is, that we use Maven3 maven-repository-metadata dependencies, and luckily, those are clone-able.

So, ultimate point is, that neglecting the exception is not enough: the operation should be performed on a clone (cloned from result of previous operation), thus, if any Operation does partial update and then throws exception (like that one above, or any new added in the future), it still will not be able to leave metadata in "partially modified"/inconsistent state.

In general +1, but only after cloning to the changeMetadataIgnoringFailures() method is added to preserve consistency.

adreghiciu:
Build 308 failed due to not changed UT.
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/308/

adreghiciu:
@cstamas You are right. That was also in my mind but looking at the operation in cause I was thinking that the operations will fist check and only then update. Which is a false presumption.
I will change methods to clone metadata.

Wdyt, shall I do that also on changeMetadata original method?

cstamas:
Sadly, we cannot assume "sensible behavior" from these Operations, as nothing prevents us to implement them to be "not nice", like modify and then verify, instead the other way (to ensure target is not inconsistent). Safest is to protect target at high level.

Doing it on both places would make sense, yes.

adreghiciu:
@cstamas Working on clones is implemented, including UT

adreghiciu:
Kicked off a new build to include latest changes: https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/311/

adreghiciu:
One more try :) https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/312/

adreghiciu:
Seems like CI had some connection problems in 312.
One more try: https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/314/

nabcos:
+1

PR #382, opened 2012-04-25
[NEXUS-4984] Reading digest form a zero bytes length file should not crash

adreghiciu:
but return empty string

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
@@ -41,6 +41,11 @@ public class MUtils
         {
             String raw = StringUtils.chomp( IOUtil.toString( inputStream, "UTF-8" ) ).trim();
 
+            if ( raw != null && StringUtils.isEmpty( raw ) )
+            {
+                return "";
+            }
+
             String digest;
             // digest string at end with separator, e.g.:
             // MD5 (pom.xml) = 68da13206e9dcce2db9ec45a9f7acd52
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
@@ -41,7 +41,7 @@ public class MUtils
         {
             String raw = StringUtils.chomp( IOUtil.toString( inputStream, "UTF-8" ) ).trim();
 
-            if ( raw != null && StringUtils.isEmpty( raw ) )
+            if ( StringUtils.isEmpty( raw ) )
             {
                 return "";
             }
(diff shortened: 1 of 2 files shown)


adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/313/

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
```diff
@@ -41,6 +41,11 @@ public static String readDigestFromStream( final InputStream inputStream )
         {
             String raw = StringUtils.chomp( IOUtil.toString( inputStream, "UTF-8" ) ).trim();
 
+            if ( raw != null && StringUtils.isEmpty( raw ) )
```
cstamas: Why `raw != null && StringUtils.isEmpty( raw )`?

Why not `StringUtils.isBlank( raw )` only?

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/MUtilsTest.java
```diff
@@ -64,4 +69,18 @@ public void testAcceptedDigests()
...
+     *
+     * @throws IOException unexpected
+     */
+    @Test
+    public void zeroLengthSha1()
+        throws IOException
+    {
+        final File sha1 = util.resolveFile( "target/test-classes/nexus-4984/zero-bytes-length.sha1" );
+        final String digest = MUtils.readDigestFromStream( new FileInputStream( sha1 ) );
+        assertThat( "Zero bites checksum file", digest, is( equalTo( "" ) ) );
```
cstamas, line 83: Zero must have sharp teeth :D

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/maven/MUtils.java
adreghiciu: I did not want to interfere with chomp returning null. If you think that is fine is a piece of cake to change that

cstamas: just realized, that if `chomp()` returns `null` you are already doomed as it will NPE at `trim()`...

adreghiciu: Right. I will change the code to not have the null check

adreghiciu: Done.

cstamas:
:beer:

nabcos:
+1

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/MUtilsTest.java
peterlynch, line 83: bytes

peterlynch:
+1

PR #383, opened 2012-04-25
NEXUS-5022: Reimplementing the change.

cstamas:
But this time with fixes for UTs that failed after original change.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/utils/RepositoryStringUtils.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/utils/RepositoryStringUtils.java
@@ -25,4 +25,10 @@ public class RepositoryStringUtils
     {
         return String.format( "\"%s\" [id=%s]", repository.getName(), repository.getId() );
     }
+
+    public static String getFullHumanizedNameString( final Repository repository )
+    {
+        return String.format( "%s[contentClass=%s][mainFacet=%s]", getHumanizedNameString( repository ),
+            repository.getRepositoryContentClass().getName(), repository.getRepositoryKind().getMainFacet().getName() );
+    }
 }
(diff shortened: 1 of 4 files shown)


cstamas:
CI scheduled (another job still running, so this might be stil 404 if job not started)
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/315/

cstamas:
CI voted +1 (this time, _both_ slaves)

nabcos:
+1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/utils/RepositoryStringUtils.java
```diff
@@ -25,4 +25,10 @@ public static String getHumanizedNameString( final Repository repository )
     {
         return String.format( "\"%s\" [id=%s]", repository.getName(), repository.getId() );
     }
+
+    public static String getFullHumanizedNameString( final Repository repository )
+    {
+        return String.format( "%s[contentClass=%s][mainFacet=%s]", getHumanizedNameString( repository ),
+            repository.getRepositoryContentClass().getName(), repository.getRepositoryKind().getMainFacet().getName() );
```
peterlynch, line 32: -1
shouldn't this protect against null ? Also, please use the goodies helper class instead and stop using String.format

cstamas, line 32: This is same implementation (done in same way) as method above. NPE will happen in both, if you pass in `null`.

The Goodies are not yet released, and template was added in latest snap, so that would mean releasing Goodies, bumping it in nexus and then we could use it....

jdillon:
nexus core does not depend (yet) on goodies, lets not make that change to depend on it now.

goodies formatter won't help for npe anyways, only helps for invalid format + speed.

Lets merge this and move forward

cstamas:
As JD says, but also the fact that this is the _third_ method implemented in same way as previous two (in class RepositoryStringUtils), I believe this is not worth -1.

Merging in the change. In future, we will start using goodies in Nexus Core, but for now, it is not (yet) there.

peterlynch:
so the logic is we already have methods that can throw NPE, so lets not fix them and and instead copy and make another with same problem?

I could live without goodies for now, but adding a checkNotNull() in there wasn't a big stretch. At least in so doing we know what was null.

cstamas:
Agreed. But that would throw NPE instead of... NPE...
On May 11, 2012 5:41 PM, "Peter Lynch" <
reply@reply.github.com>
wrote:

peterlynch:
repository.getRepositoryContentClass().getName(), repository.getRepositoryKind().getMainFacet().getName()

if throws NPE, which is null - repository, ContentClass or main facet, or repo kind?

PR #386, opened 2012-04-29
NEXUS-4948 improve test logging and remove CounterListener after done

peterlynch:
We see on grid the testExpiration\* methods are fragile, I assumed it was due to asserting against async event propagation. ( but I may be wrong )

Added logging that will likely prove this in the case that this test fails again.

Also each time the private doTestExpiration was called a new CounterListener was being added to application event multicaster and not removed for the next test call. This wasn't directly breaking tests, but it wasn't helping either - since the next test calling this method would add another CounterListener instance into the chain, slowing things down even more. So now we explicitly remove any CounterListener added when we are through.

Still don't like the fact the testExpiration\* methods depend on eventing.

### DIFF
--- a/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/M2RepositoryTest.java
+++ b/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/M2RepositoryTest.java
@@ -303,7 +303,9 @@ public class M2RepositoryTest
     public void testExpirationAlwaysUpdate()
         throws Exception
     {
+        getLogger().info("Start testExpirationAlwaysUpdate() for '/spoof/maven-metadata.xml' at (" + System.currentTimeMillis() + ")");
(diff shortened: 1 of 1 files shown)


nabcos:
The events are not asynchronous IIUC - so the assertions should not have been hit while the CounterListener is still processing events.

But I agree, the test should be rewritten. I also have the feeling as it is right now it tests too much, way too many components involved here for something that maybe should only be testing `M2Repository#isOld`.

peterlynch:
But logging can't hurt us here though and may give more info?

peterlynch:
logging should show any time differences between when counter listener gets events and when the test asserts on its counter.

nabcos:
yes, seems like I forgot to add '+1'

peterlynch:
Going to also merge this to 2.0.x since we have duplicate issue NEXUS-5020

PR #389, opened 2012-05-02
Making references canonical.

cstamas:
Consumed JD's change but added more to it.

Changes from canonical-references branch are consumed, and added more to it.

AppContext bump is needed due to a bug in AppContext, that did not allow for a child to override a value if parent already had it.

This makes it almost 100% except the ".." bit in "nexus-work" in the dump at the boot (later, the canonical path is used, but Appcontext is unaware what those strings represent, hence, is not canonicalizing it).

### DIFF
--- a/nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
+++ b/nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
@@ -27,6 +27,8 @@ import org.apache.shiro.subject.Subject;
 import org.apache.shiro.util.ThreadContext;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
+import org.codehaus.plexus.personality.plexus.lifecycle.phase.Initializable;
+import org.codehaus.plexus.personality.plexus.lifecycle.phase.InitializationException;
 import org.codehaus.plexus.util.StringUtils;
 import org.sonatype.configuration.ConfigurationException;
 import org.sonatype.configuration.validation.InvalidConfigurationException;
@@ -71,6 +73,7 @@ import org.sonatype.plexus.appevents.ApplicationEventMulticaster;
 import org.sonatype.security.SecuritySystem;
 
 import com.google.common.base.Function;
+import com.google.common.base.Throwables;
 import com.google.common.collect.Collections2;
 
 /**
@@ -83,7 +86,7 @@ import com.google.common.collect.Collections2;
 @Component( role = NexusConfiguration.class )
 public class DefaultNexusConfiguration
     extends AbstractLoggingComponent
-    implements NexusConfiguration
+    implements NexusConfiguration, Initializable
 {
     @Requirement
     private ApplicationEventMulticaster applicationEventMulticaster;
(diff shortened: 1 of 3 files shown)


nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
```diff
@@ -437,34 +459,33 @@ public File getWorkingDirectory( final String key, final boolean createIfNeeded
             }
         }
 
-        return keyedDirectory;
+        return canonicalize( keyedDirectory );
     }
 
-    public File getTemporaryDirectory()
+    public synchronized File getTemporaryDirectory()
```
peterlynch, line 473: why sync?

nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
```diff
@@ -437,34 +459,33 @@ public File getWorkingDirectory( final String key, final boolean createIfNeeded
...
         if ( temporaryDirectory == null )
         {
-            temporaryDirectory = new File( System.getProperty( "java.io.tmpdir" ) );
-
-            if ( !temporaryDirectory.exists() )
+            final File file = new File( System.getProperty( "java.io.tmpdir" ) );
+            if ( !file.exists() )
             {
-                temporaryDirectory.mkdirs();
+                file.mkdirs();
```
peterlynch: What if mkdirs returns false? Javadoc here should document the attempt at dir creation, retry a few times on 'false', and if still can't make it - log a WARN messages since a temp is usually crucial to operation.

nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
```diff
@@ -437,34 +459,33 @@ public File getWorkingDirectory( final String key, final boolean createIfNeeded
...
         if ( configurationDirectory == null )
         {
-            configurationDirectory = new File( getWorkingDirectory(), "conf" );
-
-            if ( !configurationDirectory.exists() )
+            final File file = new File( getWorkingDirectory(), "conf" );
+            if ( !file.exists() )
             {
-                configurationDirectory.mkdirs();
+                file.mkdirs();
```
peterlynch: same comments here as for getTemporaryDirectory()

nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
cstamas, line 473: Because this method modified internal state of the component.

nexus/nexus-app/src/main/java/org/sonatype/nexus/configuration/application/DefaultNexusConfiguration.java
cstamas: The javadoc should be inherited from `org.sonatype.nexus.configuration.application.ApplicationConfiguration` that does documents the dir creation.

But agreed, it should be handled somehow with the log, even if we did not handle it so far.

cstamas:
CI scheduled:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/330/

cstamas:
CI voted +2

jdillon:
looks okay to me

PR #390, opened 2012-05-07
NXCM-3951 product name cleanup

peterlynch:
add some tests and fix product name use in logs and web.xml


peterlynch:
no news is good news - given if this breaks some name rendering - it is easily noticed I am merging this.

PR #395, opened 2012-05-16
Use velocity comments

jdillon:
Uses #\* *# velocity comments to avoid rendering copyright headers.  Reduces ~785 pointless bytes from response stream.


nabcos:
there's config missing for the license plugin, these comments will be overridden in the next run of license:format. see dcb71a895b6e4c79f5519ef3739bfbe84bc61458, but that's still on a branch.

cstamas:
:beer:

jdillon:
@nabcos why was this change on your branch and not on master?

A feature branch isn't really a back door to merge in unrelated changes, which this appears to be... and while I think its a good change I think it should have been done on master.

PR #396, opened 2012-05-17
Make our modules (especially plugins) more resilient to changes.

cstamas:
DummyWhatever implementation was just dumb. So, using Mocked
implementation of the given interface to make this plugin
more resilient to newly added methods to the interface, that
is NOT in this module (but somewhere else).

This will be even more important once we explode the build.


adreghiciu:
+1

PR #398, opened 2012-05-18
Nxcm 4077 nexus download settings usertoken support

jdillon:
Adds support for pluggable interpolation, and includes $[userToken.] bits.

Pending support to encrypt w/master password muck.


jdillon:
Master-password bits are now in, some details up here:

https://docs.sonatype.com/display/Nexus/User+Token+Testing#UserTokenTesting-nexus%3Asettingsdownload

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/settings/usertoken/UserTokenDTO.java
```diff
@@ -0,0 +1,46 @@
+//
+// This file was generated by the JavaTM Architecture for XML Binding(JAXB) Reference Implementation, v2.2.4 
+// See <a href="http://java.sun.com/xml/jaxb">http://java.sun.com/xml/jaxb</a> 
+// Any modifications to this file will be lost upon recompilation of the source schema. 
+// Generated on: 2012.05.18 at 02:12:38 PM PDT
```
nabcos: This does not look like it's still generated source. Needs a license header.

nabcos:
+1

nexus-maven-plugins/nexus-maven-plugin/src/main/java/org/sonatype/nexus/plugin/settings/usertoken/UserTokenDTO.java
jdillon: Thx, didn't notice (it started out as a copy, then I just removed all that muck)

jdillon:
Minor changes to fix headers and ensure default settings from properties file are configured last when interpolating.

peterlynch:
a few of these new classes could use some unit tests.

jdillon:
BTW, I had been waiting on most work on tests until @brianf could review and validate this is indeed what the spec was intending, which I think he did yesterday.

jdillon:
I've added a basic test for the customizer and interpolation bits.  Not sure there is much value in more UT for the other bits.  I'd add one for the encryption stuff except that the DefaultSecDispatcher in all its Plexus glory has private field injection of dependencies making it rather cumbersome to test with... and the code for the MasterPasswordEncryption stuff is/should be as close to what is in Maven, so I'd rather not refactor it much/or at all.

nexus-maven-plugins/nexus-maven-plugin/pom.xml
```diff
@@ -358,7 +345,13 @@
...
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-interpolation</artifactId>
-      <version>1.13</version>
+      <version>1.15</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.sonatype.sisu.siesta</groupId>
+      <artifactId>siesta-client</artifactId>
+      <version>1.1-SNAPSHOT</version>
```
peterlynch, line 354: just in case siesta is now staged 1.1 and snap bumped to 1.2-SNAPSHOT

peterlynch:
+1

PR #399, opened 2012-05-19
[NEXUS-5060] Avoid JS error for hosted repositories mirror tab selection

jdillon:
Avoids JS error, not sure if this is a proper fix to the problem however, as in not sure if there should have been a predefinedMirrorDataSource for HostedMirrorEditor or not.


adreghiciu:
+1

nabcos:
+1, AFAICT hosted repos never had predefined mirrors.

PR #402, opened 2012-05-22
o [NXCM-4156] Disable or hide fields when user is not from default source

nabcos:


jdillon:
+1

Only suggestion I have is to make a:

```
// if the user-source is not the default value (ie. external ldap realm or something)
var nonDefaultUserSource = Sonatype.user.curr.loggedInUserSource !== 'default';
```

and then use:

```
disabled: nonDefaultUserSource
```

or:

```
hidden: nonDefaultUserSource
```

Slightly optimizes to avoid having go inspect the Sonatype.user.curr.loggedInUserSource property each time and is also less verbose (probably making slightly more readable too).

PR #403, opened 2012-05-22
NEXUS-5049: Partial backport to Nexus 2.0.x

cstamas:
This is the "C" change from the overall changes made as
part of issue.

It simply _bans_ username used for AA when AA is disabled.

Locally built, UT/IT pass.


peterlynch:
+1

PR #409, opened 2012-05-24
Extjs form field value handling

dbradicich:
We have had a long outstanding issue (not a user based issue, but dev based) where the 'value' field in a Field instance was never being set as the default value.

Finally found the problem, we have an override of the afterRender function, and we weren't calling the proper parent


adreghiciu:
+1

cstamas:
:+1:

PR #414, opened 2012-05-29
NXCM-3970: Bugfix.

cstamas:
The goal was not working when the Mojo was not configured in the
POM (the "do not modify POM" case).

In case when POM was modified, and hence, the Mojo was not registered to run, this Mojo actually never kicked in (reported by JD).


jdillon:
I don't get why its not a simple aggregator goal that you point at the target/nexus-staging directory and it uploads.  why does it need to traverse all projects in a reactor?

cstamas:
Just for the record (as we had this conversation over the phone): traversal happens due to "aggregator case" (see [wiki](https://docs.sonatype.com/display/Nexus/Staging#Staging-AggregatorPOMs)).

Also, when you _explicitly specify the location of staging directory (the Mojo does not have to "figure it out")_, it will be used and there will be nothing done in traversal. Both mojos, the `deploy` and `deploy-staged` accepts the `altStagingDirectory` parameter.

nabcos:
+1

PR #418, opened 2012-06-04
NEXUS-5080: Bumping SISU Jetty8 to latest.

cstamas:
This version forces this setting (to fail Jetty when a webapp within
fails).


nabcos:
+1

PR #423, opened 2012-06-05
NXCM-3923: Extending DTO with new field.

cstamas:
This DTO is used to render the "index" page (when browsing slash-content), but seems misused in other parts too.

The new field make possible to "decorate" the item with extra (possibly lengthy) description, but leave to consumer of this DTO to render it in a way it wants.


nabcos:
+1

PR #424, opened 2012-06-05
NEXUS-5043: decode the path

cstamas:
This problem was never hit as "non standard" characters are rarely used
in GAV coordinates, and before, even if they were used, Maven2 by default
used lightweight wagon that does not encoded paths.

Today, Maven3 defaults to wagon-http that uses Apache HttpClient that does
encode paths if a "prohibited" character is used.


adreghiciu:
+1

nabcos:
+1

nabcos:
actually is there a test using paths with encoded character?

cstamas:
UT added covering this

PR #425, opened 2012-06-05
[NEXUS-4996] Allow mix case in email top level domain

adreghiciu:


nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/global/SmtpSettingsValidationPlexusResource.java
```diff
@@ -131,4 +128,13 @@ public Object put( Context context, Request request, Response response, Object p
         return null;
     }
 
+    static void validateEmail( final String email )
+        throws ResourceException
+    {
+        if ( !EMAIL_PATTERN.matcher( email ).matches() )
```
peterlynch, line 139: null email will fail here. lets prevent potential NPE, even though theory is UI will never pass it.

adreghiciu, line 139: I just extracted the method so it is testable. I could check for null, without any problem.

nabcos:
+1

adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/350/

nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/global/SmtpSettingsValidationPlexusResource.java
```diff
@@ -131,4 +129,17 @@ public Object put( Context context, Request request, Response response, Object p
         return null;
     }
 
+    static void validateEmail( final String email )
```
cstamas, line 132: I would move this method out of REST layer (into NexusEMailer?) to make it reusable.... it was totally wrong to have it here in 1st place

adreghiciu, line 132: NexusEmailer is an iface. Maybe to some util class in same package?

adreghiciu:
CI voted +1 (The failures were unrelated)

cstamas:
+1 merge this for now, and will move stuff out to emailer, as it is pending release anyway

PR #434, opened 2012-06-06
Merge nexus plugin archetype

adreghiciu:
Merge https://github.com/sonatype/nexus-plugin-archetype repository into nexus git repository (including history).
This will help us keeping the plugin archetype (somehow) in sync with Nexus and releasing them at same moment.


cstamas:
+1

cstamas:
-1 after reconsideration...

cstamas:
This should be merged under location /nexus of nexus core repository, not here.

The only "link" with Nexus Core is actually this part

```
    <requiredProperty key="nexusVersion">
      <defaultValue>${project.version}</defaultValue>
    </requiredProperty>
```

But this also means it should share "same release lifecycle" with Nexus Core (and as part of it, `${project.version}` would be always in sync.

By putting it under /nexus-maven-plugins (and the parent POM over there) would _might_ conflict with later partitioning of the build. Moreover, I am just now in the situation am about to _change_ the nexus-maven-plugins parent POM to _not_ inherit `/nexus/pom.xml` as parent but rather forge directly, as nexus parent contains nexus-typical depMgt entries conflicting with Maven Plugin needs. And if you think about it, it does make sense, as Plugins will usually "couple" over REST to Nexus and not directly with Nexus Core JARs...

cstamas:
and ultimately, this is _not_ a maven plugin, will _not_ share "common maven plugin" dependencies, etc...

adreghiciu:
The nexus-plugin-archetype module is now under root (/nexus)

cstamas:
+1

peterlynch:
+1 great

PR #436, opened 2012-06-08
o [NXCM-4362] Fix error dialog message on password change

nabcos:


peterlynch:
So we will always get a useful error message in the response? I am thinking the http status was included before because sometimes there was not an otherwise useful message from backend ( 500 error ?) . Maybe display the http status if we can determine there is nothing user friendlier in the response.

adreghiciu:
+1

nabcos:
server status line will now always be displayed when no message was extracted from the error response.

peterlynch:
+1

cstamas:
+1

PR #437, opened 2012-06-08
NEXUS-5033: Exposing Aether as shared from MI Indexer plugin.

cstamas:
As Maven Indexer public API contains Aether classes, they have to
be shared too, to make it able to completely cover development
against it's APIs and avoid problems like Linking Errors or
other ugly circumventions.

Changes happened:
- exposed aether-api and aether-util as "shared" from MI Plugin
- exposed maven-indexer-artifact to publish complete MI API
- plexus-utils made "provided", as it was included in plugin but
  artifact is present in core. Did not hurt (core wins), but was bloating
  the plugin size.


adreghiciu:
+1

peterlynch:
+1 trusting

peterlynch:
+1 marked as release note

PR #441, opened 2012-06-11
NXCM-4366: Removed Index customizer

cstamas:
Groovy has no UI/JS elements (was planned to add, but never done)
Hence, the referenced JS is not present in the plugin either,
as it's not implemented at all.


peterlynch:
+1

adreghiciu:
+1

PR #442, opened 2012-06-11
o [NXCM-4120] Make user profile tabs bookmarkable

nabcos:


adreghiciu:
+0 (I have no clue what the JS does)

peterlynch:
+1

nexus/nexus-webapp/src/main/webapp/js/nexus/error/ErrorHandler.js
```diff
@@ -56,7 +56,7 @@ Nexus.error.ErrorHandler = function() {
         args[x] = arguments[x];
       }
       try {
-        if ( '?debug' === windows.location.search )  {
+        if ( '?debug' === window.location.search )  {
```
peterlynch, line 59: what does this affect?

peterlynch:
so will #profile without additional ";$viewName" go to Summary panel by default? Otherwise how will NXCM-3842 be fixed?

nexus/nexus-webapp/src/main/webapp/js/nexus/error/ErrorHandler.js
nabcos, line 59: Whoops, would have been best in its own branch. That's the js error dialog popup you can get with `index.html?debug`, enabling it did not work.

nabcos:
The first registered view will be selected, and that's the Summary view.

PR #444, opened 2012-06-11
Use jline directly to prompt for password.  plexus-interactivty is not d...

jdillon:
...oing this correctly and its too much hassle for something so damn simple to go and sort out all of that atm.

---

Short of going and fixing all the plexus-interactivty bits, this should work around the problem of showing passwords when prompting.


nabcos:
+1

jdillon:
Barf, I may have to sort out the plexus component, as the tests are using a fake prompter for testing.

jdillon:
Updated to make a Prompter component which only handles the password bits, delegates the rest to default impl.  No common sub-class here to pick up support bits, had to delegate.

adreghiciu:
+1

cstamas:
+1

PR #446, opened 2012-06-12
o [NEXUS-5064] Enable connection pooling for ssl ldap connections as well

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/362/

### DIFF
--- a/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java
+++ b/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java
@@ -20,6 +20,8 @@ import javax.naming.NamingException;
 import javax.naming.ldap.InitialLdapContext;
 import javax.naming.ldap.LdapContext;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.shiro.realm.ldap.AbstractLdapRealm;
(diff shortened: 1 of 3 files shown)


adreghiciu:
+1

cstamas:
+1

peterlynch:
-1 as is
no unit tests for setting those env props
plus the javadoc for setUsePooling(boolean usePooling)  is now wrong

cstamas:
Well, as far as I see, according to this

http://docs.oracle.com/javase/jndi/tutorial/ldap/connect/config.html

The change looks okay. Do we want to test Sun JNDI? ;)

peterlynch:
um no that would be integration test...I mean given xyz conditions, ensure pooling is _configured_

too bad... if we wrote a unit test for this stuff  - perhaps we would notice that searchBase is checked for not null, but then never used...

https://github.com/sonatype/nexus/blob/master/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java#L211

nabcos:
I actually think Nexus only uses the PlexusLdapContextFActory, which IIRC seems to have been a copy of the DefaultLdapContextFactory

peterlynch:
Is DefaultLdapContextFactory being used anywhere? Hopefully there is a good reason we have both.

peterlynch:
Anyone know why we have both context factories?

nabcos:
It looks like `PlexusLdapContextFactory` was to be removed, comment from [BasicActiveDirectoryTest](https://github.com/sonatype/nexus/blob/master/nexus/nexus-core-plugins/nexus-ldap-plugin-parent/ldap-common/src/test/java/org/sonatype/security/ldap/realms/BasicActiveDirectoryLdapSchemaTest.java#L75):

```
    // FIXME: this test is not autmated, and now it is BROKEN, but this needs bo be resolved, as the
    // PlexusLdapContextFactory has been removed
```

This test was aparently not exercised for a long time, it is excluded in the POM and manually running will give nothing but NPEs.

EDIT: Not even NPEs, but there's apparently the ApplicationEvent muck missing from classpath, tests do not even set up correctly.

peterlynch:
thx for the test

PR #447, opened 2012-06-13
o [NXCM-4203] Add refresh tool to sec/users detail tabs

nabcos:
The tool is only visible if the active tab has a `refreshContent`
method, the same as for the user profile refresh bits.


cstamas:
+1

peterlynch:
+1

adreghiciu:
+1

PR #454, opened 2012-06-20
NXCM-4391: Reformat

cstamas:
No significant code change (there is but minor),
bit this is actually about reformat.

The whole issue NXCM-4391 seems to "misfire" (as it seems it's not the wastebasket), but since I already tied up this source, let's push it.


nabcos:
+1

adreghiciu:
+1

PR #455, opened 2012-06-22
NEXUS-5109: fix

cstamas:
Simply, using Plexus Dispose lifecycle, to purge the EHCache (this
under the hud actually removes only this repositorie's entries since
shared caches implemented).


adreghiciu:
+1

nabcos:
+1

PR #456, opened 2012-06-22
Update to jline 1.0 which should support windows 32 and 64 bit

jdillon:
Update to jline 1.0 to get windows 64 bit dll support.

### DIFF
--- a/nexus-maven-plugins/nexus-maven-plugin/pom.xml
+++ b/nexus-maven-plugins/nexus-maven-plugin/pom.xml
@@ -188,10 +188,19 @@
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


jdillon:
Doesn't appear that jline 1.0 is functioning at all :-(

jdillon:
Maybe it is working, but not in a "new-style" cygwin terminal... yup.  Old style and cmd.exe seems to be okay.

adreghiciu:
+1

nabcos:
+1

peterlynch:
+1
We can live with new style cygwin terminal not working. Main thing Windows users have cmd prompt to use and it works.
If jline fixes new cygwin terminal support in future, we'll pick it up then.

PR #457, opened 2012-06-22
[NXCM-4395] Add jsoup to handle parsing out non-OK response entities int...

jdillon:
...o something more reasonable to present to users.

https://issues.sonatype.org/browse/NXCM-4395


adreghiciu:
+1 beside system.out

nabcos:
+1

peterlynch:
I'd like to see a test for this. As it is I have to manually go test what it does.

cstamas:
+1

PR #460, opened 2012-06-25
Move indexing related tasks descriptors into indexing plugin

adreghiciu:
When we extracted legacy indexing as a plugin we forgot to move the indexing related tasks descriptors alongside the tasks. This means that if the indexing plugin is removed from plugins repository, the tasks UI will still show the indexing tasks but when saving it will result into 500 (server error) as the tasks that implements the functionality are not available.


adreghiciu:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/363/

cstamas:
:beer:

adreghiciu:
CI voted +1 (unrelated Win failure)

jdillon:
+1

peterlynch:
+1
I'd be very interested to know if we can write a general test that makes sures no turds are left behind like this. This is the second or third time I saw this type of mistake.

PR #461, opened 2012-06-25
[NEXUS-5110] Show scheduled tasks menu item just based on the fact that user has read privilege

adreghiciu:


cstamas:
+1

jdillon:
+1

nabcos:
+1

peterlynch:
operations accessible from this menu are protected on server side as well?

adreghiciu:
Dunno, but would be stupid not to be ;)

More, the operations are not permmited from UI as add/delete/save buttons are disabled when no rights

PR #463, opened 2012-06-27
o [NXCM-4389] Set up favicon cross-browser compatible

nabcos:


peterlynch:
can we have a test to make sure favicon can be fetched? I think this has regressed a few times since I have worked on Nexus.

jdillon:
.png what browsers use a .png file for favicon?

nabcos:
http://www.kevinleary.net/how-to-add-cross-browser-favicon-website/
-> "Cross Browser Approach" is what I followed

nabcos:
https://builds.sonatype.org/view/nexus/job/nexus-oss-its-feature/364/

peterlynch:
+1

PR #464, opened 2012-06-27
NXCM-4391: not a fix, a "best effort" instead

cstamas:
This change is NOT fixing the cause of OOM at all, but improves
Jetty's capability to at least warn in logs about it.

The original reason for issue NXCM-4391 was another bug (OOM in P2),
but Jetty's handling of Errors (OOM is Error) made us realize that there was
NO any notification about death and removal of the thread. The
OOM went totally unnoticed (original investigation started due to
unreleased locks, there were held by "no one", as the thread holding was
dead actually).

Again, this change only enables "some" notification about extreme
conditions (like thread dying coz of some Error, as Exceptions
ARE logged by Jetty). Only "some", as in OOM state strange things
happens, but still, this is "best effort" to say "hey dude, you'd
better reboot your JVM" at least.

P2 issues (the actual cause of OOM that lead us to this recognition)
are handled in P2 plugin (finally blocks cleanup and memory optimization).

### DIFF
--- a/nexus/nexus-oss-webapp/src/main/resources/content/conf/jetty.xml
+++ b/nexus/nexus-oss-webapp/src/main/resources/content/conf/jetty.xml
@@ -23,6 +23,9 @@
 <!--                                                              -->
 <!-- ============================================================ -->
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
+    <Set name="threadPool">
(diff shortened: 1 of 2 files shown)


peterlynch:
can you release sisu-jetty and use the release version instead. versions are cheap

cstamas:
Released sisu-jetty8-1.3 and using that.

PR #465, opened 2012-06-28
o [NXCM-4426] Set index.html to be rendered with IE8 engine

nabcos:
Ext JS 2.3 is not supported for IE9 anyway.


cstamas:
+1

peterlynch:
+1

jdillon:
why not

PR #466, opened 2012-06-28
o [NXCM-4423] add Nexus.log method, let '?debug' turn on logging

nabcos:


nexus/nexus-webapp/src/main/webapp/js/nexus/common.js
```diff
@@ -0,0 +1,26 @@
...
+/*global Ext, Nexus, Sonatype, console*/
+
+Ext.namespace('Nexus');
+
+/**
+ * @static
+ * @param msg {String} The message to log.
+ */
+Nexus.log = function(msg) {
+  if (Sonatype.config.isDebug && console) {
```
peterlynch: shouldn't you check config is not undefined? Actually I forget if that is important

nabcos: it may be undefined if one of the extjs extensions or utility method call Nexus.log _on setup_, not when they are used on runtime of the Nexus UI.

jdillon:
+1

cstamas:
+1

PR #467, opened 2012-06-28
o [NXCM-4422] Disable profile summary's user id field

nabcos:
One-line fix in     7229d84, introduced by a change to make external users read-only.


peterlynch:
+1 ok move issue to 2.1 done once merged

PR #468, opened 2012-06-28
Make possible to configure when nexus is started/stopped

adreghiciu:
As in for each method or once per class.


jdillon:
+1

nexus/nexus-launcher/src/test/java/org/sonatype/nexus/bundle/launcher/NexusStartAndStopStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+@Inherited
+@Documented
+public @interface NexusStartAndStopStrategy
+{
+
+    Strategy value() default Strategy.EACH_METHOD;
+
+    public static enum Strategy
+    {
+        EACH_TEST, EACH_METHOD;
```
peterlynch: Wouldn't it make sense to document what these mean?

peterlynch:
I hate trying to use test code that is not doc'ed. Other than that +1

PR #471, opened 2012-07-02
NEXUS-5107: Wait for event inspectors too

cstamas:
Code doing server side config changes did wait for tasks, but did
not wait for (async) events, that are sometimes invoked _after_ or
_by_ some tasks.

While this is not a direct fix, this only assures that both, tasks
and async event inspectors as "calm" after method returns from
performing a config change.

But, why Nexus977 IT extends AbstractNexus1923?


nabcos:
+1

PR #473, opened 2012-07-05
o [NXCM-4428] Hide the password label for external users

nabcos:


peterlynch:
+1 not sure this is even worth a pull request.

nabcos:
True, didn't really think about it.

PR #475, opened 2012-07-09
o [NXCM-4425] Turn off calculation of page scroll offset for IE

peterlynch:
Main Nexus UI does not scroll ever, so this calculation is unnecessary
and slows down UI noticably (300ms<>1ms for a privilege detail view in
my tests). This is still a bad hack, but fortunately other browsers are
not affected.

[Peter] This fix improves performance in my tests to match other browsers with no apparent


peterlynch:
this fix was tested by @nabcos and @peterlynch can't imagine others have a problem with it

PR #476, opened 2012-07-10
Merge nexus client and lucene indexer client into nexus oss

adreghiciu:
The reason behind this merge is to have nexus lucene client where it belongs (near the plugin & model) and so be consistent with the other models and clients. But since client depends on nexus-client-core we have to move also this into nexus oss repository.

Beside being forced to move nexus-client-core in my view nexus oss tree is anyhow the place it belongs, at least until w split the build.

This change will also reduce the number of repositories that have to be built.


adreghiciu:
There are a number of small pull requests that can be merged once this is merged:

https://github.com/sonatype/nexus-enterprise/pull/233
https://github.com/sonatype/nexus-capabilities-plugin/pull/10
https://github.com/sonatype/nexus-staging-plugin/pull/23 
https://github.com/sonatype/nexus-maven-plugins/pull/3

peterlynch:
+1

nexus/nexus-client-core/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+          <groupId>com.thoughtworks.xstream</groupId>
+          <artifactId>xstream</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-library</artifactId>
```
peterlynch, line 66: odd that rest model uses hamcrest

PR #477, opened 2012-07-11
o [NXCM-4426] Update BrowseButton extension

nabcos:
Previous version of this extension had rendering problems in IE8 document mode, showing the file input button following the pointer on e.g. the artifact or license upload buttons.


peterlynch:
:beer: looks magic!

PR #482, opened 2012-07-16
o [NEXUS-4064] Fix menu actions to display when a repo does not publish the URL

nabcos:


cstamas:
+1

peterlynch:
+1 so now all menu items display despite url not being published?

nabcos:
Yes, there is no difference repos with published/unpublished URL.

There is one weird behavior that's no regression:
For 'unpublished' repos, you get no context menu in the lower pane, when you right click on artifacts. The method that assembles the menu is never called, didn't want to debug that just now. (It's the same behavior for 2.1/2.0.6).

peterlynch:
> For 'unpublished' repos, you get no context menu in the lower pane, when you right click on artifacts. The method that
> assembles the menu is never called, didn't want to debug that just now. (It's the same behavior for 2.1/2.0.6).

Ok. Issue filed for that? we can push that out in another release.

nabcos:
https://issues.sonatype.org/browse/NEXUS-5146

PR #488, opened 2012-07-19
NXCM-4504: Ability to use Verifier with Maven2 and Maven3

cstamas:
Verifier does not limit maven versions to be used, but how we integrated
Verifier with ITs did.

So, this is "liberation" of Maven Verifier, while keeping old behavior for old
ITs.

The module using different version of Maven (other than ITs default 2.2.1) would still need to get it and unpack it. This change is only about _the possibility_ to use other version, but does not force you to do so.

---

Note: while commits in this pull are marked as NXCM-4455, there was a confusion on my side. Read comment in [NXCM-4504](https://issues.sonatype.org/browse/NXCM-4504) why this happened.


peterlynch:
+1 decent docs!

adreghiciu:
+1

PR #489, opened 2012-07-20
NEXUS-5166: Fix

cstamas:
The original fix made no sense, as there was never -1 arriving here, but
merely -60 (the configuration value on UI was multiplied by 60 in a method
calling up to here). Also, EHCache 2.x became more stricter with negative
TTLs than EHCache 1.x was, so before EHCache version bump this was not
an issue.

Now, if you set on the UI the field "Not Found Cache TTL" to any negative
number, the NFC Element will not have it's own TTL but global cache
default TTL will be applied, which is in case of default settings "never".
Still, the element will be evicted from the cache when it fills up, and
eviction algorithm (I think default is LFU) finds it as good candidate
for eviction.


rseddon:
+1

I've been testing with this, and with -1 set I don't see any outbound request until the cache overflows.

jdillon:
+1

cstamas:
Thanks Peter!

PR #490, opened 2012-07-23
NXCM-4516: Walker fix

cstamas:
When Walker invoked with starting path that was not pointing to
collection, it simply returned doing nothing. Otherwise, it would
recursively "dive" in.

This simply fix contains a small refactoring, pulling out
the "walkItem" part, and in main entry point, if the starting
item is not a collection, then direct call to this method is made.

UT added covering this case, that asserts that walk does happen
to the 1 targeted file item.


cstamas:
CI
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/372/

nabcos:
+1

adreghiciu:
+1

peterlynch:
+1

cstamas:
CI voted +1, merging this.

PR #491, opened 2012-07-24
Add new methods that allows creation of repositories with specified artifact/metadata max age

adreghiciu:
Still, keep the old behavior that was not setting this values to not break existing ITs


nabcos:
+1

cstamas:
Note: this will work in current (broken) way, and "use default" implies defaults will be applied which the value `0` is not.

Maybe something like "backward compat and broken" ?

nabcos:
Just rename it to USE_BROKEN =D

cstamas:
+1 with rename :D

PR #494, opened 2012-07-27
Swap black magic to white magic

cstamas:
Introduction of "noit" (and "nout") standard profiles.
Changes:
- by default "mvn clean install" behaves as expected (uniform builds)
- to build quickly with no ITs "mvn clean install -Pnoit" (there is nothing funky, like double negation of presence of a "quick" sysprop etc)
- to build quickly as hell "mvn clean install -Pnoit,nout" (ditto, as above except UTs are skipped too)
- removed all the funky magic (like deps added when no quick build performed etc)

All these _produce valid and complete builds_, no more builds without documentation or wrong deps used (like those with tests classifier).

Advice:
- developers should set the "noit" profile to be activated by default in their settings.xml on their workstations.
  Hence, they would need only "mvn clean install -Pnout" for ahem quick builds
- release should be performed by doing simply "mvn release:prepare release:perform"


nabcos:
the quick sysprop turned off compilation of test classes and packaging of test jars as well

cstamas:
And that was one of the problems: you did not compile them ("dead code"), and the inter-module dependencies were not correct either (as some module depended on another's "-tests" one, but it was either present in your local repo from prev build or was downloaded -- again, from another build).

That was incorrect, as following those as "guidelines", we could skip even the maven-jar-plugin and who knows what else from the build. But question is, what is the result of such a build? Also, we talk about how much time difference?

cstamas:
This here is not trying to _reimplement_ the fundamentally broken "quick" builds, but to provide a _more correct solution_ instead, while retaining the _correct output of the build execution_.

nabcos:
nexus oss was "clean" in the sense that no -tests dependencies were produced or needed. in nexus-enterprise nuget plugin was not set up correctly and would download -tests IIRC. 

with the quick build I was trying to find the minimal set of things to build to get a working nexus dist as fast as possible. that's why I wanted tests and tests jars skipped as well.

cstamas:
Closing as changes will income.

PR #495, opened 2012-07-30
o [NEXUS-5158] Show login window when anonymous is disabled and user is not logged in

nabcos:
UI will show the (not closable) login window if it receives 401 for the status resource.
If it receives any other error, it will display `Warning: Could not connect to Nexus.` on the welcome page.


cstamas:
+1

adreghiciu:
+1

PR #500, opened 2012-08-01
o [NEXUS-5193] Fix LDAP URL check

nabcos:
Precondition was introduced when creating tests to the ldap context factory, but did not allow for ldaps:// URLs.


cstamas:
+1

peterlynch:
+1

PR #501, opened 2012-08-01
Drop support for RESTART command as t is JSW specific (will not be possible to implement it via installer launcher)

adreghiciu:


cstamas:
+1

PR #502, opened 2012-08-01
[NEXUS-5188] Copy keep alive threads from sisu-jsw-utils

adreghiciu:


nexus/nexus-bootstrap/pom.xml
```diff
@@ -44,6 +44,18 @@
...
+      <artifactId>litmus-testsupport</artifactId>
+      <version>1.3</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit-dep</artifactId>
+      <version>4.10</version>
+      <scope>test</scope>
+    </dependency>
```
jdillon, line 58: The point of litmus-testsupport is so that we do not have to specify any other primary deps in a pom... like this one for junit-dep.  Its also means that all versions for those primary test deps are all in one place and not scattered through out all projects.

litmus-testsupport:1.3 should already include junit:junit-dep 4.10, why are you adding it again here?

nexus/nexus-bootstrap/src/main/java/org/sonatype/nexus/bootstrap/KeepAliveThread.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public KeepAliveThread( final int port )
+        throws IOException
+    {
+        this.socket = new ServerSocket( port, 1, InetAddress.getByName( "127.0.0.1" ) );
+        this.running = true;
+
+        setDaemon( true );
+        setName( "Keep Alive" );
+    }
```
jdillon, line 47: Did we chat about having the monitor handle this server-side end-point?  So that there is only one thread running which can server STOP + PING responses?

nexus/nexus-bootstrap/pom.xml
adreghiciu, line 58: Because Nexus via dm enforces to 4.8.x and that results in problems

nexus/nexus-bootstrap/src/main/java/org/sonatype/nexus/bootstrap/KeepAliveThread.java
adreghiciu, line 47: I do not recall but yes we can do that.

PR #505, opened 2012-08-01
[NEXUS-5089] Launcher improvements

adreghiciu:
Depends on sonatype/sisu-bundle-launcher#4 and sonatype/sisu-litmus#8


nexus/nexus-bootstrap/pom.xml
```diff
@@ -44,6 +44,18 @@
...
+      <artifactId>litmus-testsupport</artifactId>
+      <version>1.3</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit-dep</artifactId>
+      <version>4.10</version>
+      <scope>test</scope>
+    </dependency>
```
jdillon, line 58: junit-dep should be picked up transitively from litmus-testsupport.  Please do not duplicate these core test deps in projects, as it makes version management a pain.

adreghiciu, line 58: The problem is that Nexus main pom uses 4.8.x in DM and that results in failures.

jdillon, line 58: fffff, okay.  weren't we on a path to updating nexus to use updated junit @peterlynch ?

peterlynch, line 58: yes, was waiting for 2.1 to be released before potentially destabilizing a bunch of tests. I'll move this task to my short list.

adreghiciu:
Pull is now ready to be reviewed/merged

jdillon:
I'm still not fond of the LogProxy or the "nicer" log format.  so +0.75

adreghiciu:
I explained to you why we have the proxy. If is not there I cannot reuse those classes in Nexus versions bellow 2.1 and will be pity to not support testing against those versions or make copies of this tasks to be used in 2.0.x versions.

jdillon:
Ya I still don't like it though :-P

peterlynch:
The new log pattern was fine with me. Going back to %date by default is surely a mistake. My head didn't hurt looking at the new log output either ;). Configurable property is OK, but that is just icing and still leaves a less than ideal default log pattern.

I appreciate Alin trying to make it better.

Anyways none of this warrants holding up this merge any longer.

+1

nexus/nexus-bootstrap/src/main/java/org/sonatype/nexus/bootstrap/monitor/ShutdownIfNotAliveThread.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        finally
+        {
+            close( socket );
+        }
+    }
+
+    void shutdown()
+    {
+        log.info( "Shutting down as there is no ping response on port {}", port );
```
peterlynch: since this is an abnormal way to shutdown, shouldn't this be warn?

nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/NexusBundle.java
```diff
@@ -25,6 +25,8 @@
     extends WebBundle<NexusBundle, NexusBundleConfiguration>
 {
 
+    File getNexusDirectory();
+
     File getWorkDirectory();
```
peterlynch, line 43: These two methods need javadoc with examples of what you mean in the context of the real bundle.

nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/internal/NexusITLauncher.java
```diff
@@ -0,0 +1,143 @@
...
+import java.lang.reflect.Method;
+
+import org.sonatype.nexus.bootstrap.monitor.CommandMonitorThread;
+import org.sonatype.nexus.bootstrap.monitor.ShutdownIfNotAliveThread;
+import org.sonatype.nexus.bootstrap.monitor.commands.PingCommand;
+import org.sonatype.nexus.bootstrap.monitor.commands.StopApplicationCommand;
+import org.tanukisoftware.wrapper.WrapperManager;
+
+/**
+ * TODO
```
peterlynch: Explain when and why you would use this class

peterlynch:
+1 other than javadoc required in a few places.

adreghiciu:
@peterlynch All your concerns were addressed above

PR #510, opened 2012-08-06
NXCM-4560 junit410 upgrade

peterlynch:
Includes upgrades to hamcrest, litmus, junit, mockito, powermock.
Trying to keep friendly versions of the test libraries in sync.

Previous feature job failed due to unrelated IT hang

Trying again:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/384/


peterlynch:
more hung ITs - I am pretty sure that is the only problem I keep running into.

Pull should be fine. Can't pull NXCM until this one gets done. And won't bother with anymore feature builds until https://github.com/sonatype/nexus/pull/505 is in.

peterlynch:
@jdillon do you want me to revert 6e8d78909 ? I did notice at one point all three versions of litmus were being used.

nabcos:
+1

peterlynch:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/388/

peterlynch:
CI +1 https://builds.sonatype.org/job/nexus-oss-its-feature/388/

nexus-maven-plugins/pom.xml
```diff
@@ -155,32 +155,26 @@
...
         <artifactId>hamcrest-core</artifactId>
-        <version>1.3.RC2</version>
+        <version>1.3</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.sonatype.sisu.litmus</groupId>
         <artifactId>litmus-testsupport</artifactId>
-        <version>1.2</version>
+        <version>1.4-SNAPSHOT</version>
```
jdillon: One of the points of litmus is to consolidate the dependencies and version management for junit/hamcrest/etc.... so that projects simply have to depend on litmus to get the right version of everything.

peterlynch: well yes, and so was nexus-test-common before that - but in fact in some cases it seemed direct declaration of dependency was required in the past -  some cases a compile scope dependency was depending on junit version x and that was overriding nexus-test-common transitive dependency of same artifact at scope test.

so in this pull I went for the minimal changes I thought would be unlikely to change the dependency graph

I can take another look to see what removing all the explicit test dependencies does this time around

jdillon: I think if we are going to spend the effort to update the deps, we should spend the effort to minimize there declarations as well.

cstamas:
+1

peterlynch:
needs another grid build

peterlynch:
CI gives another +1 

https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/401/

peterlynch:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/402/

peterlynch:
looks stable, merging

PR #515, opened 2012-08-13
o [NEXUS-5202] On Logout, only change location hash to '#welcome'

nabcos:
The whole page was reloaded before (with window.location = ...),
triggering many more network requests than needed.


peterlynch:
makes me a little nervous, needs much testing once merged

peterlynch:
+1

nabcos:
we can also do the reload and keep the query bit intact, but it's much cleaner this way, IMO.

adreghiciu:
+1

cstamas:
+1

PR #517, opened 2012-08-13
AbstractMavenRepositoryTemplate should use AbstractRepositoryTemplateProvider

heroldus:
The constructor argument of AbstractMavenRepositoryTemplate should use AbstractRepositoryTemplateProvider like the super constructor does. This makes it much easier to register a new template because your TemplateProvider does not need to inherit from DefaultRepositoryTemplateProvider and override its methods and annotations.

Regards,
Sebastian Herold


cstamas:
:beer:

adreghiciu:
+1

PR #525, opened 2012-08-28
o [NEXUS-5228] Ensure bug compatibility for suppressStatus option

nabcos:
Some plugins use strings to specify the status code. Previous change in
this area got rid of implicit type conversion on equality check and
"broke" that behavior.


peterlynch:
+1

peterlynch:
no objections and we need to verify this is 2.1.3 build. merging...

PR #521, opened 2012-08-22
o [NEXUS-5215] Let httpclient not follow redirects to index pages

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/397/

PR #526, opened 2012-08-29
NEXUS-5215 remote index page

peterlynch:
replaced pull request on wrongly named branch sonatype/nexus#521


peterlynch:
merging because 521 was open for 6 days and no one blinked, plus this will be verified again in 2.1.3 build

PR #529, opened 2012-09-03
Automatically include nexus-it-helper-plugin

adreghiciu:
- By default all NexusITSupport based ITs will include the nexus-it-helper-plugin.
- Create a new module for nexus clients for REST resources available in above plugin.
- Automatically log start/end of tests. This info will appear in nexus.log and will make easier to track what happened in nexus dring test


nabcos:
+1

cstamas:
:kiss:

PR #530, opened 2012-09-04
NEXUS-5118: Rationalizing logging in bundle

cstamas:
Logging between Nexus 1.9, 2.0, 2.1 changed a lot, and 2.2 should sort all the user reported problems, like:
- Jetty and Nexus uses different logging frameworks (Jetty uses ConsoleLogger while Nexus logback in 2.0 for example)
- duplications of logback libraries in 2.1
  etc

CI:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/407/


nexus/nexus-logging-extras-appender/src/main/java/org/sonatype/nexus/logback/EventTarget.java
```diff
@@ -0,0 +1,26 @@
...
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+
+/**
+ * An target that receives logging events from {@link ForwardingAppender}.
+ * 
+ * @author cstamas
+ * @since 2.2
+ */
+public interface EventTarget
```
peterlynch, line 23: If this is all about logging, is there a reason you did not call this LoggingEventTarget?

peterlynch:
checked out the branch and tried this out a bit - seems to log things correctly - and nexus war is not affected ( still contain logback impl ), so appears all fine.

If there are problems this is solving, are there tests we can write to demonstrate the problems and prove this changes fixes it?

peterlynch:
https://issues.sonatype.org/browse/NEXUS-5118?focusedCommentId=167404&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-167404

cstamas:
CI passed on both slaves, but new file missed a license header (fixed in c0ee3dd7742af44c5f56a2e6748b3e0b698362f2)

adreghiciu:
+1

nabcos:
+1

PR #531, opened 2012-09-05
NEXUS-5241: Simple fix

cstamas:
A simple fix fixing the issue, easily backportable to 2.1 line too.


adreghiciu:
+1

nabcos:
If we have a whitelist of things to include, do we need a way for plugins to contribute to that whitelist? The exclusions are for maven-specific things (which is still in nexus-core, by chance), what about p2/obr/nuget/yum/ruby?

EDIT: actually a blacklist might also need plugin contributions

cstamas:
Exactly that is the "proper" fix. This fix is just a shortcut that is
easily backportable to 2.1.x

Thanks,
~t~ mobile

nabcos:
OK, I did not get that. Should we merge this only to 2.1.x then, and leave the issue open for 2.2?

PR #533, opened 2012-09-05
Implement a nexus client for server configuration

adreghiciu:
For now only http proxy supported. Rest of configuration will be added as needed.

### DIFF
--- a/nexus/nexus-client-core/pom.xml
+++ b/nexus/nexus-client-core/pom.xml
@@ -109,6 +109,12 @@
       <scope>compile</scope>
     </dependency>
 
+    <dependency>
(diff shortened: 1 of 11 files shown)


cstamas:
+1 but how about authc support too (fine if not yet needed), as username/pw for proxy...

nabcos:
+1

adreghiciu:
@cstamas review

cstamas:
:beer:

nabcos:
is there  a UT with an optional segment? I find that structure with optional segments etc. a little confusing, but in general +1

adreghiciu:
JerseyHttpProxyTest is also testing that as part of testing HttpProxy

PR #534, opened 2012-09-06
o [NXCM-1508] Rewrite IT to be more reliable

nabcos:
The IT was running three tasks, then checking once whether one was sleeping and
then just waited for all tasks to be in the default state.

Now it's running three tasks, waits until one is sleeping (failing only
after 10s), then waits for completion of all tasks and asserts that they
actually ran after they were created.

It is still possible for this IT to fail without cause if the tasks are finishing too fast, so that no task will end up sleeping (because it can just start immediately).


nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/411/

PR #535, opened 2012-09-06
o [NEXUS-4609] Add additional wait for tasks *and* events after the index update step

nabcos:
IT was not waiting for processing of indexer events.


adreghiciu:
+1

PR #541, opened 2012-09-14
Nxcm 4643 launcher shutdown hangs

jdillon:
Looking for feedback on some changes which could help reduce hanging NX instances.


nabcos:
Is the ExitCommand ever used with these changes?
Other changes make sense, +1

peterlynch:
I think the logging was there before because there were cases where we thought we had a nexus running with shutdown monitoring, but turned out we didn't. But if all that logging causes blocking shutdown, then it doesn't much matter.

probably should just merge this and see what happens downstream...

jdillon:
ExitCommand is not used directly, I added it in case someone wanted to manually send an EXIT to gracefully terminate.

nabcos:
I tried these changes with some builds (smart proxy, staging (actually built the tag for 2.1, but that should be valid as well), outreach, trial-plugin), all succeeded so this pull does not make it worse ;)

I agree with Peter, we should just merge it.

peterlynch:
26 hung processes is enough to justify trying this out instead. Merging...

PR #544, opened 2012-09-17
NEXUS-5249: Do not propagate exception in case that downloading the remote index failed

cstamas:
This pull fixes the original issue, but also:
- still retains required behaviour from NEXUS-4275: If a remote index fails to download due to 401/403/50x problems we should end the "download index" task with a "broken" state
- implements NEXUS-3424 When remote index could not be downloaded, log stack trace only when debug is enabled

CI scheduled (prev job still running):
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/416/

### DIFF
--- a/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
+++ b/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
@@ -109,6 +109,7 @@ import org.sonatype.nexus.proxy.repository.Repository;
 import org.sonatype.nexus.proxy.repository.ShadowRepository;
 import org.sonatype.nexus.proxy.storage.local.fs.DefaultFSLocalRepositoryStorage;
 import org.sonatype.nexus.proxy.utils.RepositoryStringUtils;
+import org.sonatype.nexus.util.CompositeException;
 import org.sonatype.nexus.util.SystemPropertiesHelper;
 import org.sonatype.scheduling.TaskInterruptedException;
 import org.sonatype.scheduling.TaskUtil;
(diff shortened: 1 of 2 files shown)


cstamas:
CI +1'd
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/416/

nabcos:
+1

ifedorenko:
-1

I am probably missing something, but there are still cases when problems with one repository index will stop index download task. Either downloadRepositoryIndex should not throw any exceptions or downloadAllIndex should implement catch-log-continue logic.

cstamas:
Yes, as required by NEXUS-4275 (mentioned in original description), there _must be cases_ like 401/403/50x problems that needs to stop index download task (and make it "broken state"). This change _retains_ NEXUS-4275 behaviour, introduced in 2.1.

If downloadRepoIndex does not throw anything, task would never become "broken". Same for downloadAllIndex.

ifedorenko:
Still -1. 

"Broken" task end state does not mean the task cannot attempt index download for all repositories first. Just collect all exceptions in downloadAllIndex, then throw a compound exception.

cstamas:
Might be, but this is not what I read from NEXUS-4275 description: "If a remote index fails to download due to 401/403/50x problems we should end the "download index" task with a "broken" state."

I am fine with what you propose, so issue should be updated accordingly (or linked to this NEXUS-5249) with remark it _changes_ existing 2.1 feature.

ifedorenko:
I don't see a contradiction here -- NEXUS-4275 is about single repository, NEXUS-5249 is about all repositories. Also, in both cases the task will "end" in "broken" state. 

Stopping index download for one repository when another has issues does not make sense.

cstamas:
That's the point, neither do i see contradiction (between what was implemented in 2.1 and the issue). In short, maybe the spec was too "short", or not explicit enough, not stating "stop immediately" or "stop after processing all repositories" ;)

NEXUS-4275 and NEXUS-5249 is both about UpdateIndexTask/RepairIndexTask (there is no "single repo" or "all repo" task, it's same task differently configured), so we still talk about same part of code

ifedorenko:
Please change the code to (attempt) download all indexes and report any errors at the end. Failure in one repo index download must not prevent index download for other repository.

cstamas:
Closing this pull for now, a new one will be created.

PR #545, opened 2012-09-18
NEXUS-5216: Using released nexus-archetype-commons.

cstamas:
Just parked on branch, as upon release there is some time
needed to make the artifact appear in Central, until then
the build would become broken?

So, this trivial pull is only here, that when 1.2 appears here:
http://repo1.maven.org/maven2/org/sonatype/spice/nexus-archetype-common/

Then just press merge.


ifedorenko:
I was under impression that nexus CI used rso, and should be able to resolve nexus-archetype-commons immediately after it was deployed, no?

cstamas:
Frankly, am unsure about this: how is something deployed to "releases" repository, that is rsync-ed to Central, and  it's proxy central-proxy is member of public-grid group getting into CI.

But it does not matter, as I see, 1.2 landed in Central.

PR #546, opened 2012-09-18
Add ShutdownHelper so we can delegate std-java or jsw implementations to...

jdillon:
... exit/halt

https://issues.sonatype.org/browse/NXCM-4643


ifedorenko:
+1

peterlynch:
Any reason not to call WrapperManager.hasShutdownHookBeenTriggered() somewhere?

jdillon:
@peterlynch Where would you put it?

peterlynch:
+1

I looked into WrapperManager code, it looks like calling stop will do the right thing - and we no longer call System.exit() anymore with these changes.

peterlynch:
this change has support and we desire grid stability ;) merging..

PR #547, opened 2012-09-20
NEXUS-5249: Update Indexes task stops processing on RemoteItemNotFoundException

cstamas:
This pull fixes the original issue, but also:
- still retains required behaviour from NEXUS-4275: If a remote index fails to download due to 401/403/50x problems we should end the "download index" task with a "broken" state
- implements NEXUS-3424 When remote index could not be downloaded, log stack trace only when debug is enabled

### DIFF
--- a/nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/CompositeException.java
+++ b/nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/CompositeException.java
@@ -1,0 +1,114 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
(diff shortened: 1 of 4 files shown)


cstamas:
CI job scheduled:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/419/

ifedorenko:
+1

cstamas:
CI
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/422/

cstamas:
CI voted +1

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/CompositeException.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * A composite {@link Exception} descendant, that is able to collect multiple causes to have them throw at the end of
+ * some batch processing for example. Inspired by code from <a href=
+ * "http://stackoverflow.com/questions/12481583/exception-composition-in-java-when-both-first-strategy-and-recovery-strategy-fai"
+ * >Stack Overflow</a>.
+ * 
+ * @author cstamas
+ * @since 2.2
+ */
+public class CompositeException
```
peterlynch, line 35: Unit test for this?

nabcos:
+1

nexus/nexus-utils/src/main/java/org/sonatype/nexus/util/CompositeException.java
cstamas, line 35: Added UT

ifedorenko, line 35: Why do we need a unit test to cover CompositeException? As far as I understand, there are two important aspects of behaviour around CompositeException -- the download all indexes operation must fail if any of the indexes failed to download and the original IOException is logged to allow troubleshooting. As far as I can tell, neither of these two depend on CompositeException.

PR #548, opened 2012-09-21
o [NEXUS-5118] Set org.eclipse.jetty logging to INFO

nabcos:
With the changes for the bundle logging, Nexus logback controls jetty logging as well. So if you set the logback.properties to DEBUG, you would also get lots of jetty logging...

Please sanity check this. Is this logback-nexus.xml used throughout the bundles or are there other places that need changes?

### DIFF
--- a/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
+++ b/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
@@ -38,6 +38,7 @@
   <logger name="org.restlet" level="WARN" />
   <logger name="org.apache.commons" level="WARN" />
   <logger name="org.apache.shiro.web.filter.authc.BasicHttpAuthenticationFilter" level="INFO" />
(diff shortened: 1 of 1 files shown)


cstamas:
+1, looks fine.

`/nexus-oss-webapp-2.2/conf/logback.xml` is used by Jetty during booting (is picked up automatically by logback, as this `conf` directory is on classpath). 

When Nexus (a webapp within Jetty) boots, it "takes over" the control, and directly configures logback from `/sonatype-work/nexus/conf/logback.xml` using Logback APIs.

Hence, the change looks sane, but we will have to document how to debug Jetty, if needed (undo this line or set to DEBUG), as UI will not help in this case.

peterlynch:
+1
we need a how to configure Nexus logging faq in the knowledge base

adreghiciu:
+1

nabcos:
Created issue:
NEXUS-5272: Add Knowledge Base entry for logging config

PR #549, opened 2012-09-21
o [NEXUS-5227] Set 'jettyPlexusCompatibility=true' by default

nabcos:
Properties were removed with the comment that all those are set in default.properties from nexus-bootstrap, but jettyPlexusCompatibility is not.


nabcos:
Closed because as this setting should go into default.properties ([if at all](https://github.com/sonatype/nexus-enterprise/pull/259#issuecomment-8805983)).

PR #550, opened 2012-09-24
NEXUS-5192: Implementing the property to control "local avail check"

cstamas:
Making it default to false (reverse behaviour as so far), but still
making it possible to be enabled if needed.

This change reverts how NX behaved so far.

CI:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/423/


cstamas:
CI +1'd
(but both slave jobs failed as CompositeExceptionTest from master was missing license header, fixed in b4bfb187ac48e0a2fb0e3856c3e19f7a0f5764bb)

ifedorenko:
+1 

code format looks proper in the diff

nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/treeview/DefaultMergedTreeNodeFactory.java
```diff
@@ -159,18 +118,65 @@ protected ResourceStoreRequest getResourceStoreRequest( String path )
...
+                }
+            }
+        }
+        catch ( ItemNotFoundException e )
+        {
+            // mute it, probaly not available locally
+        }
+        catch ( Exception e )
+        {
+            // TODO: do something here?
```
nabcos: at least some logging?

adreghiciu:
+1

nabcos:
+1 but maybe handle the TODO? It's only for UI though, so ignoring it may be fine, but then the comment should say so.

PR #551, opened 2012-09-25
Use the path of remote url to check if the remote storage request is made for a collection

adreghiciu:
Our check was using the full url that could include the query part as in for example "http://foo.com/bar?x=y/", case when we get a false positive

Also added UTs for checking. The problem is only on http client 4, the old one behaves as expected


ifedorenko:
Is this required for any 2.2 work?

adreghiciu:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/424/

adreghiciu:
@ifedorenko In my view yes as in 2.2 we ship with http 4 as default client and this is a regression.

E.g. retrieving the mirrors for  a p2 repository will fail for a mirror URL like 
http://www.eclipse.org/downloads/download.php?format=xml&file=/releases/juno/201206270900/

(see the slash by the end)

nabcos:
+1

cstamas:
:kiss:

adreghiciu:
CI +1 - build failed with an unrelated failure (ux only)

PR #552, opened 2012-09-27
Simple change to make ITs work

cstamas:
Minimal change in AppContext story needed for ITs to work.

As Plexus tries to copy the passed in map (instead of delegating to it as SISU does),
now passing in the flattened AppContext as context.

Also, implemented nexus and nexus-test property file
loading, since they are optional (last ctor param is false, if not
found it will be silently ignored).


cstamas:
CI job started:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/425/

adreghiciu:
+1 Tested via central secure ITs

adreghiciu:
CI voted +1

ifedorenko:
+1

PR #553, opened 2012-09-27
NEXUS-5280: ContextMemberProvider should always return same list

cstamas:
... of members.

At least during search processing is running. As it is now, it is directly
hooked onto group, that if group changed during search, causes to have
different repositories locked or unlocked. In case of member addition,
an IllegalMonitorStateEx would happen (a non-locked context would
be unlocked) and in case of member removal, the context of
removed member would remain locked.

### DIFF
--- a/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
+++ b/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
@@ -21,6 +21,7 @@ import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
```diff
@@ -453,19 +441,23 @@ public void updateRepositoryIndexContext( String repositoryId )
...
+            // is a group OR repo path changed OR we have an isIndexed transition happening
             if ( repository.isIndexable() )
             {
                 if ( ctx == null
-                    || ( !ctx.getRepository().getAbsolutePath().equals( repoRoot.getAbsolutePath() ) || ctx.isSearchable() != repository.isSearchable() ) )
+                    || ( repository.getRepositoryKind().isFacetAvailable( GroupRepository.class )
+                        || !ctx.getRepository().getAbsolutePath().equals( repoRoot.getAbsolutePath() ) || ctx.isSearchable() != repository.isSearchable() ) )
                 {
                     // recreate the context
                     addRepositoryIndexContext( repositoryId );
```
ifedorenko: Not a big deal, but I think this snippet has less code duplication and is a little easier to follow

```
        // remove context, if it already existed (ctx != null) and any of the following is true:
        // is a group OR repo path changed OR we have an isIndexed transition happening
        if ( ctx != null
            && ( repository.getRepositoryKind().isFacetAvailable( GroupRepository.class )
                || !ctx.getRepository().getAbsolutePath().equals( repoRoot.getAbsolutePath() ) || ctx.isSearchable() != repository.isSearchable() ) )
        {
            // remove the context
            removeRepositoryIndexContext( repositoryId, false );
            ctx = null;
        }

        // (re)create the context
        if ( repository.isIndexable() && ctx == null )
        {
            addRepositoryIndexContext( repositoryId );
        }
```

ifedorenko:
I think some UT coverage is missing here. For example, all UTs still pass even if I change the code to remove but not add group repo context during updateRepositoryIndexContext.

nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
cstamas: Applied in subsequent commit.

nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
```diff
@@ -453,24 +441,23 @@ public void updateRepositoryIndexContext( String repositoryId )
...
-                    || ( !ctx.getRepository().getAbsolutePath().equals( repoRoot.getAbsolutePath() ) || ctx.isSearchable() != repository.isSearchable() ) )
-                {
-                    // recreate the context
-                    addRepositoryIndexContext( repositoryId );
-                }
-
+                // recreate the context
+                addRepositoryIndexContext( repositoryId );
                 // set include in search/indexable
                 setRepositoryIndexContextSearchable( repositoryId, repository.isSearchable() );
```
ifedorenko: setRepositoryIndexContextSearchable call seems redundant. Why is it needed here?

ifedorenko:
+1

merge it!

adreghiciu:
+1

PR #427, opened 2012-06-06
[NEXUS-3427] Manually save null values for user agent customization and additional query params

adreghiciu:
Signed-off-by: Alin Dreghiciu adreghiciu@gmail.com

PR #554, opened 2012-10-01
NEXUS-5280: ContextMemberProvider should always return same list of members

cstamas:
But, we forgot to handle a special case: there is one case, when instead
of the group member change, a member configuration change might cause that
group context list MUST change, the groups
ContextMemberProvider pre-built list must change: that when
Repository#setIndexable() is changed (if change from true to false, context
is removed from NexusIndexer, if changed from false to true, context is being
created). This case was not handled, and groups (who's member was made
non-indexable) was still containing and handing over a "stale", already
shut down context to indexer components.

In other words, with original change, the list remain unchanged, even if the context was
shut down and removed from NexusIndexer, but group continued to use it, that
caused NPE (IndexSearcher is closed and nullified on shut down contexts),
but also some cleanup problem exists in MI and NPE caused IllegalMonitorState
(NPE happened during locking, so not all context were actually locked,
but against all were unlock tried).

Now, repository updates are propagated to groups where it is contained.

Added UT that tests exactly this case: making a repo from indexable
to non-indexable and back to indexable, while verifying that it's group
is not running into those sort of problems.

CI:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/426/

### DIFF
--- a/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/events/IndexerNexusStoppedEventInspector.java
+++ b/nexus/nexus-core-plugins/nexus-indexer-lucene-plugin-parent/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/events/IndexerNexusStoppedEventInspector.java
@@ -31,7 +31,6 @@ import org.sonatype.plexus.appevents.Event;
 public class IndexerNexusStoppedEventInspector
     extends AbstractEventInspector
 {
-
     @Requirement
     private IndexerManager indexerManager;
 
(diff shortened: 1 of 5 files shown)


nabcos:
+1

cstamas:
There's more to it. Nexus with NEXUS-5280 implemented lost it's capability to load configuration where a group g1 references as member g2, that is not YET loaded (a valid scenario). This caused that ITs in job above, but also in procurement suite to fail. Fix on way.

cstamas:
New CI:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/427/

cstamas:
CI #427 voted +1

peterlynch:
Does any of this fix the subtasks of NXCM-4720?

I like the fact we have a test now, and it is well commented as to what it does during the test but lets be clear is not a 'unit' test. It tests that a bunch of independent units are all playing nice together - it has value to prove something works, but I'd prefer to see unit test methods at a more granular level, testing changes to those specific public methods  with mocks, etc. I expect the test was written this way for legacy reasons and to bolt it into existing legacy test framework.

+1

PR #555, opened 2012-10-02
Let UI logging also use the LogHelper resource

nabcos:
This change lets UI log to the nexus.log file when the it-helper plugin is installed.

Useful IMO because you can get real debug logging (if used in the UI, ATM it's only usertoken that uses this), but it potentially adds many GET requests to the loghelper resource when clicking through the UI.

### DIFF
--- a/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/resources/static/loghelper/js/Nexus.Log.js
+++ b/nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/resources/static/loghelper/js/Nexus.Log.js
@@ -1,0 +1,43 @@
+/*
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+
+/*global Ext, Sonatype, Nexus*/
+
+(function() {
+  var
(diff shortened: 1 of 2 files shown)


peterlynch:
seems just related to improving IT logging, so +1 for that idea - looks fragile though if the plugin group/artifact id changes since getVersion... method swallows exceptions and returns null if version is not found. Not clear if we can make this fail fast,  otherwise existing method seems like it will fail in the background without anyone noticing.

nabcos:
It's actually not used for ITs, only when you manually go and actually use the UI. Next steps could be to move loghelper and this js to its own plugin (or to pluginkit), and enable the UI remote logging with a capability. Then we could actually deliver this without performance impact, and turn it on if need be.

(Should probably not go to pluginkit though - or do we want to install pluginkit for IT runs..?)

(BTW, I'm pretty sure nearly every plugin uses the getVersion code snippet. pluginkit may have a better method to do that, but that's not usable for oss plugins.)

peterlynch:
is this needed for 2.2 for something?

nabcos:
No, I had this in my mind for quite a while and started it because I thought I would need to debug usertoken UI.

nexus/nexus-test-harness/nexus-it-helper-plugin/src/main/resources/static/loghelper/js/Nexus.Log.js
```diff
@@ -0,0 +1,43 @@
...
+
+            fn.apply(this, array);
+          };
+        };
+
+  Nexus.Log.debug = Nexus.Log.debug.createSequence(curry(log, 'DEBUG'));
+  Nexus.Log.info = Nexus.Log.info.createSequence(curry(log, "INFO"));
+  Nexus.Log.warn = Nexus.Log.warn.createSequence(curry(log, "WARN"));
+  Nexus.Log.error = Nexus.Log.error.createSequence(curry(log, "ERROR"));
+  Nexus.log = Nexus.log.createSequence(curry(log, "DEBUG"));
```
peterlynch, line 41: why the mixed case on the word 'log'

nabcos, line 41: callable function in contrast to a "class" holding functions

peterlynch:
so this is for manually a testing via UI a nexus setup by IT?

nabcos:
If the it-helper plugin is installed, UI logging (usually only in the js browser console, and you have to set a variable to get debug logging) will also send all log messages to the nexus.log.

Useful for manually debugging UI, although there's not much that uses even Nexus.log ATM.

cstamas:
Is this pull stale? Or why is it not merged?

PR #556, opened 2012-10-03
NEXUS-5288: Appender class changed.

cstamas:
The appender did change between 2.1 and 2.2. Problem with
"upgrade step" is that log manager is started from a servlet
context listener, way before nexus is, hence, nexus upgrade
step would be able to do anything way after logback configuration
is loaded (and errors would be emitted about Appender).

After chatting with Alin, this looked the simplest: introduce
a marker interface, that would be implemented by those
implementations of LogConfigurationParticipant, that does NOT
want to let users tamper with their configuration. Event system
config participant is such LogConfigurationParticipant (see it's
config file, states "DO NOT EDIT").

When LogbackLogManager detects that a participant implements this
marker interface, it will not check for config file existence
(as it happened before this change) and write it out when not present,
but it will "blindly", always write it out, potentially replacing
the changed file.

Now, the LogbackNexusEventSystemLogConfigurationParticipant implements
the NonEditable marker interface and this resolves the problem,
as on upgrade (from X to 2.2), 2.2 will simply overwrite the
logback-event.xml.

### DIFF
--- a/nexus/nexus-logging-extras/src/main/java/org/sonatype/nexus/log/LogConfigurationParticipant.java
+++ b/nexus/nexus-logging-extras/src/main/java/org/sonatype/nexus/log/LogConfigurationParticipant.java
@@ -14,11 +14,28 @@ package org.sonatype.nexus.log;
 
 import java.io.InputStream;
 
+/**
+ * A component contract that wants to provide extra logging configuration, participate in configuration of logging in
+ * Nexus.
+ * 
+ * @author adreghiciu
+ */
 public interface LogConfigurationParticipant
 {
-    
     String getName();
 
     InputStream getConfiguration();
 
+    /**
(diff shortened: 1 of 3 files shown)


cstamas:
CI job:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/428/

cstamas:
CI voted +1, but IT that is failing in master too (or was, when this branch was created) nxcm4591 failed here too, but is completely unrelated.

adreghiciu:
+1

cstamas:
Tested manually, on boot the file `logback-events.xml` is overwritten on every boot, as it should be (to fix this issue, but also to prevent manual tampering with it)

peterlynch:
+1

PR #557, opened 2012-10-03
Log timing of http remote requests made by Nexus

adreghiciu:


adreghiciu:
@peterlynch first we get this in and then we can document it

peterlynch:
@adreghiciu sure was more of a reminder

nabcos:
+1

adreghiciu:
+1 on @cstamas changes

peterlynch:
+1 its good

jdillon:
I think this would have been better implemented with something like javasimon, which adds more detail and has the ability to be exposed via jmx.

Its used by an aspect in smartproxy for reference:

https://github.com/sonatype/nexus-smartproxy-plugin/blob/master/nexus-smartproxy-plugin/src/main/java/com/sonatype/nexus/plugins/smartproxy/internal/TimingAspect.java

PR #558, opened 2012-10-03
[NXCM-4737] hack in query-string customizations via QueryStringBuilder s...

jdillon:
...o that remote browsing secure.central gets the proper request paramemters.

https://issues.sonatype.org/browse/NXCM-4737

### DIFF
--- a/nexus/nexus-core-plugins/nexus-rrb-plugin/src/main/java/org/sonatype/nexus/plugins/rrb/QueryStrings.java
+++ b/nexus/nexus-core-plugins/nexus-rrb-plugin/src/main/java/org/sonatype/nexus/plugins/rrb/QueryStrings.java
@@ -1,16 +1,15 @@
-/*
- * Copyright (c) 2007-2012 Sonatype, Inc. All rights reserved.
+/**
+ * Sonatype Nexus (TM) Open Source Version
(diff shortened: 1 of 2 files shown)


jdillon:
As noted in FIXME comment, this support should probably be done automatically in AhcProvider.  Though as I was not even aware AHC was in use anymore, to avoid potentially breaking something else I've limited this change only to rrb.

jdillon:
https://builds.sonatype.org/job/nexus-oss-its-feature/430/ FAILED in license headers (otherwise all other bits passed)

peterlynch:
wondering about a test..not sure since ahc goes away soon?

otherwise +1

jdillon:
There was not already a generic test for the class this lived in, and I didn't see too much value in making one just to add this additional test.  We already have tests for the builder used here, and I don't think we need to test if AHC works, and as you noted it should go away anyways.

nabcos:
+1, not pretty but as long as we're waiting for https://issues.sonatype.org/browse/NEXUS-4689 ...

jdillon:
BTW, if the changes to HttpClientUtil to configure() release() are made public, then we could really re-implement this using httpclient4 w/o much problem.

PR #560, opened 2012-10-04
HttpClient 4x being slow with HTTPS connections.

cstamas:
Changes in here:
- bumped version to httpclient 4.2.1 + httpcore 4.2.2 (latest GA)
- applied recommended changes to connManager, parameters handling
- catching IllegalStateException thrown by httpClient connmgr in case when JVM lacks support for TLS (was swallowed without this!)
- removing explicit connection close, to make use of pooling
- hostname verification for HTTPS connections (new in HC4x, not done by HC3x!) left in place, but due to pooling it is made only once, when connection created and not after it (when connection is reused).

These changes makes the originally 2.5x times slower HC4 (vs HC3) against HTTPS servers about 3.0x times faster!

Related issue: https://issues.sonatype.org/browse/NEXUS-5291


cstamas:
CI https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/431/

peterlynch:
+1

cstamas:
CI linux passed, win had one unrelated (but search related) IT failure. I count this as +1

adreghiciu:
+1

PR #561, opened 2012-10-05
[NEXUS-5294] Display warning and disable save for users from external realms

nabcos:
External users had all fields disabled but still could save their data,
resulting in a message box stating that LDAP is a read-only realm.

This pull adds a warning message above all fields that external users are read-only, and disables 'Save' functionality.

I scheduled [the issue](https://issues.sonatype.org/browse/NEXUS-5294) as 2.2/next because it's an easy "fix" but does not need to go into 2.2.

### DIFF
--- a/nexus/nexus-webapp/src/main/webapp/js/nexus/profile/Summary.js
+++ b/nexus/nexus-webapp/src/main/webapp/js/nexus/profile/Summary.js
@@ -30,6 +30,15 @@ Nexus.profile.Summary = function(config) {
 
   items = [
     {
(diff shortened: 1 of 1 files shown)


peterlynch:
Strange. I click the issue link and it takes me to NEXUS-5295 even though it's linked to NEXUS-5294.

However, can we get a screenshot of the change to the issue?

Also this may be an improvement in UI, but I am not sure how it helps user change their email address. :)

cstamas:
+1

PR #562, opened 2012-10-05
[NEXUS-5296] Provide additional info about invocations in case of failing assertions

adreghiciu:
This commit does not actually fix the failing IT but it will give more details about invocations of org.apache.maven.index.updater.IndexUpdater#fetchAndUpdateIndex as arguments and stack trace to be able to figure it out why are there more invocations then expected.

My guess is that the other invocations are actually due to changes done to proxy repositories during test preparations which will trigger remote index download. And sometimes those are not finished by the time we set the special index manager.


nabcos:
+1

peterlynch:
+1

PR #563, opened 2012-10-05
NEXUS-5292: Implemented connection eviction

cstamas:
- Introduced a new daemon thread that monitor and performs pool eviction on regular time intervals (5s). This is the proposed pattern by ASF HttpClient4x, to implement an external monitor thread, as HC4 does not have any other "internal" eviction mechanism. It can "check stale connection" per processed request, that introduces 30-40ms overhead per HTTP request, and still, if no outbound request happens in nexus, no eviction will happen.
- HttpClientUtil class improved, extracted some reusable non-proxy specific methods, applied some recommended changes and fixed problems: proxy authorisation was overriding target authorisation, also, a warning message was present (as the code originates from HC3 support class) that NTLM is unsupported for Proxies when other scheme is used for target. This is untrue for HC4x.
- the three introduced parameters are exposed to users (pool size, pool timeout and pool keep-alive), default values are poolSize=20, poolKeepAlive=1minute and poolTimeout=poolKeepAlive.  Users might override defaults by setting system properties.

Based on comments made by JD, some generic (non proxy repository specific) methods for HttpClient creation and releasing were made public. This means, that this class
might be used as "generic" HTTPClient4x factory, to create instances preconfigured and obeying Nexus config (like use of globally configured proxy etc).

Still, historically this class was made (intended to be?) a simple static utility class, but today it is growing. Probably the best would be to factor out the "hc factory" parts into a component, pulled out from this package (as it would be nor remote storage nor proxy specific) and this static class removed completely.

Existing class UT extended with eviction test and configuration discovery tests.

### DIFF
--- a/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientUtil.java
+++ b/nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientUtil.java
@@ -91,11 +91,6 @@ public class HttpClientUtil
      */
     private static final String CTX_KEY_S3_FLAG = ".remoteIsAmazonS3";
 
-    /**
-     * Context key of a flag present in case that NTLM authentication is configured.
-     */
-    private static final String CTX_KEY_NTLM_IS_IN_USE = ".ntlmIsInUse";
-
     // ==
     // HTTPClient4x pool and connection eviction related settings
 
@@ -201,12 +196,6 @@ public class HttpClientUtil
 
         // put client into context
         ctx.putContextObject( ctxPrefix + CTX_KEY_CLIENT, httpClient );
-        // NTLM flag: if for any auth NTLM is used, it has to be set
-        if ( ( ctx.getRemoteAuthenticationSettings() != null && ctx.getRemoteAuthenticationSettings() instanceof NtlmRemoteAuthenticationSettings )
-            || ( ctx.getRemoteProxySettings() != null && ctx.getRemoteProxySettings().getProxyAuthentication() != null && ctx.getRemoteProxySettings().getProxyAuthentication() instanceof NtlmRemoteAuthenticationSettings ) )
-        {
-            ctx.putContextObject( ctxPrefix + CTX_KEY_NTLM_IS_IN_USE, Boolean.TRUE );
-        }
         // NEXUS-3338: we don't know after config change is remote S3 (url changed maybe)
         ctx.putContextObject( ctxPrefix + CTX_KEY_S3_FLAG, new BooleanFlagHolder() );
 
@@ -270,7 +259,6 @@ public class HttpClientUtil
             ctx.removeContextObject( ctxPrefix + CTX_KEY_CLIENT );
         }
         ctx.removeContextObject( ctxPrefix + CTX_KEY_S3_FLAG );
-        ctx.putContextObject( ctxPrefix + CTX_KEY_NTLM_IS_IN_USE, Boolean.FALSE );
     }
 
     /**
@@ -301,19 +289,6 @@ public class HttpClientUtil
     static HttpClient getHttpClient( final String ctxPrefix, final RemoteStorageContext ctx )
     {
         return (HttpClient) ctx.getContextObject( ctxPrefix + CTX_KEY_CLIENT );
-    }
-
-    /**
-     * Whether or not the NTLM authentication is used.
-     * 
-     * @param ctxPrefix context keys prefix
-     * @param ctx remote repository context
-     * @return {@code true} if NTLM authentication is used, {@code false} otherwise
-     */
-    static Boolean isNTLMAuthenticationUsed( final String ctxPrefix, final RemoteStorageContext ctx )
-    {
-        final Object ntlmInUse = ctx.getContextObject( ctxPrefix + CTX_KEY_NTLM_IS_IN_USE );
-        return ntlmInUse != null && Boolean.parseBoolean( ntlmInUse.toString() );
     }
 
     /**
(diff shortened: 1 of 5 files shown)


cstamas:
CI job:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/432/

cstamas:
CI voted +1

adreghiciu:
+1

ifedorenko:
-1

implementation leaks HttpClient instances when repository is removed

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientUtil.java
```diff
@@ -79,21 +91,61 @@
...
+
+    /**
+     * Key for customizing connection pool timeout. In other words, how long should a HTTP request execution be blocked
+     * when pool is depleted, for a connection. Value is milliseconds.
      */
-    public static final String CONNECTION_POOL_SIZE_KEY = "httpClient.connectionPoolSize";
+    public static final String CONNECTION_POOL_TIMEOUT_SUFFIX = "connectionPoolTimeout";
 
     /**
-     * Marker used to determine that {@link #CONNECTION_POOL_SIZE_KEY} system property is not set.
```
ifedorenko, line 94: use CONNECTION_POOL_TIMEOUT_DEFAULT = CONNECTION_POOL_KEEPALIVE_DEFAULT, if the indent is to make pool and keepalive timeouts equal,

ifedorenko:
I think single pool shared among all repositories is easier to implement. This will eliminate the need to track and release HttpClient instances and hopefully make implementation simpler.

We need to expose max number of connections in the pool and limit number of per-repository connections. And we will still need connection eviction thread. But this is all standard httpclient 4 functionality, so I assume it is used and tested already.

jdillon:
I'm confused, you added a public release() method in HttpClientUtil, but no public method to configure one.

BTW, the use-case presently in RRB actually has a proxy repository so the original methods are the desirable to be used to get the httpclient for the proxy repository so it can use it to request directory listings.

cstamas:
This is work in progress, not done yet. The `public` access is wrong on release method. Actually, the release method is not needed anymore, as no connMgr.shutdown happens there anymore. And if you'd hook in release method, you'd leak MBeans in same way as original solution was leaking clients. At least that's what I saw on your branch (that you unregister MBeans in release method, something not called when repository removed as Igor found out).

I am more and more convinced -- since luckily, HTTPClient was package protected class -- that we need to completely redo this, and for RRB (and any other component that needs "preconfigured" HTTPClient4 instance), the "factory-like" methods should be factored out, and moved out from this package, as those have nothing to do with RemoteRepositoryStorage. As you noted, RRS has very specific requirements against used HC4 client, like to have it configured to refuse accepting redirects to collections, RemoteStorageContext juggling, and since now, to have all clients share same pool etc.

Simplest would be to -- similarly as we have for AHC in org.sonatype.nexus.ahc package -- to provide a HC "factory" that would give you preconfigured instances. HttpClientUtil -- or it's successor _might_ (or might not) hook onto that factory, but it would provide it's own connection pool shared across all clients created on behalf of proxy repositories.

jdillon:
BTW, for the rrb case, it would be better to avoid needing to worry about the http/transport issues at all and simply ask the proxy repository for the content.  That would ensure that the request is made using the proper httpclient configuration (headers, params, user-agent, etc).  Even a preconfigured instance won't have all the right configuration on it, unless httpclient has AHC-like filters that can be put onto the client and automatically applied when executing requests.

For the mbean stuff, I'm sorta leaning towards handling this stuff via eventhandlers, so that when repo is added/removed/reconfigured we can automatically tear-down, then build-back up the proper mbeans to expose various aspects of the repository.

cstamas:
We have many components even today creating HttpClients (HC3, HC4 and AHC) all over Nexus: RRB, LVO, GPG validator, Insigh just to name some. So, I really think we need to make this step (maybe not in 2.2, but...)

This is exact reason why I created https://issues.sonatype.org/browse/NEXUS-4689
(even if we decide to use HC4 API instead of some "shim API" as issue initially describes, but the main point is to have one source/factory of client instances instead littered them all over the place).

And for MBean management, I do agree with you: listening to repo-added, repo-removed should work better. But, AFAIK, unsure would that make sense, since all them will use _one shared pool_.

And for RRB, I am unsure do we want to implement it this way. Originally, Proximity had ListableRemoteRepositoryStorage iface, that simply meant that "remote is listable" (it had HTTP and FTP remote repo storages, out of which FTP was listable, and same is true for Jets3t powered RRS). And the RRS handled listing on it's own... simply, if you use HTTP transport to reach S3 hosted repo, it _might work_ (it does as today), but use S3 RRS to make it listable (and much more than just listable).

In other words, today, I would keep RRB "as is", with all of it's muck and black magic, and keep it far away from RRS. For now, at least.

jdillon:
RRB doesn't need any support by nexus to do the listing, it does that by itself in its own hacky manner.  What it does need is the ability to fetch default-page views ie... stuff that ends with / which is explicitly forbidden ATM.

There is one minor hack for s3 that relies on the Server header, but otherwise it just needs the content.  I spent a few minutes and re-implemented to use remote storage request:

https://github.com/sonatype/nexus/tree/rrb-using-remote-storage

jdillon:
As for mbean stuff an the pool, ya obviously if there is only one pool then there is no need to create an mbean-per repo to expose that, but I do think we should be using more JMX exposed bits, so I'd still like to have a per-repo mbean which can expose administrative/monitoring details.

jdillon:
I've also rounded it out with a semi-port of rrb to use httpclient4 (missing all the custom headers/query-string/user-agent bits which is handled by remote storage impl) so not really a good solution:

https://github.com/sonatype/nexus/tree/rrb-using-httpclient

peterlynch:
so what's the status? is this out for now?

cstamas:
Superseded by https://github.com/sonatype/nexus/pull/567

PR #566, opened 2012-10-06
WIP: Add basic registration of mbeans

jdillon:
Sandbox to sort out how to best enable general use of JMX inside of Nexus.

### DIFF
--- a/nexus/nexus-plugins/nexus-plugin-manager/pom.xml
+++ b/nexus/nexus-plugins/nexus-plugin-manager/pom.xml
@@ -50,6 +50,11 @@
       <artifactId>nexus-proxy</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.sonatype.sisu.goodies</groupId>
+      <artifactId>goodies-jmx</artifactId>
+    </dependency>
+
     <!-- Testing -->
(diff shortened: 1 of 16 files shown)


jdillon:
atm these are brain-dead simple mbeans, but with a bit more effort we can make them dynamic and really useful.

peterlynch:
headed in the right direction +1

adreghiciu:
+1

jdillon:
Really need to sort out MXBean use as well as make some support to dynamically expose beans to avoid all the standard MBean muck.   ATM I'm only using standard MBeans to identify key bits which should be added as mbeans.

Probably also need to sort out a general MBean naming strategy.

peterlynch:
status?

jdillon:
This is still a WIP, I think there are better ways to handle this simpler, using jmxutils... so atm on hold until some of the other clean up settles down, and then I'll look at it again.

jdillon:
BTW, I'm thinking this might be better handled in an aspect weaved in at compile time. 

Some of the other timing-based code might also be better done in this manner as well.

jdillon:
Will reconsider after we have AOP support in NX how to add JMX generally to various components.

PR #567, opened 2012-10-06
HC4Provider and replacement for HttpClientUtil.

cstamas:
These changes are related to NEXUS-5291, or are direct consequences.

Main parts:
- introduction of Hc4Provider component, capable of creating HC4x preconfigured instances (aligned to global or repo-specific configurations), centralising the HC4x instance creation and it's configuration. This component offers preconfigured HC4 instances for subsystems like LVO, GPGSKS etc, taking away the concern of client creation, configuration and managing it (idea is "grab'n'use and finally toss instance away"). The pattern for these subsystems is to keep reference to Hc4Provider rather then to HC4 instance, and simply "grab" a client whenever some request is about to be made, and toss it away immediately it's done. 
- replacing "static util" class HttpClientUtil with HttpClientManager, specific to HC4x powered RRS implementation, that manages proxy repository specific needs and "extra configurations" on the HC4x instances, used by Nexus Proxy repositories.
- centralising the ClientConnectionPool to one single point to be able to "cap" instance wide values like pool max size etc -- done in Hc4Provider
- implemented proper connection eviction (not present before, in case of HC4 RRS)
- introduced exceptions to "signal" the case when outbound request failed, but not because of a remote party, but for cases when transport is incapable to perform such task (like connection pool depleted or max connection limit reached). In such cases, the affected proxy repository will not autoblock, nor will NFC get populated.
- REST API modified to honor these new exceptions and emit 503 Service Unavailable

The targeted subsystems (LVO, GPGSKS, etc) that will use newly introduced Hc4Provider usually does not make "bursted" requests, hence, the instance offered to them has NO connection reuse (pooling) intentionally set (hence, less problems to care about too).

Proxy repositories got HttpClientManager to be used by RRS for proxies only, and manages creation and release of HTTP clients needed for proxy repositories. Context muck is shifted to actual RemoteRepositoryStorage implementation.


jdillon:
A few places plexus annotations are used... others guice... why are we adding new code that uses plexus annotations?

cstamas:
Only the two event inspector is Plexus, as mixing plexus and inject annotations in class hierarchy leads to only one of them be used (i think @Inject ones).

jdillon:
@cstamas I'm not sure that is true... I just tried creating 2 inspectors one with guice another with plexus annotations:

```
import org.codehaus.plexus.component.annotations.Component;
import org.sonatype.nexus.proxy.events.EventInspector;
import org.sonatype.plexus.appevents.Event;

@Component(role = EventInspector.class, hint = "PlexusEventInspectorImpl")
public class PlexusEventInspectorImpl
    implements EventInspector
{
    @Override
    public boolean accepts(Event<?> evt) {
        System.out.println("PLEXUS: " + evt);
        return false;
    }

    @Override
    public void inspect(Event<?> evt) {
    }
}
```

```
import org.sonatype.nexus.proxy.events.EventInspector;
import org.sonatype.plexus.appevents.Event;

import javax.inject.Named;
import javax.inject.Singleton;

@Named
@Singleton
public class GuiceEventInspectorImpl
    implements EventInspector
{
    @Override
    public boolean accepts(Event<?> event) {
        System.out.println("GUICE: " + event);
        return false;
    }

    @Override
    public void inspect(Event<?> event) {
    }
}
```

And I see both of them running as expected:

```
jvm 1    | GUICE: org.sonatype.nexus.configuration.ConfigurationChangeEvent@18c78bdb
jvm 1    | PLEXUS: org.sonatype.nexus.configuration.ConfigurationChangeEvent@18c78bdb
```

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
```diff
@@ -256,14 +263,9 @@ public void storeItem( final ProxyRepository repository,
...
-            throw new RemoteStorageException(
-                e.getMessage()
-                    + " [repositoryId=\"" + repository.getId()
-                    + "\", requestPath=\"" + request.getRequestPath()
-                    + "\", remoteUrl=\"" + remoteUrl.toString()
-                    + "\"]",
-                e
-            );
+            throw new RemoteStorageException( e.getMessage() + " [repositoryId=\"" + repository.getId()
+                + "\", requestPath=\"" + request.getRequestPath() + "\", remoteUrl=\"" + remoteUrl.toString() + "\"]",
```
jdillon, line 269: superfluous use of toString()

jdillon:
Probably should also be calling method.releaseConnection() in a finally block to handle Error and other cases to ensure that the internals are cleaned up, as shown on the quickstart docs:

http://hc.apache.org/httpcomponents-client-ga/quickstart.html

cstamas:
Nope, releaseConnection() must NOT be called always, only in exceptional cases (errors). In "normal" case the wrapped response entity is returned, and the one closing it's stream will release it too. If you'd release connection in finally block, you'd lost the content you fetched too, before it would get cached or consumed in any way.

cstamas:
Applied comments and squashed changes into one commit. Changes are:
- sisuized event inspectors
- shutdown event inspector was wrongly defined as Async
- UTs (former HttpClientUtilTest) for HttpClientManager and HttpClientRemoteStorageTest fixed

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/DefaultHc4Provider.java
```diff
@@ -0,0 +1,293 @@
...
+     */
+    private final ApplicationConfiguration applicationConfiguration;
+
+    /**
+     * Parameters source for HC4.
+     */
+    private final Hc4Parameters parameters;
+
+    /**
+     * Shared client connection manager. Not pooling!
```
nabcos: "Not pooling!" needs a more detailed comment IMO. IIUC the actual connection manager that is used is a pooling connection manager, but this Provider overrides the keepAlive setting so connections are closed immediately?

cstamas: HC4 contains two connMgr implementations: "basic" and "pool". In short, "pool" one is meant to be used in multithreaded fashion, that's the only connMgr usable with them (if we do not count the deprecated oldie connMgr). Hence, for multi threaded use case, the only connMgr is pooling, but it does not mean that httpClient using it will pool (it depends not only on config, but also on remote response, response subsystem is making -- it might set Connection:close per request, etc).

So, IMO, the fact that "pool" connMgr is used does not mean nothing yet in HC4, it does not mean you WILL pool. But true, it could be noted in Javadoc that pooling is explicitly disabled, but that connMgr is used to make client instance thread safe.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4Provider.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    ClientConnectionManager create( RemoteStorageContext context );
+
+    /**
+     * Advanced. Primarily for subsystem that wants to utilize connection pooling and have full control of it's
+     * connection manager, while reusing all the logic from this provider. So far, that subsystem is Nexus Proxy
+     * repositories. Connection manager created using this method must be managed (ie. properly shut down) by the party
+     * that initially created it. This provider will only create it, but will not manage it.
+     * 
+     * @param designator
```
adreghiciu: This param does not exist

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4Provider.java
```diff
@@ -0,0 +1,82 @@
...
+     *      level authentication</a>
+     */
+    HttpClient leaseShared();
+
+    /**
+     * Releases the leased globally shared instance. TODO: is this really needed?
+     * 
+     * @param httpClient
+     */
+    void releaseShared( HttpClient httpClient );
```
adreghiciu: Why do you add the params at all of there is no javadoc?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4Provider.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param context
+     * @return connection manager with pool evicting thread started.
+     */
+    ClientConnectionManager create( RemoteStorageContext context );
+
+    /**
+     * Advanced. Primarily for subsystem that wants to utilize connection pooling and have full control of it's
+     * connection manager, while reusing all the logic from this provider. So far, that subsystem is Nexus Proxy
+     * repositories. Connection manager created using this method must be managed (ie. properly shut down) by the party
+     * that initially created it. This provider will only create it, but will not manage it.
```
adreghiciu: This looks like a copied javadoc from above ^

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4ParametersImpl.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.concurrent.TimeUnit;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.sonatype.nexus.proxy.storage.remote.RemoteStorageContext;
+import org.sonatype.nexus.util.SystemPropertiesHelper;
+
+/**
+ * Simple parameters implementation that neglects the "designator" muck for now. It could be used to differentiate
```
adreghiciu: What is a designator?

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4ParametersImpl.java
```diff
@@ -0,0 +1,127 @@
...
+ * there is a pool involved. While it might be, it is merely limiting the connection count (max and per route), as in
+ * case of shared connection manager, pooling is intentionally shut down!
+ * 
+ * @author cstamas
+ * @since 2.2
+ */
+@Singleton
+@Named
+public class Hc4ParametersImpl
+    implements Hc4Parameters
```
adreghiciu: Do we expect divers implementations? Like?

cstamas: Like sourced from AppContext, sourced from System Properties (this one here) and others. @nabcos already proposed to renamed classes to better reflect from where are they sourced, so in this case this class should better be named as `SystemPropertiesHc4Parameters` or such.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4ParametersImpl.java
cstamas: Will update Javadoc. "Designator" is a dropped idea that was present initially (ability to "target" some parameter per proxy repository, not available anymore as pool is shared).

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4ProviderImpl.java
```diff
@@ -0,0 +1,309 @@
...
+        this.sharedHttpClient = null;
+    }
+
+    @Override
+    public synchronized void shutdown()
+    {
+        if ( sharedConnectionManager != null )
+        {
+            sharedConnectionManager.shutdown();
+        }
```
adreghiciu: Shouldn't sharedConnectionManager be set to null here?

cstamas:
CI job of changes up to 8c7736db4be6b5c5737da153e8ade8d6802d7570 passed on CI:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/434/

Build failed due to a source file missing src header, fixed in subsequent commits.

cstamas:
CI job up to change 4679bf1b6369dd7297f6840b8f026730cd9c0bba passed:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/436/

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
```diff
@@ -456,16 +452,13 @@ HttpResponse executeRequest( final ProxyRepository repository,
             if ( stopwatch != null )
             {
                 stopwatch.stop();
-                timingLog.debug(
-                    "[{}] {} {} took {}",
-                    new Object[]{ repository.getId(), httpRequest.getMethod(), httpRequest.getURI(), stopwatch }
-                );
+                timingLog.debug( "[{}] {} {} took {}", new Object[] { repository.getId(), httpRequest.getMethod(),
+                    httpRequest.getURI(), stopwatch } );
```
jdillon: Should guard this, there is enough object creation and method calls here to warrant isDebugEnabled()

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/httpclient/EvictingThreadTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void connectionEvictedIn5Seconds()
+        throws Exception
+    {
+        final ClientConnectionManager clientConnectionManager = mock( ClientConnectionManager.class );
+
+        final EvictingThread underTest = new EvictingThread( clientConnectionManager, 1000, 100 );
+        underTest.start();
+
+        Thread.sleep( 300 );
```
jdillon: Tests that require sleeping to validate functionality are generally bad.

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/httpclient/Hc4ProviderImpl.java
```diff
@@ -0,0 +1,483 @@
...
+     * Constructor.
+     * 
+     * @param applicationConfiguration the Nexus {@link ApplicationConfiguration}.
+     * @param userAgentBuilder UA builder component.
+     * @param multicaster the event multicaster
+     */
+    @Inject
+    public Hc4ProviderImpl( final ApplicationConfiguration applicationConfiguration,
+                            final UserAgentBuilder userAgentBuilder,
+                            @Nullable final ApplicationEventMulticaster multicaster )
```
ifedorenko: What is the scenario when multicaster == null?

cstamas: In some very old "legacy" UTs we use PlexusTestCase and the object graph is not complete... Will take a peek into them.

cstamas:
CI job with changes up to rename (up to 34b19c33c9529be61821d07362b749fa76e53697 inclusive) +1'd:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/440/

New job with package rename on the way:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/441/

cstamas:
Locally started full build (`mvn clean install -U -Pit`) passed with latest changes (same as job 441 is building):

```
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 1:34:35.076s
[INFO] Finished at: Wed Oct 10 15:22:52 CEST 2012
[INFO] Final Memory: 136M/374M
[INFO] ------------------------------------------------------------------------
[cstamas@zaphod nexus (hc4provider)]$ 
```

cstamas:
CI job 441 passed fine:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/441/

ifedorenko:
+1 please merge

PR #569, opened 2012-10-10
[NEXUS-5233] Use networkaddress.cache.ttl instead of sun.net.inetaddr.tt...

jdillon:
...l

https://issues.sonatype.org/browse/NEXUS-5233


cstamas:
+1

nabcos:
+1

PR #570, opened 2012-10-11
Improving the build: javascript handling

nabcos:
- Introduce default  yuicompressor configuration for nexus-plugins
- Ease UI debugging by not compressing js by default, only on release (profile optimize-js)
- Align js location and get rid of redundant resource bundle exports
- Add jslint for "newer" javascript files in nexus-webapp (Nexus.\* namespace)


jdillon:
Remember to enable optimize-js profile on grid builds once this is in.

jdillon:
Running -Poptimize-js causes the build to fail due to failure on warnings being enabled...

```

[INFO] generate aggregation : /Users/jason/ws/sonatype/nexus-oss/nexus/nexus-webapp/target/ui-compressed/js/sonatype-all.js
[INFO] sonatype-all.js (1080227b)
[INFO] generate aggregation : /Users/jason/ws/sonatype/nexus-oss/nexus/nexus-webapp/target/ui-compressed/style/sonatype-all.css
[INFO] sonatype-all.css (7942b)
[INFO] nb warnings: 1994, nb errors: 0
```

1994 warnings... neato.  

Unless we plan on fixing these warnings, we should remove:

```
<maven.yuicompressor.failOnWarning>true</maven.yuicompressor.failOnWarning>
```

and/or set this to false.

BTW, I've noticed that when optimize is enabled, jslint does some extra checking... unsure why this is.

nabcos:
oh, I did not run the optimize profile after making jslint mandatory - although I find it weird because I tried to configure jslint to only check the new namespace, which also works without the optimize-js

jdillon:
yar, as I mentioned when optimize is enabled jslint is kicked in

jdillon:
anyhoo I'm +1 if rew fix the failOnWarning, it looks like the server is behaving well with this change.

nabcos:
Weird, it's the compress goal that is warning, but I don't think it runs the full jslint check, more of a 'where can you shave off some bytes'. It also looks like it ignores `<jswarn>false</jswarn>` configuration. 

EDIT: It's not ignoring, but failOnWarning forces jswarn to true.

nabcos:
Last push addresses the failing build with optimize-js profile. I also cleaned up the commit history and added better comments.

jdillon:
+1

adreghiciu:
+1

PR #571, opened 2012-10-12
[NXCM-4764] Locking resource did not used the correct path

adreghiciu:
If for example path was  ending with g/a/a-v.jar the path was incorrectly set to "g".


cstamas:
+1

nabcos:
+1

PR #573, opened 2012-10-13
Deprecate old httpclient components

jdillon:
Marks all non-httpclient4 components as deprecated with details.


adreghiciu:
+1

peterlynch:
+1 - except an issue number in Jira should be tied into this commit message and marked as release note item.

jdillon:
Created issue:

https://issues.sonatype.org/browse/NEXUS-5307

Added to merge commit message.

PR #581, opened 2012-10-18
[NEXUS-5341] Encode HTML entities only for outgoing data

nabcos:


nexus/nexus-rest-api-model/src/main/java/org/sonatype/nexus/rest/model/HtmlEscapeStringConverter.java
```diff
@@ -46,8 +46,9 @@ public boolean canConvert( Class type )
...
 
     @Override
-    public Object fromString( String str )
+    public String toString( Object str )
     {
-        return StringEscapeUtils.escapeHtml( str );
+        return StringEscapeUtils.escapeHtml( (String)str );
     }
+
 }
```
cstamas, line 54: See org.sonatype.nexus.rest.model.XStreamConfiguratorLightweight as this affects nexus-client-core too (actually all clients using it)?
Shouldn't in clients this happen in opposite way then, if this change is applied on server?

nabcos, line 54: Good question. I was under the impression that this would not affect REST clients, because the behavior would be effectively the same as before:
- before the change, resources and clients would see html entities in incoming data
- after the change, resources sees literal values and clients see html entities in incoming data

If the clients use the same XStream configuration, they always did the wrong thing, because clients would escape the already escaped values in incoming data (right?):

`client -*server-side escape*-> resource -*client-side escape*-> client`

So after this change, they would do:

`client -*client-side escape*-> resource -*server-side escape*-> client`

peterlynch:
+1

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/461

adreghiciu:
+1

nabcos:
Closing this (and referring) pulls because of inconsistencies in UI
assumpations about incoming data. More work is needed to be able to handle
HTML/XML uniformly.

Parts of UI expect "internal" data to not be escaped, e.g. in artifact links
sent back from indexer plugin - query parameters are broken because they
contain '&', other parts expect escaped (config file viewer).

ATM I think this scheme would be best:
- Nexus REST backend never escapes HTML characters. (best for Java-based REST clients that will not render responses to a browser)
- Handle HTML escaping in javascript for all ajax responses.
- Fix all UI to always expect escaped HTML.

Most changes in here and the referenced pulls stay valid, though, so the
branches will stay around for now.

PR #582, opened 2012-10-18
Prepare for test suite

adreghiciu:
1. Switch to not install plugins if they exist in bundle used by IT
   This switch will allow running ITs against a bundle without installing the plugins specified by IT, in case that those plugins already exist in bundle (compared by groupId / artifactId)
2. Make possible to filter nexus bundle coordinates passed to its
   The new ITs are passing as ctor the coordinates fo teh bundle to run against. Make possible to filter that value using project/system properties

### DIFF
--- a/nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundle.java
+++ b/nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundle.java
@@ -15,15 +15,20 @@ package org.sonatype.nexus.bundle.launcher.support;
 import static org.sonatype.nexus.bootstrap.monitor.CommandMonitorThread.LOCALHOST;
 import static org.sonatype.sisu.bl.jsw.JSWConfig.WRAPPER_JAVA_MAINCLASS;
 import static org.sonatype.sisu.filetasks.FileTaskRunner.onDirectory;
+import static org.sonatype.sisu.filetasks.builder.FileRef.file;
 import static org.sonatype.sisu.filetasks.builder.FileRef.path;
 import static org.sonatype.sisu.goodies.common.SimpleFormat.format;
 
 import java.io.File;
+import java.io.FileFilter;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.net.ConnectException;
+import java.util.Enumeration;
 import java.util.Map;
 import java.util.Properties;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Provider;
@@ -50,7 +55,9 @@ import org.sonatype.sisu.bl.support.TimedCondition;
 import org.sonatype.sisu.bl.support.port.PortReservationService;
 import org.sonatype.sisu.filetasks.FileTaskBuilder;
 import org.sonatype.sisu.goodies.common.Time;
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
+import com.google.common.collect.Maps;
 
 /**
  * Default Nexus bundle implementation.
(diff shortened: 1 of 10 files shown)


peterlynch:
+1 looks ok, but I suggest a feature build to be sure.

adreghiciu:
Here it is: https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/451/

adreghiciu:
CI voted +1

jdillon:
Does this need to be coordinated with plugin changes? I saw some plugins were updated for this stuff...

adreghiciu:
This will not break any existing code. When this is in we can merge the changes to plugins to make use of what we did here.

nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundle.java
```diff
@@ -258,6 +277,122 @@ protected boolean applicationAlive()
...
+    public void doPrepare()
+    {
+        super.doPrepare();
+        if ( useBundlePluginsIfPresent )
+        {
+            removePluginsPresentInBundle();
+        }
+    }
+
+    private void removePluginsPresentInBundle()
```
nabcos, line 290: Wrongly named? IIUC this removes plugins in sonatype-work/plugin-repository

adreghiciu, line 290: No. The switch is to prefer bundle plugins but the way it works is that it lays down then overlays may add plugins (that are added to sonatype-work plugin repository, plugins that are scanned and removed if they exist in bundle

adreghiciu, line 290: @ben if you can think of a better name let me know

nexus/nexus-test-harness/nexus-test-environment-maven-plugin/src/main/java/org/sonatype/plugin/nexus/testenvironment/AbstractEnvironmentMojo.java
```diff
@@ -1326,4 +1341,87 @@ private void createMarkerFile( String markerName )
...
+                    {
+                        plugins.put( gaCoordinates.get(), plugin );
+                    }
+                }
+            }
+        }
+        return plugins;
+    }
+
+    private Optional<String> getPluginGACoordinates( final File mainJar )
```
nabcos, line 1373: Duplicated code?
https://github.com/sonatype/nexus/pull/582/files#L1R342

nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundle.java
nabcos, line 290: Hm, removeRedundantPluginsInWorkDir? The name to me suggests that this method removes plugins from the bundle, but I see where you're coming from. removePluginsAlreadyPresentInBundle?

nabcos, line 290: Or maybe just a bit of javadoc to clarify behavior

adreghiciu:
Yes, some code is duplicated as launcher and nexus env plugin they do not have a common ground

nabcos:
+1

nexus/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundle.java
nabcos, line 290: And maybe I should change my HipChat alias to @nabcos as well to not alert someone from github to our discussions =)

PR #583, opened 2012-10-18
Use litmus avoid injection

jdillon:
Improvements to tests.

Going to leave this pull asis, will fix others later... way too much for a single pull.


peterlynch:
1

jdillon:
Looks like this is causing a few tests to fail downstream in nexus-proxy... atm unsure why.

jdillon:
dropped last commit, unrelated to test impl changes, and setting workingDirectory causes issues, will sort out other changes in other commits to clean up profiles.

nabcos:
+1

peterlynch:
setting workingDirectory remains an issue with sonar generating proper test coverage so glad you avoid that here - will need to look into fixing patching sonar to do the right thing because setting jacoco.exec report path does not help

peterlynch:
+1

PR #584, opened 2012-10-18
Remove RestInfo, which uses plexus specific apis to build up, and appear...

jdillon:
...s to not even be shown in the console.

Drop unused code that depends on plexus ComponentDescriptor


jdillon:
IIUC this was replaced by enunciate generate documentation.  The current plugin-code does not even show this information.  IMO it should be removed.

cstamas:
Note: the payload of restinfos is correct, and enlists all the "provided" REST resources by given plugin...

+1 to remove this

adreghiciu:
+1

PR #590, opened 2012-10-19
Use jboss tattletale plugin

jdillon:
Use official jboss tattletale maven plugin.

Includes a bunch more reports too.  Configuration has the same excludes (which we really should fix).  Changed where the profile for this exists, its now a profile in nexus/pom.xml to enable the module, vs a profile in the module itself.


nexus/pom.xml
```diff
@@ -1788,13 +1802,21 @@
...
     <module>nexus-stories</module>
     <module>nexus-test</module>
     <module>nexus-test-harness</module>
   </modules>
 
   <profiles>
+    <profile>
+      <id>tattletale</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
jdillon: May actually want to make this only run on grid things and for releases, not always run.

peterlynch: would like to avoid 'activeByDefault' for any profile.

Instead propose we add this to 'it' profile

jdillon: agree

peterlynch:
+1

nabcos:
+1

adreghiciu:
+1

cstamas:
+1

peterlynch:
added note to not use the other one

https://github.com/sonatype/tattletale-maven-plugin/blob/master/README.md

PR #592, opened 2012-10-19
[NXCM-4559] remove nexus-perf module

peterlynch:
moved entire nexus-perf to internal nexus-grid project until we have a better way of publishing our perf test data


jdillon:
+1

PR #594, opened 2012-10-19
WIP: Move current REST API into a plugin

cstamas:
And mark it as "legacy". This makes Core Restlet-free, plexus-restlet-bridge free
and life better than before.

Any plugin contributing REST API needs to declare a dependency on this new plugin
for now (to use "legacy REST infra in unmodified form"), but once we introduce some other 
REST imple, by updating resources in plugins
the dependency will cease too. Ultimately, nexus-legacy-restlet-api-plugin will
remain alone, with no dependant, and easily removed.

Also, since Core classloader now does not have Restlet, plexus-restlet, and
any other "webtech" loaded (is now a real headless app), the doors are open
to introduce any other remoting we want. _As a plugin_, to keep us
away from mistakes in past!

Also, while security is needed in core, and is in core, the one module
security-rest-api, that is adding security rest apis is now also in this
plugin as it also uses "legacy REST".


cstamas:
Further notes: If we move out the CommonsHttpClient3x powered RRS out of core, the HC3x is also movable out from core (along with modified connMgr from nexus-proxy). Same for AHC (the AHC provider in nexus-proxy).

adreghiciu:
+1

jdillon:
tested running locally... looked like it was working to me, only a cursory test however: +1

peterlynch:
agree this is the correct direction from testing perspective

nabcos:
I have [some test failures in LDAP tests](https://gist.github.com/3f10795dadb8905179d0), did you run the full testsuite yet?

cstamas:
LDAP test failures present for me too, caused by "provided" scope on nexus-legacy-rest-api plugin (and those are not passed on as transitive to plugin dependants). Fix on the way.

jdillon:
I think we should rename these modules to nexus-restlet-1x-model and nexus-restlet-1x-plugin.  "legacy" here isn't super-useful, since we don't actually have a replacement for it yet.  So instead of assuming its legacy (which it hopefully will be soon) just call it what it is... REST support for Restlet 1.x

jdillon:
I also think we should verify that we can make a Siesta/Jersey plugin for new REST resources, to ensure that nothing else needs to get changed... before we merge this.

jdillon:
gid "org.sonatype.nexus.plugins.rest" probably isn't needed...

jdillon:
BTW, I was able to get siesta up in a plugin next to this one... still some more validation needs to be done, but this is defs on the right path.

PR #600, opened 2012-10-21
More profile cleanup

jdillon:
Based on https://github.com/sonatype/nexus/pull/599

PR #601, opened 2012-10-21
Testsuite cargo cleanup

jdillon:
Cleaned up the cargo configuration, actually have it run (hasn't run in at least 8 months, due to changes which would have causes jetty6 and jetty8 tests to fail if they had).

Removed profile, use forkmode=always (notes inline as to why, its a trade off for a bit slower running testsuite execution).

Based on https://github.com/sonatype/nexus/pull/600


jdillon:
https://builds.sonatype.org/job/nexus-oss-testsuite-feature/2/jdk=java-6x,label=linux/ SUCCESS

https://builds.sonatype.org/job/nexus-oss-testsuite-feature/jdk=java-6x,label=linux/2/testReport/org.sonatype.nexus.integrationtests.nexus3860/

peterlynch:
+1

also further changes in sonatype/nexus#605 should go with

jdillon:
Yup, I'll merge in 605 when this goes in.

PR #602, opened 2012-10-21
Core it logging fixes

jdillon:
Fix logging in core-its, still not 100% controllable for test execution, but at least now they are all using the same logging system.

Keeps test and embedded NX server logging in the same location (sysout) and with failsafe+testng capturing each test due to forkmode=always we end up with a consistent log for each test. 

Based on https://github.com/sonatype/nexus/pull/601

PR #603, opened 2012-10-23
Kill site configuration

jdillon:
Lets kill all of this and then reimplement it when we plan to start building sites again.


jdillon:
Based on https://github.com/sonatype/nexus/pull/602

PR #604, opened 2012-10-23
Kill selenium

jdillon:
Remove unmaintained selenium muck.


jdillon:
Based on https://github.com/sonatype/nexus/pull/602

nexus/nexus-test-harness/nexus-test-harness-its/pom.xml
```diff
@@ -260,20 +278,79 @@
...
+                    -->
+                    <artifactItem>
+                      <groupId>org.apache.tomcat</groupId>
+                      <artifactId>apache-tomcat</artifactId>
+                      <version>6.0.29</version>
+                      <classifier>bundle</classifier>
+                      <type>tar.gz</type>
+                      <outputDirectory>${project.build.directory}/containers</outputDirectory>
+                      <!-- Excludes do not function with tar.gz bundles, and tar.gz bundle unpacking is very sloooooow -->
+                      <!--<excludes>**/webapps/**</excludes>-->
```
peterlynch: so why not use zip?

nexus/nexus-test-harness/nexus-test-harness-its/pom.xml
```diff
@@ -260,20 +278,79 @@
...
+                    <!-- Required for nexus3860/* (cargo) tests -->
+                    <artifactItem>
+                      <groupId>org.sonatype.nexus</groupId>
+                      <artifactId>nexus-webapp</artifactId>
+                      <type>war</type>
+                      <outputDirectory>${project.build.directory}/nexus/webapp</outputDirectory>
+                    </artifactItem>
+                    <!--
+                    Versions here must be kept in sync with nexus3860/* (cargo) tests
+                    The Tomcat distributions referenced are not actually available in central.
```
peterlynch, line 301: so the understanding is these must be available in third party repo? is ok to me since these are ITs

peterlynch:
this was very hard to review - I agree with the sentiment of all the changes however. There are probably a few things in there I missed. I want to see the test suite results after running these changes.

Perhaps a better way to do this would have been create a teste branch to hold all these changes. For each set of related changes for this branch:
- make a pull request to merge to that test branch
- immediately merge your pull request to the test branch
- rinse repeat, continuing your improvements

This way we can easily see all the related changes individually and as a whole on the same branch.
If someone had an issue with a particular merge, it is easier to revert the merge or parts of it after the  fact.

jdillon:
In this case, given this is built upon another pull, best to look at the actually change:

https://github.com/sonatype/nexus/commit/a315fb7503c3a031042be21db20faf16104e5f36

nexus/nexus-test-harness/nexus-test-harness-its/pom.xml
jdillon: There are no zip artifacts for these guys, these aren't official deployed artifacts by the tomcat group, these are things that have been deployed in rso's thirdparty repository.

https://repository.sonatype.org/content/repositories/third-party/org/apache/tomcat/apache-tomcat/

jdillon:
@peterlynch I think once the build is generally cleaned up them, pulls like this won't happen very often.  Its a special case here given how deteriorated the build is and the amount of change needed to fix it up.  Even more so that often clean up of one section depends on others, so I based the branches off of each other.

I do dot think adding more branches and merging and other process here is going to help improve things very much.  And its a lot more work to be done just to make github show you that only the last change in the set of changes here is specific to this.

I'm asking for a little lenience here, since this clean is not an easy task, and I really don't need to jump through a few more hoops just to get the pull request to look pretty.

jdillon:
```

Perhaps a better way to do this would have been create a teste branch to hold all these changes. For each set of related changes for this branch:

make a pull request to merge to that test branch
immediately merge your pull request to the test branch
rinse repeat, continuing your improvements
```

I'm not sure how that is any different that what is done now... except that the merge target is master.  If I had started out with a branch 'build-cleanup' based on master, then created branches off of that 'build-cleanup' branch, then the pulls would still have a bunch of changes, and since these groups of changes all play off the previous one it would have ended up exactly the same.

peterlynch:
the huge difference in my suggestion is that each set of changes ( kill selenium, fix cargo, etc. ) will be a pull request in github that we can comment on in it's own context without the other noise. Meanwhile you merge the the pulls immediately so are not hindered waiting for feedback. You've would have no more work to do other than creating one extra branch.

I can live with what you've done here, it's just a suggestion for next time that could be worth trying out to prove me wrong.:)

nexus/nexus-test-harness/nexus-test-harness-its/pom.xml
peterlynch: that's my point, just deploy the zip files instead, nvm I'll do it

peterlynch: sonatype/nexus#605

PR #605, opened 2012-10-23
update cargo containers

peterlynch:
- drop tomcat 5.5 as it is EOLed already
- add Tomcat 7
- use zip artifacts instead
- exclude web apps included with tomcat bundles to speed up startup time

as previously mentioned these containers are in the RSO third party repo.


jdillon:
nice

PR #606, opened 2012-10-24
Simplify build configuration

jdillon:
Loads of build configuration simplifications and clean up. 

This is probably 90% of the build issues fixed and cleaned up... still a bit more to go.

The main intent of this pull is to reduce build complication, by reducing redundant pom configuration, making effective use of DM and PM, avoiding unnecessary module coupling.

Here are more details for stuff as I look over what is in this pull for your reference...

---

One of the tenets here is to avoid using scope in DM entries.  I found that many places entries in DM were set to scope=test, and then either forced back to scope=compile in many many locations, or always used as scope=test in children poms.  The clarity of always setting the proper scope on the used entry is also a significant benefit, instead of having to dig through the DM hierarchy to find out that something is scope=test or scope=compile.  The default scope=compile should be preferred in DM and then users of the dependency can alter the scope as needed.

Some effort to remove use of project.version (or project.groupId) has also been implemented, these should generally be avoided in dependency declarations, as the value could change if consumed as a parent of a downstream project which has different values that what is expected... best to just avoid using them.  Still some work left to do here, but its fairly minor now.

Lots of pointless pom comments removed, drop pointless configuration (like packaging=jar which is the default).  Some normalization of basic pom elements (probably a bit more to normalize order of dependencies/builds/etc.

A few minor renames for clarity (ie AbstractNexusTestCase, etc).

Updated plugins to ue nexus.plugin.name and nexus.plugin.description to configure these values for app-lifecycle, and used simpler names not bound to the project.name.

Dropped a few unused dependencies, and unrolled a few as well (ie. ldap plugin used plexus-digest solely for its Hex helper, when it could have used Shiro's version instead and avoid the extra dependency, etc).

Killed some unused legacy crud in the plexus configuration for integration tests... probably more to be done here, but I'm not 100% sure what all that is for ATM.

Generally avoid use of properties for versions when they are not required to shared between modules, or not declared many times in the same module.  Removed most redefinitions of dependencies and plugins with same or different versions, generally prefer DM or PM for defining a version, some exceptions however for dependencies which are module specific and used in one place only.  Here we don't want to add pointless DM entries for a dependency that is only every used in one place, and really never should be used anywhere else.  A side effect of having these properties is that folks feel free to start using them to pull in other dependencies, and not putting them into DM properly.  It also means that when they are removed, then often the property doesn't get removed.

Prefer using DM to set excludes for common dependencies, instead of redefining the excludes everywhere (potentially more here to clean up).

Added common resources/testResource definition in parent to allow filtered + non-filtered, renamed any directories needed to ensure functionality with new locations for filtered bits.

Updates plexus-component-metadata to use latest qdox which can parse latest sources, enable this on all modules instead of adding this to each module that might need it.

Added some more logging to the environment plugin... for my own sanity mostly so I know wtf its doing (and not just hanging for 30s after allocating ports), still not showing all of the detail however.  Nuked UnpackTestsMojo unsure what this was for, it was causing some compile errors, easier to make it go way (if its really needed, we can bring it back and fix it).

Update to latest jsoup.

Disabled jettyconsole, we don't use this... and it does take a bit to run.


adreghiciu:
Dude, if it builds +1

jdillon:
I'll link to a it build once its done, a few more bits to sort out.

jdillon:
I believe most of the issues are now squashed, pending grid resources to sanity check.

jdillon:
https://builds.sonatype.org/job/nexus-oss-its-feature/466/ SUCCESS

adreghiciu:
Go ahead

cstamas:
GO, DIEGO, GO!

PR #608, opened 2012-10-24
Move Spice Timeline into nexus-timeline-plugin

cstamas:
Spice-Timeline (https://github.com/sonatype/spice-timeline) was yet another external project used by Nexus only.

This change:
- removes this dependency and moves the codebase into nexus-timeline-plugin, as it is the only consumer of spice-timeline component.
- spice-timeline got multiple fixes (version 3.0), bringing them over
- lucene bumped to 3.6.1
- preventing it from being memory hog (fix applied as Igor proposed,
  no more error and leak prone "iterator like" constructs but way simpler
  callbacks)
- improving efficiency by using Lucene 3.5 APIs and sharing Searchers
  (older version was creating instance per thread, hence, all the caches
  were created too)

Original repo was here (still there, but warning added):
https://github.com/sonatype/spice-timeline

Now, timeline is directly merged into nexus-timeline-plugin, with all the improvements.

See
https://issues.sonatype.org/browse/NEXUS-5351


peterlynch:
significant enough change for a jira issue so it gets recorded in release notes

peterlynch:
+1 with jira issue outlining benefits

cstamas:
Created https://issues.sonatype.org/browse/NEXUS-5351

jdillon:
+1

jdillon:
BTW, since this is the only plugin using rome, can we see if its happy with the latest 1.0 release (we are using 0.9 ATM):

https://issues.sonatype.org/browse/NEXUS-5323

adreghiciu:
+1

cstamas:
Rome update to 1.0 immediately hit this very same problem:
http://www.gavinwillingham.com/classloading-in-rome.html

cstamas:
Leaving Rome at 0.9 for now, as after several attempts could not circumvent the Rome CL issues, will need a bit more. Rebasing and leaving 0.9 for now, as it works without any problem.

cstamas:
CI voted +1:
https://builds.sonatype.org/job/nexus-oss-its-feature/468/

Despite being red, all ITs passed, but one newly added file EntryListCallback was missing a license header.

PR #609, opened 2012-10-24
Extend RequestProcessor to be able to block item retrieval

adreghiciu:
Extends RequestProcessor to be able to block an item retrieve request.

For example a request processor could make a request to Insight to check if the item to be retrieved has any vulnerabilities.
Another example an blacklist could stop retrieval on an artifact that is know to have problems.

Also implemented the blocking to collections in case of a repo that is not browseable by using a request processor.

---

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/RequestProcessor.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/repository/RequestProcessor.java
@@ -14,9 +14,13 @@ package org.sonatype.nexus.proxy.repository;
 
 import javax.inject.Singleton;
 
+import org.sonatype.nexus.proxy.AccessDeniedException;
+import org.sonatype.nexus.proxy.IllegalOperationException;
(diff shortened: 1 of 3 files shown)


cstamas:
+1. As we talked over phone, this is also the right direction for the procurement2.

nabcos:
+1

cstamas:
As I said +1 is okay, but the "browsable" check should IMO remain in core, as with this solution (moved to reqProcessor that is a component), there might be some ordering problem later. Like, we have some "heavy lighting" reqProcessor that does a LOT, only to have next one say "no, this is not browsable".

I'd stick the non-browsable check back into AbstractRepo than have it as component, I am uneasy with moving this logic out.

adreghiciu:
@cstamas okay

cstamas:
Not to mention how ReqProcessors currently are not components, they needs to be manually inserted into repo instances, hence, this above is not complete, as you removed this check, but who will insert it?

cstamas:
+1

adreghiciu:
Replaced by https://github.com/sonatype/nexus/pull/610

PR #611, opened 2012-10-24
Send exception message as reason on client error

adreghiciu:

### DIFF
--- a/nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
+++ b/nexus/nexus-rest-api/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
@@ -878,7 +878,7 @@ public abstract class AbstractResourceStoreContentPlexusResource
             }
             else if ( t instanceof IllegalArgumentException )
             {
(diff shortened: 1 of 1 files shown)


cstamas:
+100

And same should be done for 

org.sonatype.nexus.rest.artifact.AbstractArtifactPlexusResource.handleException(Request, Response, Throwable)

That is actually dupe of this here (handle method)

peterlynch:
Like the idea but...

What assurance do we have an exception message does not contain content we do not want to send to client?

Also at what layer are these messages escaped? Restlet handles it automatically?

adreghiciu:
@peterlynch we will see what messages are out there
I dunno about escaping.

peterlynch:
an option might be to make it configurable to include exception message in status, default off - until we have some better way to determine this does not expose some information we shouldn't

not sure how useful then this becomes.

jdillon:
jsoup can probably sanitize if needed:

http://jsoup.org/cookbook/cleaning-html/whitelist-sanitizer

jdillon:
Generally I think this is good, so +1, but I do share concerns about the message content... but eh, probably only one way to really find out if its sending appropriate content or not ;-)

cstamas:
Escaping is already handled by NexusStatusService, as that is the restlet component rendering these pages.

About the content, "The proof of the pudding"...  so merge this

PR #612, opened 2012-10-24
Remove the groovy-console plugin, it has lots of problems, is very under...

jdillon:
...-tested and well... not something we would ever recommend anyone using.

---

I believe this has at least a few problems… running arbitrary code inputed by user, leaks script classes each execution, no unit or integration tests, etc.

I find it very confusing that its called "nexus-groovy-console-plugin" when it doesn't really provide any console, only a task… but that is minor.

I think as we are cutting the legacy fat, we should trim this off.

@rseddon 

```
I don't know of anyone who is using this plugin.
```

@cstamas 

```
Kill it. As you say, is problematic, unsure who would use it in prod env, one groovy line and instance down :D
```


adreghiciu:
+1

cstamas:
+1

jdillon:
oops, wrong button

jdillon:
doesn't appear to be anything in bundles for this, so merging now.

PR #613, opened 2012-10-24
Use HC4 client 4.2.x from DM

jdillon:
This should help resolve issues with nexus-launcher changes... mvn was completely omitting the httpclient dependencies as jetty hc4client depends on hc4 4.1, but we have 4.2 on everything else.

Fun!


jdillon:
Appears anything using new-style testsuite framework is broken because of this (usertoken, central-secure, smartproxy, etc).

adreghiciu:
+1. I'm gonna merge this without waiting for another +1

PR #616, opened 2012-10-25
NEXUS-5250: Update to latest MI

cstamas:
This is an update to latest MI (5.0.0 as of today).

Most notable changes is fixed locking, as deadlocks are now avoided.

The "worst case" where a search operation (search on UI, or even archetype catalog retrieval, that also sources info from index) would become blocked during reindex is somewhat "fixed": context being reindexed will not be searched, and catalog for reindexed repository will be empty.


peterlynch:
+1

cstamas:
Superseded by https://github.com/sonatype/nexus/pull/637 

Closing this pull.

PR #618, opened 2012-10-25
Retry handlers fix

cstamas:
Seems that Hc4Provider misses to set retry handler (default is 3).

But since AbstractProxy class also handler retries, we actually get 3x3 retries?

(seen in RSO logs)

### DIFF
--- a/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientManagerTest.java
+++ b/nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientManagerTest.java
@@ -23,15 +23,24 @@ import org.apache.http.ProtocolException;
 import org.apache.http.RequestLine;
 import org.apache.http.StatusLine;
 import org.apache.http.client.RedirectStrategy;
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.http.impl.client.StandardHttpRequestRetryHandler;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.protocol.HttpContext;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
 import org.sonatype.nexus.apachehttpclient.Hc4Provider;
+import org.sonatype.nexus.apachehttpclient.Hc4ProviderImpl;
+import org.sonatype.nexus.apachehttpclient.PoolingClientConnectionManagerMBeanInstaller;
+import org.sonatype.nexus.configuration.application.ApplicationConfiguration;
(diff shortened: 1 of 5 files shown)


jdillon:
Aside from all the ancillary changes in this pull... +1

peterlynch:
-1 please add a unit test

jdillon:
@peterlynch the main change her is configuring StandardHttpRequestRetryHandler (an hc4 class) on the HttpClient instance... what exactly would a unit test be testing?

peterlynch:
@jdillon that values from getRetrievalRetryCount() are set as the number of retries

jdillon:
ic, so the passing of global config on to httpclient, okay.

cstamas:
Rebased against latest, added UT that tests Hc4Provider (is retryCount properly applied?), and test for HttpClientManager (is retryCount properly nullified, as these instances are used in RRS)

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientManagerTest.java
```diff
@@ -99,4 +128,21 @@ public void doNotFollowRedirectsToDirIndex()
...
+
+    @Test
+    public void doNotHandleRetries()
+    {
+        final DefaultHttpClient client =
+            (DefaultHttpClient) new HttpClientManagerImpl( hc4Provider, userAgentBuilder ).create( proxyRepository,
+                globalRemoteStorageContext );
+        // check is all set as needed: retries should be not attempted, as it is manually handled in proxy repo
+        Assert.assertTrue( ( (DefaultHttpClient) client ).getHttpRequestRetryHandler() instanceof StandardHttpRequestRetryHandler );
+        Assert.assertTrue( globalRemoteStorageContext.getRemoteConnectionSettings().getRetrievalRetryCount() != 0 );
```
peterlynch, line 140: would prefer use hamcrest or Fest Asserts

In future avoid using assertTrue/assertEquals when testing measurable conditions so that when it fails, we have a more meaningful message than something wasn't true. Using assertThat in this case could for example tell us what it was equal to or instance of.

peterlynch:
+1 thanks for tests

cstamas:
CI
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/488/

cstamas:
Not merging, CI shows some problems with content validation ITs.... will check out what happened here

cstamas:
Rebased and undid the fix. The bug is still there, even if the IT will pass. It will pass as HC4 will retry more then once (on it's own, using it's own retry handler) as 1st times it will fail. IT log will be full of NoHttpResponseExceptions.

cstamas:
Recorded issue NEXUS-5368

PR #619, opened 2012-10-25
Update core dependencies

jdillon:
sisu/guice/guava & related.

### DIFF
--- a/nexus-runtime-platform/pom.xml
+++ b/nexus-runtime-platform/pom.xml
@@ -23,6 +23,10 @@
     <relativePath />
   </parent>
 
+  <!--
+  FIXME: This does not actually provide deps on the Nexus runtime, but only on the sisu/guice/plexus and basic support libraries.
+  -->
+
   <groupId>org.sonatype.nexus</groupId>
   <artifactId>nexus-runtime-platform</artifactId>
   <name>Nexus : Runtime Platform</name>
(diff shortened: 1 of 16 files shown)


jdillon:
will link to grid build on success.

jdillon:
https://builds.sonatype.org/job/nexus-oss-its-feature/470/ lnx=SUCCESS win=FAILED 

1 failure in DefaultFSLocalRepositoryStoragePerformance2IT.testRetieveItemWithLastAccessUpdate, which I think this should sort out:

https://github.com/sonatype/nexus/pull/622

IIUC this is unrelated to the dependency changes.

peterlynch:
+1

nexus-runtime-platform/pom.xml
```diff
@@ -23,115 +23,114 @@
     <relativePath />
   </parent>
 
+  <!--
+  FIXME: This does not actually provide deps on the Nexus runtime, but only on the sisu/guice/plexus and basic support libraries.
```
peterlynch, line 27: nexus runtime means ' what common bits does nexus runtime need at runtime'

jdillon, line 27: Fair enough, though I don't think that really warrants a separate top-level pom just for that... but perhaps its needed by something in pro?

cstamas, line 27: Yes. Its used to align sisu and other deps as we had problems with that
before

Thanks,
~t~ mobile

cstamas, line 27: Forgot to add: align them between oss and pro. Just as JD said

Thanks,
~t~ mobile

jdillon:
Another win-only failure:

https://builds.sonatype.org/job/nexus-oss-its-feature/471/

org.sonatype.nexus.proxy.walker.FixedRateWalkerThrottleControllerTest.testDoesItHelpAtAll

PR #620, opened 2012-10-25
Even more cleanup

jdillon:
Probably a bit more pending, I'll update once its ready for merge.

PR #621, opened 2012-10-26
Update enunciate

jdillon:
Update to latest enunciate, adds nexus-enuciate-theme, pending configuration


jdillon:
Based on https://github.com/sonatype/nexus/pull/620 (again sorry about that)

PR #622, opened 2012-10-26
The two ITs were wrongly preparing state to be tested.

cstamas:
As "lastRetrieve" timestamp updates are not always updated, and also IT was doing a sleep to "ensure" that timestamp will differ. It does not sometimes, especially on wins. Now we set the timestamp in prep phase to some value "way in past", ensuring update will happen.


peterlynch:
+1

jdillon:
+1

PR #623, opened 2012-10-26
Fixing UT to adhere Timeline changes.

cstamas:
Two UTs from TimelineTest failed (rightfully) on Win,
and one was @Ignored. All three of them fixed:
- the two win failures were due to Lucene bump from 3.0 to 3.6.1
  that uses NIO and mmapped files, hence, locks the index files and
  is impossible to "overwrite" then behind the back of Lucene from same
  JVM process (as UT originally did: mangled files of opened lucene index)
- the ignored test was made non-ignored, but placing unsupported
  segment files, that makes index broken (as lucene refuses to read them
  and repair will happen)

All these tests are actually testing state after repair, and doing
reads, additions and purge operations on repaired index.


peterlynch:
+1

jdillon:
+1

cstamas:
Was this pull req closed intentionally?

jdillon:
I merged it... or at least I thought I merged it... wtf...

PR #629, opened 2012-10-31
NEXUS-5354: Deadlock fix

cstamas:
sync access and locks were causing a deadlock. Now, sync method
access is removed, and same is done using lock (synced access
guaranteed by exclusive lock).

Changes:
- removal of synchronized method calls (one participant of deadlock)
- redoing same method serialisation of start/stop methods using exclusive lock
- isStarted made protected, ensured is always called from already locked regions only
- extracting indexer methods, fixed the retries (before they might attempt to retry even if it was unneeded
- smaller fixes
- comments added what is being done


peterlynch:
are these changes expected to be covered by existing tests?

cstamas:
CI job:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/499/

cstamas:
@peterlynch yes, especially the MultithreadTimelineTest, that in fact somewhat "caught" (well, deadlocked) this problem.

cstamas:
Pre CI: master was broken, and failed UT was brought over here by a rebase.
New CI: https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/501/

cstamas:
Forgot to rebase
https://builds.sonatype.org/job/nexus-oss-its-feature/502/

cstamas:
CI +1'd

jdillon:
Are we waiting on anything else for this?

nabcos:
+1 for not waiting anymore ;)

jdillon:
+1 to merge it

ifedorenko:
-1

---

Timeline lucene index can be looked at as a single resource accessed by multiple threads. Threads that perform start, stop or repair operations require exclusive access. Threads that perform add, retrieve or purge operation can access the index simultaneously. Single ReentrantReadWriteLock is the most straightforward way to implement this synchronization behaviour. More complicated implementation is at least redundant but likely is also incorrect and/or deadlock prone. Proposed implementation is rather hard to review and I did not thoroughly analyze all code paths, but I believe IOException in a thread doing add/retrieve/purge operations will trigger index repair without proper guard against concurrent add/retrieve/purge requests from other threads.

---

Access to DefaultTimelinePersistor rollingIntervalMillis, persistDirectory, lastRolledTimestamp and lastRolledFile members is not properly synchronized. A thread setting these members through (non-synchronized) doStart can run concurrently with getDataFile and readAllSinceDays and synchronized persist methods.

---

Specific recommendations
- Remove DefaultTimeline.repairLock and rework implementation to use timelineLock.writeLock around open, close and repair operations and timelineLock.readLock around add, retrieve and purge. Make sure that all code paths that touch indexer are guarded by timelineLock. Remove isIndexerHealthy as part of this rework. 
- Remove AbstractStartable and AbstractLockingStartable abstract classes. With only one meaningful implementation -- DefaultTimeline -- they only complicate the code, making it harder to understand.
- Synchronize all code paths that access DefaultTimelinePersistor instance members.
- Although not strictly related to locking, repair-and-retry logic during add, retreive and purge operations seems unnecessary yet adds significant complexity. The only case I can think of it is useful is when the lucene index files are deleted on filesystem... and I am not sure if rebuilding the index is expected/desired behaviour in this case nor if the current implementation is actually working. I think it is enough to simply stop processing new add/retrieve/purge requests after IOException but do not attempt recover from it in any way, not even close the timeline.

jdillon:
thx Igor... btw, wtf does timeline do actually.

If its all RSS shit, and that stuff is all noted to be crap anyways and likely to kill NX instances, why don't we just kill this stuff?

cstamas:
@ifedorenko your analysis is pretty much correct, but there has to be added few explanations and/or remarks.

timeline is actually "container" for resources: persistor and indexer, and is accessed by multiple threads.

persistor -- that performs only write operations -- should be synchronised, otherwise it would concurrently write to same file as you noted in some of the previous comments of yours.
indexer -- that performs read/write (retrieve/add+purge) operations -- should NOT be synchronised, since Lucene itself doing this is well optimised for concurrent access.

On the other hand, operations like "start", "stop" should be done in exclusive access.

But, problem is repair, as plain exclusive access is a no go. Think following scenario: multiple threads doing deploys (hence, events will be emitted that Timeline will want to process), but your index got corrupted. Probably all the threads trying to add to Lucene will get IOEx, and all of them will try -- or better, stand into the queue -- to repair the index.

With plain exclusive lock, when 1st thread tries and succeeds to fix, the next blocked thread will _again_ repair it, and so forth, with all the queued repairing threads.

# 

If not repairing index at all is an option, i think that would be the simplest, as the repair step is that one that adds great deal of complexity to this code. Without repair, yes, Timeline becomes trivial: simply, when indexer dies, it will not be used anymore....

So, is repair a requirement?

cstamas:
@jdillon Timeline is a component that records entries in one dimension (time) and exposes them and makes them searchable.

Right now, there is only one "consumer" of Timeline, and yes, those are RSS feeds, but it can be used for other use cases too, when you need to store/search series of timestamped data. Timeline itself is not RSS specific at all, it deals with "records", and the conversion happens on top of it, for RSS case, in FeedRecorder/Reader

ifedorenko:
@cstamas like I said, timeline repair complicates the code but does not affect number of locks. multiple locks are not required to implement "don't repair the same problem twice" logic, all what's needed is a way to detect "same problem" reliably and this does not require additional locks, simple restart counter in DefaultTimelineIndexer should do it. 

Btw, current proposed implementation makes redundant repair somewhat less likely but does not guarantee it will never happen. Nothing stops multiple threads concurrently perform isIndexerHealthy check, then go about doing their business, then get IOException, then each do repair.

ifedorenko:
MUCH better, but still -1

Read lock is unlocked twice in case of IOException during repait

DefaultTimelinePersistor member access synchronization is still broken.

Few other minor comments
- "if ( !started )" checks everywhere are redundant. Enough to check in doTryRepair only.
- DefaultTimelinePersistor.start is really setConfiguration
- DefaultTimelineIndexer generation is already protected by lock in DefaultTimeline, no need to use AtomicInteger
- DefaultTimelineIndexer.close() calls everywhere are redundant, the indexer will be closed/opened as part of repair

ifedorenko:
+1

Merge it. I am tired of this back and forth.

PR #631, opened 2012-11-01
Drop nexus-plugins module group

jdillon:
Drop the nexus/nexus-plugins module, move contents up under nexus/*

This module didn't change the groupId or add any additional configuration for it children, so dropping it.


adreghiciu:
+1

PR #633, opened 2012-11-02
[NEXUS-5361] Remove CheckSum browse + search applet

jdillon:
https://issues.sonatype.org/browse/NEXUS-5361


jdillon:
Not really anything that can be verified by a its build... I don't believe there are any specific its for this.

cstamas:
+1

peterlynch:
+1

jdillon:
I guess we want to keep the checksum search, but just kill the digestapplet (which wasn't my original understanding... but oh well).

nabcos:
+1

PR #634, opened 2012-11-02
Use maven-shade-plugin

jdillon:
Replaces maven-assembly-plugin generation of -cli.jar executables with maven-shade-plugin.


peterlynch:
no tests that I could see, I guess you made sure it still works +1

jdillon:
It runs, that all I tested... I sent mail to the list for more details on how these things are supposed to work.

jdillon:
Gonna merge this for reference, then drop these modules.

PR #642, opened 2012-11-06
Removing the remnants of fasterXML too.

cstamas:
Original change left JacksonXMLMarshaller in place, that is unused
and became equiv of the JacksonJSONMarshaller.

Also, as JD spotted, there was an intention to "pretty print"
the JSON, but it was never properly implemented. Code fixed,
but commented out, as am unsure do we really want to fiddle with it
now (only updated attrs would be pretty printed with change enabled,
while this does not interfere with marshalling, it might introduce
perf penalty i am unsure about)

Cleans up the changes made in
https://github.com/sonatype/nexus/pull/626


adreghiciu:
+1

jdillon:
+1

PR #644, opened 2012-11-07
Automatically handle Jersey ClientResponse

adreghiciu:
Avoid as much as possible UniformInterfaceException and:
- In case of 404 throw an NotFoundException
- In case of 400 with an ErrorResponse throw a specific exception

### DIFF
--- a/nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
+++ b/nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
@@ -1,0 +1,68 @@
+/**
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.client.internal.rest.jersey;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Collections;
+import java.util.List;
+
+import org.sonatype.nexus.client.internal.msg.ErrorMessage;
+import org.sonatype.nexus.client.internal.msg.ErrorResponse;
+
+/**
+ * @since 2.3
+ */
+public class BadRequestException
+    extends RuntimeException
+{
+
+    private final List<ErrorMessage> errors;
+
+    public BadRequestException( final ErrorResponse response )
+    {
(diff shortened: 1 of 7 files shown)


nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
```diff
@@ -0,0 +1,68 @@
...
+        if ( errors != null && !errors.isEmpty() )
+        {
+            final StringBuilder sb = new StringBuilder();
+            for ( final ErrorMessage error : errors )
+            {
+                if ( sb.length() > 0 )
+                {
+                    sb.append( "\n" );
+                }
+                sb.append( error.getMsg() );
```
nabcos: Shouldn't this also mention id when the ID is not "*"?

adreghiciu:
Without this every subsystem impl have to handle this by their own.
More, in most of current subsystem impls we do not handle and that mens that responses remain unconsumed which results in jersey exceptions

nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+
+import org.sonatype.nexus.client.internal.msg.ErrorMessage;
+import org.sonatype.nexus.client.internal.msg.ErrorResponse;
+
+/**
+ * @since 2.3
+ */
+public class BadRequestException
+    extends RuntimeException
```
cstamas: NexusClientException ?

nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/NotFoundException.java
```diff
@@ -0,0 +1,22 @@
...
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.client.internal.rest.jersey;
+
+/**
+ * @since 2.3
+ */
+public class NotFoundException
+    extends RuntimeException
```
cstamas: NexusClientException ?

cstamas:
Consider using NexusClientException as top level class for exceptions (is also runtime ex).

nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
adreghiciu: +0, how you guys want it. Something like [id] message ?

adreghiciu:
@cstamas c726fa653c850b258aa35d6395834ae1727bcd4e

cstamas:
:kiss:

nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
nabcos: Not sure how generic the message usually is, and I think adding [id] would not hurt.

nabcos:
+1

nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/internal/rest/jersey/BadRequestException.java
adreghiciu: Added id in 4b1af7d39f3ad67409218da5e6a146e256b8d692

PR #649, opened 2012-11-08
Simplify plugin deps

jdillon:
Make use of nexus-plugin-api and nexus-plugin-test-api to bring in most deps as provided that plugins need.


cstamas:
+1

nabcos:
+1

PR #651, opened 2012-11-08
Squash the build.

cstamas:
After recent changes (and near future ones, like the introduction of restlet1x plugin),
there is really no more reason to keep complicated module layout.

Changes:
- modules removed: nexus-utils, nexus-api, nexus-configuration, nexus-proxy, nexus-app, nexus-plugin-manager
- module "cleaned": nexus-configuration-model, now holds only modello models. Remains to be reusable (for some CLI tools)
- module nexus-core introduced, that is actually The Core of Nexus

No source changes happened, just move around. POMs were updated, removed all
occurrences of removed modules, introduced new module.


adreghiciu:
+1 I hope that now we also get rid of duplicated packages in different modules. can you ensure that?

cstamas:
Closing this pull, will be redone again by me, once master settles.

PR #653, opened 2012-11-09
Archetype update

mosabua:
Small comment to add grid repo and updated compiler plugin (used ancient version before).

### DIFF
--- a/nexus-plugin-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/nexus-plugin-archetype/src/main/resources/archetype-resources/pom.xml
@@ -21,6 +21,11 @@
   <dependencies>
 
     <!--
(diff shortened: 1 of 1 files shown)


cstamas:
:beer:

peterlynch:
I missed why central is not enough?

mosabua:
This is what you get with central only.. https://gist.github.com/4048052

peterlynch:
ok as long as https://builds.sonatype.org/view/nexus/job/nexus-oss-its-external-developer/ eventually passes, which mimics what an external developer will do

peterlynch:
+1

mosabua:
Well... the change did not affect that and the build is currently broken anyway :-(

PR #655, opened 2012-11-10
Use HC4Provider for PR attachments

nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/512/


nexus/nexus-app/src/main/java/org/sonatype/nexus/error/reporting/NexusPRConnector.java
```diff
@@ -12,291 +12,44 @@
...
     extends HttpClientConnector
-    implements Disposable
 {
 
-    private static final Logger logger = LoggerFactory.getLogger( NexusPRConnector.class );
-    
-    private NexusConfiguration config;
-
-    private final UserAgentBuilder uaBuilder;
+    private final org.sonatype.nexus.apachehttpclient.Hc4Provider httpClientProvider;
```
jdillon: odd FQCN here given that its also imported.

jdillon:
+1

jdillon:
BTW, the Nexus2862UrlRealmIT failures are know to be due to the restlet1x change... so I think we should merge this.

PR #656, opened 2012-11-12
Log & send event about failing tasks also in case of Errors

adreghiciu:
This was missing cases like ClassNotFoundException which could happen for example in a plugin that was depending on some dependency of core that changed


cstamas:
+1

nabcos:
+1

PR #658, opened 2012-11-13
NEXUS-5372: Alternative fix

cstamas:
Introduced new format, to replace the old broken
format, while keeping backward compat. In cases when
records are no longer than 255 bytes at least.

As in real life, records are longer than 255, this does not "salvage" the historical data (that is already written out corrupted), but at least does not break the build. Also, it uses proper Protobuf API to write out delimited records, is not rolling it's own to do so.


ifedorenko:
-1

Lets remove V1 and V2 formats. V2 was broken and there is no point in providing support for it. V1 was before V2, so any historic timelines customers might have are broken.

Please provide a unit test for records longer than 255 bytes.

cstamas:
Applied Igor's comments.

nexus/plugins/nexus-timeline-plugin/src/main/java/org/sonatype/timeline/internal/DefaultTimelinePersistor.java
```diff
@@ -106,14 +100,12 @@ protected synchronized void persist( final TimelineRecord... records )
         OutputStream out = null;
         try
         {
+            out = new FileOutputStream( getDataFile(), true );
```
ifedorenko: I'd use BufferedOutputStream here.

nexus/plugins/nexus-timeline-plugin/src/main/java/org/sonatype/timeline/internal/DefaultTimelinePersistor.java
```diff
@@ -262,12 +254,12 @@ else if ( filePtr >= result.size() )
         try
         {
             in = new FileInputStream( file );
```
ifedorenko: I'd use BufferedInputStream here

ifedorenko:
+1 with the comments above

nabcos:
+1

PR #659, opened 2012-11-13
Use timestamp as path prefix for aggregated js file

nabcos:
This is WIP and not working correctly right now, more a reminder for future work and RFC whether someone spots what's wrong here. The reason for this is better cache busting in browsers.

For some reason the ExternalStaticResources is not attached to restlet correctly. It's processed in [NexusApplication](https://github.com/sonatype/nexus/blob/timestamped-resource-url/nexus/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/NexusApplication.java#L210), but still gives 404 for the timestamped URLs.

This is a hack anyway because it uses an ExternalStaticResource for sonatype-all.js. It should rather use classpath lookup, but that seems to not work at all for the location of the js files here (they are included in the WAR root directory). This will get better when we can wrap the UI stuff into a plugin.

### DIFF
--- a/nexus/nexus-webapp/pom.xml
+++ b/nexus/nexus-webapp/pom.xml
@@ -261,7 +261,7 @@
               <aggregations>
                 <aggregation>
                   <insertNewLine>true</insertNewLine>
-                  <output>${compressed-dir}/js/sonatype-all.js</output>
+                  <output>${project.build.outputDirectory}/js/sonatype-all.js</output>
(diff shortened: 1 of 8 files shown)


jdillon:
Why not move the core js muck this needs to load in to the WEB-INF/classes directory?

nabcos:
... Because it's too obvious? I'll try that tomorrow.

nabcos:
Changes squashed into ui-extjs3 branch.

PR #660, opened 2012-11-14
Get rid of Timeline runtime repair

cstamas:
Repair should be tried only on "boot". During runtime, if corruption is detected, the Indexer should be simply shot down, marked as "broken" and no further operation tried against it.

On next boot, it will be repaired (or tried at least), but no data loss occurs, as the persistor will still write the records.

### DIFF
--- a/nexus/plugins/nexus-timeline-plugin/src/main/java/org/sonatype/timeline/internal/DefaultTimeline.java
+++ b/nexus/plugins/nexus-timeline-plugin/src/main/java/org/sonatype/timeline/internal/DefaultTimeline.java
@@ -116,17 +116,12 @@ public class DefaultTimeline
                             "Timeline index is succesfully repaired, the last "
                                 + configuration.getRepairDaysCountRestored() + " days were restored." );
                     }
-                    catch ( IOException ex )
-                    {
-                        markIndexerDead( ex );
-                        throw ex;
-                    }
                     catch ( Exception ex )
                     {
+                        // do not propagate the exception for indexer
+                        // we have persistor started, and that's enough
                         markIndexerDead( ex );
-                        throw new IOException( "Failed to repair indexer!", ex );
                     }
-
                 }
                 DefaultTimeline.this.started = true;
                 getLogger().info( "Started Timeline..." );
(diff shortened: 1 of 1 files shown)


nexus/plugins/nexus-timeline-plugin/src/main/java/org/sonatype/timeline/internal/DefaultTimeline.java
```diff
@@ -247,20 +274,18 @@ E doIt()
 
     }
 
-    protected <E> E doTryRepair( final Work<E> work )
+    protected <E> E doShared( final Work<E> work )
     {
         if ( timelineLock.readLock().tryLock() )
         {
             if ( started && !indexerIsDead )
```
ifedorenko: Lock is not unlocked if  "started && !indexerIsDead" evaluates to false.

ifedorenko:
-1

doShared is both incorrect (see the comment above) and is generally too complicated. IOException during work.doIt() most likely means something really bad, like filesystem corruption or too many open files. Get exclusive lock and do markIndexerDead does not add much, something like this should be sufficient

```
protected <E> E doShared( final Work<E> work )
{
    if ( timelineLock.readLock().tryLock() )
    {
        try
        {
            if ( started && !indexerIsDead )
            {
                try
                {
                    return work.doIt();
                }
                catch ( IOException e )
                {
                    indexerIsDead = true;
                    getLogger().warn( "Timeline index got corrupted and is disabled. Repair will be tried on next boot.",
                                      e );
                }
            }
        }
        finally
        {
            timelineLock.readLock().unlock();
        }
    }
    return null;
}
```

This will also allow simpler and cleaner start() code and get rid of doExclusively.

Nitpicking, remove unused private TimelineConfiguration configuration member.

jdillon:
How about this:

```
rm -rf nexus/plugins/nexus-timeline-plugin
```

cstamas:
Applied Igor's comments. Jason's comment not taken seriously yet.

ifedorenko:
Nitpicking. "if !started" checks everywhere are redundant, the check in doShared is enough. Otherwise +1, pending outcome of the discussion if nexus should or should not start if timeline cannot be opened.

adreghiciu:
+1

cstamas:
CI liked this too (linux slave is blue, did not wait for win slave):
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/521/

PR #661, opened 2012-11-14
Squash-o-matic output

cstamas:
Squashes the Nexus Core modules into one "nexus-core" module. No Core code change, just shuffling them around.

Changes:
- nexus-core module introduced, with all the sources from following modules: nexus-utils, nexus-api, nexus-configuration, nexus-proxy, nexus-app and nexus-plugin-manager
- removed modules (well, "moved" them into):
- cleanup in UTs and ITs, mostly dependencies but some changes were done too, as suddenly the "mocked" environment became functional (many UTs from api, config or proxy were written by assuming Nexus Core is not functional there, as key code is added in app module).

CI went fine:
https://builds.sonatype.org/view/nexus-features/job/nexus-oss-its-feature/517/


cstamas:
I am +1, but this will introduce _another_ wave of CI builds being broken. So, it should be decided "when" we do this.

jdillon:
I defs think we should ensure we have stabilized the current set of changes first and have functional nexus-oss and nexus-professional + nexus-pro-trial-installer distributions.

IIUC this change should probably be less of a pain, since last round of changes were made to have plugins depend on nexus-plugin-api and not any of the other nexus-proxy, nexus-app, etc modules.

cstamas:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF1-2

cstamas:
CI started
https://bamboo.zion.sonatype.com/browse/NXO-OSSF1-5

cstamas:
CI "passed": one search related IT failed

Related issue for tracking
https://issues.sonatype.org/browse/NEXUS-5470

PR #662, opened 2012-11-14
Implement maven artefact upload/delete (parts copied from yum plugin)

adreghiciu:


nexus/nexus-client-core/src/main/java/org/sonatype/nexus/client/core/subsystem/artifact/UploadWithCoordinatesRequest.java
```diff
@@ -0,0 +1,156 @@
...
+    private final String extension;
+
+    private final File file;
+
+    private final File pomFile;
+
+    public UploadWithCoordinatesRequest( String repositoryId, String groupId, String artifactId, String version,
+                                         String packaging,
+                                         String classifier,
+                                         String extension, File file )
```
peterlynch, line 46: is this really how our formatting rules format files?

adreghiciu:
After discussion with Tamas we decided that this kind of client is not encouraged. One should use the conntent upload client

PR #663, opened 2012-11-14
NEXUS-5250 fixed empty GUI keyword search results

ifedorenko:
This was a regression introduced during maven-indexer 5.0.0
rework. SearchNGIndexPlexusResource was not correctly falling
back to uncollapsed response when query returned small number
of records.


cstamas:
+1

PR #665, opened 2012-11-15
Stabilize IT: Add 'cancellable' parameter to SleepRepositoryTask

nabcos:
dba2d76ad065a741cd4f664e92d00232ba0405b1 changed the basic assumption for the test (SleepRepositoryTask is not cancellable). When the task checks for interruptions, the 'CANCELLING' state may be over too fast for the IT to pick it up.


nabcos:
https://builds.sonatype.org/job/nexus-oss-its-feature/518/

PR #666, opened 2012-11-15
[NEXUS-3728] use 'su -' intead of 'su -m'

jdillon:
https://issues.sonatype.org/browse/NEXUS-3728


nabcos:
Not sure I followed the script correctly, but IIUC you could set variables in your shell
(PIDDIR) that are not used anymore when RUN_AS_USER is set, because they
will not be preserved without -m.

That said, I still think 'su -' is better -> +1

adreghiciu:
+1

PR #669, opened 2012-11-21
Use the reference factory to get the baseUrl

jdillon:
https://issues.sonatype.org/browse/NXCM-4804


cstamas:
So, was it reference factory or non-relative URL? Or both?

Anyway, both changes looks fine: :beer: :beer:

nabcos:
It was the fact that $nexusRoot was not giving the right URL, and we (unnecessary) had the absolute URL in the HTML.

jdillon:
@nabcos I was wondering why the full URL was used here ;-)

PR #670, opened 2012-11-22
NEXUS-5368: Stale connections in pool

cstamas:
Explicitly controlling the duration for how long
is connection kept alive (it's total lifespan).

Some servers announce their "wish" for how long to
keep connection alive with non standard "Keep-Alive"
response header, and HC4 obeys that header. But some
servers does not do this (like RSO) and are silently
dropping the connection, while HC4 in this case
sets expiration time of the connection to "indefinite"
(forever).

With this change, we take over explicit control over
keep alive duration, and we prevent HC4 to pool connections
forever.

Note: this change also _renames_ one badly named parameter, as that one had no direct connection with real Keep-Alive support. It was about "for how long we should keep connection in pool idle".


cstamas:
This change does not fix the problem completely (originally, problem is remote end sneakily closing connection, not sending FIN packet to us, nor telling us they will close their end of socket in the future), but will likely lessen the garbage in logs, lines like these:

```
jvm 1    | 2012-11-22 15:38:39 INFO  [qtp151804362-32] - org.sonatype.nexus.apachehttpclient.Hc4ProviderImpl$DefaultHttpClientImpl - I/O exception (org.apache.http.NoHttpResponseException) caught when processing request: The target server failed to respond
jvm 1    | 2012-11-22 15:38:39 INFO  [qtp151804362-32] - org.sonatype.nexus.apachehttpclient.Hc4ProviderImpl$DefaultHttpClientImpl - Retrying request
```

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderImpl.java
```diff
@@ -158,36 +170,38 @@
...
-        this.evictingThread = new EvictingThread( sharedConnectionManager, getConnectionPoolKeepalive() );
+        this.evictingThread = new EvictingThread( sharedConnectionManager, getConnectionPoolIdleTime() );
         this.evictingThread.start();
         this.eventBus = Preconditions.checkNotNull( eventBus );
         this.eventBus.register( this );
         this.jmxInstaller.register( sharedConnectionManager );
-        getLogger().info( "{} started up (keep-alive {} millis), listening for shutdown.", getClass().getSimpleName(),
-            getConnectionPoolKeepalive() );
+        getLogger().info(
+            "{} started up (connectionPoolMaxSize {}, connectionPoolSize {}, connectionPoolIdleTime {} ms, connectionPoolTimeout {} ms, keepAliveMaxDuration {} ms)",
```
peterlynch: I would change 'started up' to 'started'

peterlynch: unless we have concept of 'started down'

peterlynch:
not unit test but otherwise +1

PR #671, opened 2012-11-22
NEXUS-5301: Adding logback level change propagator

cstamas:


peterlynch:
your xml appears well formed :+1: 

have no opinion on the rest

PR #672, opened 2012-11-22
NEXUS-3442: Bumped restlet

cstamas:
For changes see:
https://github.com/sonatype/restlet1x/pull/1

### DIFF
--- a/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
+++ b/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
@@ -35,7 +35,9 @@
   <logger name="org.sonatype.nexus.rest.NexusApplication" level="ERROR" />
   <logger name="httpclient" level="INFO" />
   <logger name="org.apache.http" level="INFO" />
-  <logger name="org.restlet" level="WARN" />
+  <logger name="org.restlet" level="INFO" />
+  <!-- To mute Restlet's Access Logging LogService -->
+  <logger name="org.restlet.Component.LogService" level="WARN" />
   <logger name="org.apache.commons" level="WARN" />
   <logger name="org.apache.shiro.web.filter.authc.BasicHttpAuthenticationFilter" level="INFO" />
   <logger name="org.apache.shiro.web.filter.mgt.DefaultFilterChainManager" level="INFO"/>
--- a/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
+++ b/nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
@@ -36,8 +36,6 @@
   <logger name="httpclient" level="INFO" />
   <logger name="org.apache.http" level="INFO" />
   <logger name="org.restlet" level="INFO" />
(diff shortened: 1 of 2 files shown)


cstamas:
One problem with this pull already identified. Now that the "flow" is repaired (Restlet again detects properly "client closed connection early" case, and does not retry to modify already committed response), it does log it on INFO level.

Nexus ships with Restlet log level set to WARN, but I think it's a historical, and exactly the reason was this problem fixed in this pull. Or maybe not. Peter?

In short, the new message is not printed in log, as it has INFO level, while Restlet is "muted" to WARN level in log back-nexus.xml shipped with Nexus.

cstamas:
Realised the problem Resltet 1.1 suffers from LogService problem, that produces access log from Restlet, and Nexus logs are filled with them when setting logger "org.restlet" to INFO level. By inspecting code (and remembering), I realised that Resltet 1.1 _gives no means_ to shut that damn service off.

Proposal: Add this to logback-nexus.xml as seems more appropriate

  &lt;logger name="org.restlet" level="INFO" /&gt;
  &lt;logger name="org.restlet.Component.LogService" level="WARN" /&gt;

Hence, _mute_ the damn LogService thing.

peterlynch:
+1

cstamas:
I am about to release V6 patch instead, with following changes:
https://github.com/sonatype/restlet1x/pull/2
https://github.com/sonatype/restlet1x/pull/3

Later will completely disable LogService, so no need for 2nd commit in this pull and Log level trickery, but will also save us few CPU cycles.

Once V6 out, will update this pull (remove logback changes, just setting Restlet to INFO and bump V5 with V6)

nexus/nexus-logging-extras/src/main/resources/META-INF/log/logback-nexus.xml
```diff
@@ -35,7 +35,9 @@
   <logger name="org.sonatype.nexus.rest.NexusApplication" level="ERROR" />
   <logger name="httpclient" level="INFO" />
   <logger name="org.apache.http" level="INFO" />
-  <logger name="org.restlet" level="WARN" />
+  <logger name="org.restlet" level="INFO" />
+  <!-- To mute Restlet's Access Logging LogService -->
+  <logger name="org.restlet.Component.LogService" level="WARN" />
```
nabcos: Is that logger name correct?
And didn't we turn off the LogService in sonatype/restlet1x#3? So muting it here will do nothing.

cstamas: This commit is part of pull https://github.com/sonatype/nexus/pull/672 that was prepped yesterday for V5.
Today morning I made two new pull requests on https://github.com/sonatype/restlet1x and preparing V6.

So yes, with V6, this pull will need update. See my comment on
https://github.com/sonatype/nexus/pull/672#issuecomment-10652981

cstamas:
Once V6 out, this patch is now reworked to apply it. No more log level trickery, Restlet runs on INFO level.

V6 contains V5 + RemoteIPFinder fix + LogService shut down (this means, LogFilter is not installed either, so CPU cycles saved)

PR #674, opened 2012-11-23
Implementing size() method.

cstamas:
Default implementation was returning -1 constant for
getSize(), that made Jersey to stick with Chunked
transfer encoding (did not know exact size of http message body).

Some HTTP Servers does not like it, like nginx.

This is applied to _all_ client subsystems.

Fixes:
https://github.com/yegor256/jcabi/issues/47


cstamas:
Um, I meant to say, fixes:
https://github.com/sonatype/nexus-maven-plugins/issues/13

cstamas:
writeTo and getSize are invoked at different times, and provider should be stateless, while Jersey API did not give me any "context" (like some place to stuff resulting XML payload and reuse the same in other method call).

so, yes, it is marshalled twice, as there is (am unaware) of any other solution to this, and to keep provider stateless.

PR #676, opened 2012-11-26
[NEXUS-4864,NEXUS-5169] Add better mimetype detection/validation configuration

nabcos:
With this change, Nexus will look up '/nexus.mimetypes' from classpath.
This property file must have the following format:
- Configure additional mimetypes:
  A mapping from file name extension to a comma separated list of mime types (e.g. `bz2: application/x-bzip2,application/x-bzip`). No whitespace is allowed in the list of mime types. Mime types listed in this file will be used for content validation.
- Override mimetypes:
  Same as adding mime types but prefixed with 'override.' (e.g.  `override.html: text/xhtml,application/xml`). These mime type definitions will override the builtin mime types. The first listed mimetype is the 'primary' mime type and will be used by Nexus as the downstream content type.


nabcos:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF11-5

nexus/nexus-api/src/main/resources/nexus-builtin.mimetypes
```diff
@@ -0,0 +1,13 @@
...
+ear = application/zip
+pom = application/xml,application/x-maven-pom,text/xml
+xml = application/xml,text/xml
+tar = application/x-tar
+swc = application/zip
+swf = application/x-shockwave-flash
+gz = application/x-gzip,application/x-tgz
+tgz = application/x-tgz,application/x-gzip
+bz2 = application/x-bzip2
+tbz = application/x-bzip2
```
jdillon: Any reason this isn't named *.properties?  Then license muck can figure out how to put a header on this?

jdillon:
+1

nexus/nexus-api/src/main/resources/nexus-builtin.mimetypes
nabcos: no real reason... We already had mime-types.properties in use for MimeUtils2 configuration, so I chose something different to not have a user confuse mime-type.properties and nexus.mimetypes. That's no reason for the builtin types, though. I'll change that file.

cstamas:
:beer:

adreghiciu:
+1

nexus/nexus-api/src/main/resources/nexus-builtin.mimetypes
jdillon: Why are there 2 files to configure the mime times?  I'm confused...

nabcos: mime-types.properties is for the MimeUtil2 lib. I do not want to change that, if we configure those mimetypes 'the new way' they would override the MimeUtil2 behavior and break customer installations that had to supply their own mime-types.properties.

PR #678, opened 2012-11-27
NEXUS-5412 Return of the retry.

cstamas:
Change 5ddf6a171337ddf8e1a881e63d1ab9ec4a7042ba removed too much.
There was also buried in deep the "retry if result set is small"
case.


ifedorenko:
-1

You don't need for loop that executes 1 or two times. move loop body to a method and call the method second time if needed.

cstamas:
Applied Igor's comment: body moved out to a method, called 2nd time if needed.

cstamas:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF16-1

cstamas:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF16-2 passed

ifedorenko:
I still need to review the code, but this change does not go in without corresponding regression test.

cstamas:
Updated existing UT, that was brain dead.

ifedorenko:
-1  the unit test does not test anything relevant it seems, i.e. it passes with or without main code changes.

Also, the code appears way too complicated for what it is supposed to do, it is almost impossible to follow and review properly in its current form. 

I've opened https://github.com/sonatype/nexus/pull/687 to illustrate possible simplified implementation. It collects and sorts all NexusNGArtifact instances with single straightforward for-each loop, then returns either collapsed or expanded SearchNGResponse. The simplified code may even perform better on average because it does not need to perform second index query to "uncollapse" response, but code readability if my main concern here.

cstamas:
Proper UT covering this case added in 7532a241fda494fe9e7cc2500a436fc5886d771d (master branch by mistake).
After merging origin/master here, confirmed that test (that fails on master) passes on this branch: bug is fixed

PR #681, opened 2012-11-28
[NEXUS-5399] Expose "Item Max Age" on UI / via REST

adreghiciu:
The new resource field is an optional Integer to avoid existent code to unwanted override of item max age with zero. In that case existent default value will be kept.


cstamas:
+1

adreghiciu:
CI :+1: https://bamboo.zion.sonatype.com/browse/NXO-OSSF19-2

nexus/nexus-webapp/src/main/webapp/js/repoServer/resources.js
```diff
@@ -153,6 +153,7 @@ define('repoServer/resources',['sonatype/all'], function(){
       notFoundCacheTTL : 'This controls how long to cache the fact that a file was not found in the repository.',
       artifactMaxAge : 'This controls how long to cache the artifacts in the repository before rechecking the remote repository. In a release repository, this value should be -1 (infinite) as release artifacts shouldn\'t change.',
       metadataMaxAge : 'This controls how long to cache the metadata in the repository before rechecking the remote repository. Unlike artifact max age, this value should not be infinite or Maven won\'t discover new artifact releases.',
+      itemMaxAge : 'This controls how long to cache the items in the repository before rechecking the remote repository. The value is used for repository items which are nor artifacts nor metadata.',
```
nabcos, line 156: shouldn't this be 'neither ... nor'?

nabcos:
+1

nexus/nexus-webapp/src/main/webapp/js/repoServer/resources.js
adreghiciu, line 156: I think the native english speaking guys should change it

PR #682, opened 2012-11-29
NEXUS-5400: Uploaded file trashed by rebuild md

cstamas:
Since Nexus 2.0 a requirement for repositories is that
from their root ("baseDir"), they must reside on _same volume_
(due to attributes). This change that builds upon that,
and introduces repo "tmp" folder. With this change, we do
not put "tmp" files anymore in the folder where they will
reside, but instead into a "/.nexus/tmp" folder under
same hierarchy (parent folders). Due to "one repo on one volume"
requirement, rename will be still cheap, and not cp+rm operation.
Hence, it does not matter is file in same folder (as before)
or other folder, as long rename happens on single volume.

This still does not resolve other problem revealed: locking.
Rebuild md deletes empty folders (more precisely, what it
"thinks" is empty, see Ben's comment on this issue) without
doing any locking, using the "unprotected" API. But, this was
like it since day 0 it seems.


cstamas:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-1

nabcos:
+1

ifedorenko:
Why not teach walker how to honour .nx-upload files?

cstamas:
Problem is that FSPeer was hiding these entries from caller. If we let those "leak", we would not only need to teach walker, but also many other components to honour those.
As I mentioned on scrum, this is still WIP and is not the change I am happiest with, but is least obtrusive...

adreghiciu:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-3

adreghiciu:
Test for NEXUS-5400?

adreghiciu:
CI :-1:

cstamas:
Yes, this needs some love, since empty folders are not cleaned up properly, or that one failing IT is too eager on empty folders found in storage and needs an exclude (on where to check for empty folders).

cstamas:
CI :+1: 
https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-5

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
```diff
@@ -443,7 +471,7 @@ protected void handleRenameOperation( File hiddenTarget, File target )
             if ( success )
             {
                 getLogger().info( "Rename operation succeeded after {} retries on {} --> {}",
-                    new Object[] { i, hiddenTarget.getAbsolutePath(), target.getAbsolutePath() } );
+                                  new Object[]{ i, hiddenTarget.getAbsolutePath(), target.getAbsolutePath() } );
```
adreghiciu, line 474: With slf4j api version we use now you do not need anymore the new Object[]

adreghiciu:
:+1: 
I still do not see an IT specifically testing uploading vs rebuild MD task, yet that kind of test sounds not easy to do. More, now given that temp files are in a different place, rebuild MD should not touch them

nexus/nexus-proxy/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
cstamas, line 474: @adreghiciu that code was changed by formatter, not by me. I did not want to touch anything more than needed in here (but reformat slipped in).

cstamas:
Yes, I did not cover this case with IT as it would be a bit involved to create one. Probably will need some manual testing.

PR #683, opened 2012-11-30
ide workspace dependency resolution for nexus plugins

ifedorenko:
This pull request enables support for m2e workspace dependency resolution for nexus plugins.

### DIFF
--- a/nexus/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/repository/UserNexusPluginRepository.java
+++ b/nexus/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/repository/UserNexusPluginRepository.java
@@ -51,7 +51,7 @@ final class UserNexusPluginRepository
 
     public int getPriority()
     {
-        return 50;
+        return 100;
     }
 
     // ----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


peterlynch:
fine

nexus/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/repository/UserNexusPluginRepository.java
```diff
@@ -51,7 +51,7 @@ public String getId()
 
     public int getPriority()
     {
-        return 50;
+        return 100;
```
jdillon, line 54: why?  can priority not be negative in the case?  so that the custom xml muck can be -50 and have a higher priority?

jdillon:
This is a hack.  To be really useful will require more tooling work and support to really be useful, asis now the plugin dev-env can quickly get out of sync with the real env configuration.   It will end up with a different runtime environment that what the server would normally operate under.  I think this is not the right direction, though as a short-term-limited hack is fine.  I think a better solution would be replace the plugin manager with a dynamic module system and the implement fast build/redeploy of real plugins into the server.

That said, I'm +0 on this change... though if this is intended or will become (even if not intended) as the primary means for developing NX plugins, them I'm -1.

nexus/nexus-plugin-manager/src/main/java/org/sonatype/nexus/plugins/repository/UserNexusPluginRepository.java
ifedorenko, line 54: No good reason. I just can't think in negative priorities I guess.

PR #685, opened 2012-11-30
[NEXUS-5258] Do not save default remote storage provider

nabcos:
This pull request will change the config behavior to not save the system default remote storage provider into the repo configuration. 

Repositories with old 'apachehttpclient3x' will stay that way. Repositories using 'apachehttpclient4x' (the current default) will be saved without a provider hint to grab the system default on load in the future. This value can already be set via a system property. This makes "upgrading" the default storage provider effortless, as all repositories using the default before will instantly use the new default.

In hindsight, I'm not sure if that was the intention of the issue. As it is now, a new repository will have the current system default applied, and will explicitely save that value into the configuration. This behavior actually matches the issue description ("use system default on create").

Comments?


cstamas:
:beer:

adreghiciu:
Unit tests? like 
- create a repo, save and check is not saved if default
- create a repo with a non default provider, save and check is set

Feature build?

nabcos:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF21-3

adreghiciu:
:+1:

PR #690, opened 2012-12-04
NXCM-4814 avoid system-wide indexer lockups

ifedorenko:
All indexer locks are acquired with a time out (default 60 seconds).
Lock timeouts are logged as errors but otherwise ignored. This can
result in individual index corruption and index-related exceptions,
particularly on windows, but should prevent system-wide lockups due
to infinite wait for lock acquisition.


ifedorenko:
I merged&pushed this to master http://blogs.msdn.com/b/seliot/archive/2011/04/25/i-don-t-always-test-my-code-but-when-i-do-i-do-it-in-production.aspx , but I will still appreciate code review of this change.

PR #691, opened 2012-12-04
Bumped MI version

cstamas:
Using latest MI Internal release


ifedorenko:
Please merge. There is no need to review trivial changes like this.

PR #692, opened 2012-12-05
[NXCM-4587] Unescape HTML entities for SMTP valiation resource

nabcos:
The smtp config validation resource actually uses a different DTO than the actual settings resource... User/Password is unescaped for both now.


cstamas:
+1

peterlynch:
+1

PR #693, opened 2012-12-06
NEXUS-5422 use String to determine if two versions are equal

ifedorenko:
Aether uses Maven version comparison rules to determine if two
versions are equal, and considers 2 and 2.0 to be the same version,
which is not appropriate for index searches.

Changed the code to use aether logic to order versions, but fall
back to string comparison to determine if versions are equal or not.


peterlynch:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF32-1 so that I can experiment with the Nexus using this

ifedorenko:
the build failure is likely because I did "git push -f" to amend some changes

peterlynch:
try again https://bamboo.zion.sonatype.com/browse/NXO-OSSF32-2

cstamas:
+1

PR #695, opened 2012-12-06
Smaller fixes

cstamas:
Smaller fixes:
Those are:
- added toString to RepositoryGroupMembersChangedEvent (eases debugging)
- AbstractConfigurable, subscriber methods made final
- AbstractGroupRepository, properly overriding method instead of subscriber method. Original code was not correct,
  as the work done here belongs to prepareToSave actually.

Latter two fixes the "groups are twice called by event bus" issue, that is also
fixed by bumping EventBus to 13.0.1 (done by Alin), but still, this code now is
better and more correct.


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF34

adreghiciu:
+1

cstamas:
CI +1'd

PR #696, opened 2012-12-07
Introducing new events

cstamas:
Related to NXCM-4606


adreghiciu:
This events are fine but do we need them now? Maybe we wait with this till we have a proper solution to proxy a group problem as in the one you described where slave will create a group and sync member changes

cstamas:
I think we do need them, at least to prevent Staging using/extending RepositoryItemBatchEventAdded directly, and let it extend RepositoryItemBatchEventAddedToGroup instead.

jdillon:
Does this fix something that is currently problem after Alin's change to even ordering yesterday?

If not, then -1 to avoid adding this to solve a problem which is fixed already.

peterlynch:
related to NXCM-4606, but is this required for 2.3?

ifedorenko:
-1. agreed on fix for nxcm-4606 does not require these events.

cstamas:
Closing for now

PR #699, opened 2012-12-08
Fix linkbutton, when handler is configured, do not let even...

jdillon:
...t propagate to browser so that it attempts to follow the a.href.

Also fixed handling of disabled, should be e.stopEvent() not this.stopEvent()

https://issues.sonatype.org/browse/NEXUS-5433


peterlynch:
+1 because it is a regression

nabcos:
+1

PR #702, opened 2012-12-13
Nexus 5447 concurrent index remove

ifedorenko:

PR #697, opened 2012-12-07
NXCM-4852: EOF on deploy

cstamas:
Introduced new LocalStorage Exception to mark this, and translating it on REST level to some other status that 500 Server Error, that automagically kicks in on those response codes.

PR #703, opened 2012-12-14
NEXUS-5447: NXCM-4864: NXCM-4852: Remote EOF translation

cstamas:
Builds upon (and hence, contains it's changes until it's merged into master)
https://github.com/sonatype/nexus/pull/702

Issue NXCM-4864 revealed that pull request
https://github.com/sonatype/nexus/pull/697
is insufficient to handle case when proxy remote peer
drops connection unexpectedly.

NXCM-4864 reveals what exact exception is thrown by HC4
when "remote peer" drops connection unexpectedly. Inspecting
the exception also reveals it is not an `java.io.EOFException`
descendant, but descends directly from `java.io.IOException`.
Hence, changes in pull request
https://github.com/sonatype/nexus/pull/697
(that kicks in on EOFException) will never kick in (and
this case will be handled as "generic IO problem" and will
auto block remote peer too).

So, as we introduced a "wrapper" around the HC4 stream, it
makes sensible to have it "translate" HC4 specific exception
into one that is recognized (and handled) by Nexus Core.
Moreover, as it is highly likely that other transports will
use other exceptions (descending or not from EOFException),
using this technique we could always ensure that this case is
properly handled by Nexus Core.

Here, newly introduced InterruptibleInputStream -- since is
already HC4 RRS specific, is modified to make it more
adapted for Nexus core: is extended to translate
the `org.apache.http.ConnectionClosedException` thrown by
`org.apache.http.impl.io.ContentLengthInputStream` read methods
too.


cstamas:
Then (as post 2.3 thing), should a new wrapper be introduced to perform the exception translation?

PR #704, opened 2012-12-14
Fix MUtils to handle 'too short' checksums

nabcos:
Unguarded index of(" ") was triggering an exception and resulted in
relaying HTTP 500 status for staging close with checksum rule check.


cstamas:
+1

adreghiciu:
+1

PR #705, opened 2012-12-14
Proxy EOF detection.

cstamas:
Issue NXCM-4864 revealed that pull request
https://github.com/sonatype/nexus/pull/697
is insufficient to handle case when proxy remote peer
drops connection unexpectedly.

NXCM-4864 reveals what exact exception is thrown by HC4
when "remote peer" drops connection unexpectedly. Inspecting
the exception also reveals it is not an `java.io.EOFException`
descendant, but descends directly from `java.io.IOException`.
Hence, changes in pull request
https://github.com/sonatype/nexus/pull/697
(that kicks in on EOFException) will never kick in (and
this case will be handled as "generic IO problem" and will
auto block remote peer too).

So, we introduced a "wrapper" around the HC4 stream, to
have it "translate" HC4 specific exception
into one that is recognised (and handled) by Nexus Core.
Moreover, as it is highly likely that other transports will
use other exceptions (descending or not from EOFException),
using this technique we could always ensure that this case is
properly handled by Nexus Core.

Here, newly introduced Hc4InputStream is meant to translate
the `org.apache.http.ConnectionClosedException` thrown by
`org.apache.http.impl.io.ContentLengthInputStream` read methods.


adreghiciu:
UT/IT ?

+1 beside above

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RemoteStorageEofException.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Remote storage exception for cases when outbound request is unsuccessful, due to remote peer hang up on us unexpectedly.
+ * Main use of this exception is to "translate" various EOF notification for various Remote Repository Storage implementations
+ * into single exception handled by Nexus Core.
+ *
+ * @author cstamas
+ * @since 2.4
+ */
+public class RemoteStorageEofException
```
peterlynch: Poorly named - 'EofException' reads as E of Exception to me - could argue RemoteStorageEOFException is more consistent at least with JDK and acceptable - but would it be so bad to use 'RemoteStorageEndOfFileException' I wonder

Same with Local...

cstamas:
Applied comments, added UT for this case (remote drops connection during proxy caching).

PR #706, opened 2012-12-18
NXCM-4861: Fixing the ArtifactStoreHelper.storeItemWithChecksums method

cstamas:
In some cases, current imple can result in remote item download
for some proxy configuration, like itemAgingActive=false used by
procurement.

Fixing it, to enforce "localOnly" flag. Basically, same code
is now executed as in case of cache (where also store+retrieve)
happens, see:
org.sonatype.nexus.proxy.repository.AbstractProxyRepository#doCacheItem

Other minor cleanup (mostly javadoc) done.

Added UT that verifies that maven repo now receives a retrieve
with localOnly=true flag.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RequestContext.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RequestContext.java
@@ -30,13 +30,13 @@ public class RequestContext
     /** Context URL of the original resource requested on the incoming connector. */
     public static final String CTX_REQUEST_URL = "request.url";
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (proxy: do not attempt remote access at all, else: no effect). */
     public static final String CTX_LOCAL_ONLY_FLAG = "request.localOnly";
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (proxy: force remote access -- might still serve local if up2date, else: no effect). */
     public static final String CTX_REMOTE_ONLY_FLAG = "request.remoteOnly";
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (group: do not "dive" into members, else: no effect). */
     public static final String CTX_GROUP_LOCAL_ONLY_FLAG = "request.groupLocalOnly";
 
     /** Context key for condition "if-modified-since" */
@@ -55,15 +55,11 @@ public class RequestContext
 
     public RequestContext()
     {
-        super();
-
-        this.parent = null;
+        setParentContext( null );
     }
 
     public RequestContext( RequestContext parent )
     {
-        this();
-
         setParentContext( parent );
     }
 
@@ -180,7 +176,7 @@ public class RequestContext
     /**
      * Sets the request remote only.
      * 
-     * @param requestremoteOnly the new request remote only
+     * @param requestRemoteOnly the new request remote only
      */
     public void setRequestRemoteOnly( boolean requestRemoteOnly )
     {
@@ -207,7 +203,7 @@ public class RequestContext
     /**
      * Sets the request group local only.
      * 
-     * @param requestremoteOnly the new request group local only
+     * @param requestGroupLocal the new request group local only
      */
     public void setRequestGroupLocalOnly( boolean requestGroupLocal )
     {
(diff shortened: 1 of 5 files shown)


cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF47-1

nexus/nexus-proxy/src/test/java/org/sonatype/nexus/proxy/maven/Nxcm4861Test.java
```diff
@@ -0,0 +1,136 @@
...
+import org.sonatype.nexus.proxy.AbstractProxyTestEnvironment;
+import org.sonatype.nexus.proxy.EnvironmentBuilder;
+import org.sonatype.nexus.proxy.ResourceStoreRequest;
+import org.sonatype.nexus.proxy.maven.maven2.M2Repository;
+import org.sonatype.nexus.proxy.maven.maven2.M2RepositoryConfiguration;
+import org.sonatype.nexus.proxy.repository.Repository;
+import org.sonatype.nexus.proxy.storage.local.LocalRepositoryStorage;
+import org.sonatype.sisu.litmus.testsupport.mock.MockitoRule;
+
+public class Nxcm4861Test
```
ifedorenko: This needs more descriptive name, that tells what the test actually tests, something like Nxcm4861ArtifactStoreHelper_storeItemWithChecksumsTest

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RequestContext.java
```diff
@@ -30,13 +30,13 @@
     /** Context URL of the original resource requested on the incoming connector. */
     public static final String CTX_REQUEST_URL = "request.url";
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (proxy: do not attempt remote access at all, else: no effect). */
     public static final String CTX_LOCAL_ONLY_FLAG = "request.localOnly";
```
peterlynch, line 36: since this is public javadoc, why not link directly to ProxyRepository interface via link gives relevant context plus aids IDE refactoring and and context

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RequestContext.java
```diff
@@ -30,13 +30,13 @@
...
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (proxy: force remote access -- might still serve local if up2date, else: no effect). */
```
peterlynch: same javadoc suggestion - reference 'up2date' variable, method or something - ie. more detail

Also should be "Context flag to mark a remote request local only"

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/RequestContext.java
```diff
@@ -30,13 +30,13 @@
...
     public static final String CTX_REMOTE_ONLY_FLAG = "request.remoteOnly";
 
-    /** Context flag to mark a request local only. */
+    /** Context flag to mark a request local only (group: do not "dive" into members, else: no effect). */
```
peterlynch: would link to GroupRepository interface, not just use 'group'

Also should be "Context flag to mark a group request local only"

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/ResourceStoreRequest.java
```diff
@@ -34,20 +34,26 @@
...
     /** Used internally to track applied mappins */
-    private Map<String, List<String>> appliedMappings;
+    private final Map<String, List<String>> appliedMappings;
 
+    /**
+     * Constructor.
+     *
+     * @param requestPath
+     * @param localOnly
+     * @param remoteOnly
```
peterlynch: What's the point of the javadoc if you do not describe what the params are for

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/ResourceStoreRequest.java
```diff
@@ -349,10 +375,12 @@ public void addAppliedMappingsList( Repository repository, List<String> mappingL
     @Override
     public String toString()
     {
-        StringBuilder sb = new StringBuilder( getClass().getSimpleName() );
-        sb.append( "(requestPath=\"" );
-        sb.append( getRequestPath() );
-        sb.append( "\")" );
-        return sb.toString();
+        return "ResourceStoreRequest{" +
+            "requestPath='" + requestPath + '\'' +
```
peterlynch, line 379: why add the artificial slash here? to request path

...and before it was a leading slash, now it is added at end.

PR #707, opened 2012-12-29
[NXCM-4895] Show reason why a repository was auto blocked

adreghiciu:
For now only SSL untrusted and that host name verification failed is shown, yet that could be extended.
We can go further with this by introducing an exception -> reason converter interface so plugins could easy contribute but that could be over engineering


peterlynch:
related somewhat - https://issues.sonatype.org/browse/NEXUS-5466

cstamas:
+1

peterlynch:
Can we have a separate issue for this feature? - consider how it could be mentioned in release notes

Also a screenshot of how it is displayed in UI. I have no sense of that here.

nexus/nexus-webapp/src/main/webapp/js/nexus/repository/AbstractRepoPanel.js
```diff
@@ -283,6 +283,10 @@ Ext.extend(Sonatype.repoServer.AbstractRepoPanel, Ext.Panel, {
           {
             sOut += status.proxyMode === 'BLOCKED_AUTO' ? ' - Remote Automatically Blocked' : ' - Remote Manually Blocked';
             sOut += remoteStatus === 'available' ? ' and Available' : ' and Unavailable';
+            if (remoteStatus.search(/unavailable:/) === 0)
+            {
+              sOut += ' (' + status.remoteStatus.substr(12) + ')';
```
peterlynch: How does this protect against bad markup  - is it being escaped in case backend message needs it? Looks like in some case we have no idea what will actually get rendered, how long that text is, how long text will affect UI layout, etc.

peterlynch:
I like the idea here - but because I have seen no screenshots, there is no issue for this feature, and in some cases appears brittle, not sure about just committing as is.

nexus/nexus-webapp/src/main/webapp/js/nexus/repository/AbstractRepoPanel.js
adreghiciu: If a reason is provided, the reason will be appended in round brackets after the current text. The length is not important as the column has a fix size, so layout will not be affected. Of course the text could contain html tags but that that is unlikely to happen. Yet to be sure, @nabcos could help us to avoid that

nabcos:
I changed the reason string to be displayed on its own line. It was added in parentheses before, but the column was not rendered wide enough by default, and I think this looks better:

image

peterlynch:
I agree it looks good - thx for screenshot.

I created https://issues.sonatype.org/browse/NEXUS-5481 for this feature.

adreghiciu:
CI +1 https://bamboo.zion.sonatype.com/browse/NXO-OSST-206/

PR #709, opened 2013-01-03
Adjusted nexus-proxy unit test to support Windows builds

whyjustin:
The nexus-proxy unit tests errored out on a Windows build due to the direct string comparison of folder structure using OS specific slashes. These changes replace "/" with the OS agnostic File.seperator.


peterlynch:
+1

Thanks!

Sorry about that - our windows CI job was mostly offline during December so we missed this.

PR #710, opened 2013-01-04
Move nexus/* to ..

jdillon:
This finishes up the work started last year, which we were waiting on the squash merge before implementing.


peterlynch:
make it so

PR #712, opened 2013-01-06
Replaced testng with junit4

ifedorenko:
CI says +1 https://bamboo.zion.sonatype.com/browse/NXO-OSSF5-3

### DIFF
--- a/nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/client/nexus725/Nexus725InitialRestClientIT.java
+++ b/nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/client/nexus725/Nexus725InitialRestClientIT.java
@@ -16,6 +16,9 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.sonatype.nexus.client.NexusClient;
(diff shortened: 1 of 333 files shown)


jdillon:
YAY!

adreghiciu:
+1

cstamas:
+1

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/client/nexus725/Nexus725InitialRestClientIT.java
```diff
@@ -68,7 +68,7 @@ public void getRepoListTest()
         NexusClient client = this.getConnectedNexusClient();
 
         List<RepositoryListResource> repos = client.getRepositories();
-        Assert.assertTrue( repos.size() > 0, "Expected list of repos to be larger then 0" );
+        Assert.assertTrue( "Expected list of repos to be larger then 0", repos.size() > 0 );
```
nabcos, line 71: Please use hamcrest instead.
(SCNR :o)

ifedorenko, line 71: Yes, these do need to be cleaned up and fixed, but they are unrelated to testng->junit4 migration and should be dealt with separately.

<rant>
At least this code asserts something meaningful. There are places where tests assert static string is not null (likely due to mixed parameter order when the tests were originally migrated from junit to testng) and places tests that simply make no sense, like assertNotNull(int) I mentioned yesterday. 
</rant>

peterlynch:
+1

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/client/nexus725/Nexus725InitialRestClientIT.java
jdillon, line 71: I agree, it should using hamcrest or fest or something else, but that we shouldn't complicate the killing of testng with rewriting tests wrt to how they assert.

PR #713, opened 2013-01-07
Plugin for BouncyCastle.

cstamas:
Inspired by H2 plugin...

It serves multiple purposes:
- centralises presence of BC (currently we have 4 JARs - 3 versions scattered as different plugin deps)
- centralises JCE registration/unregistration in one single place.

This only adds new core plugin and simple code that registers/deregisters it
and marks BC classes as "plugin shared" dependencies.


plugins/bouncycastle/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+      <groupId>org.sonatype.nexus</groupId>
+      <artifactId>nexus-plugin-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk16</artifactId>
+      <version>1.46</version>
+    </dependency>
```
jdillon: We may want to include some of the other stuff here too, like what ever is included for pgp, just so that its all in one place.

plugins/bouncycastle/src/main/java/org/sonatype/nexus/plugins/bcprov/internal/BCEventInspector.java
```diff
@@ -0,0 +1,58 @@
...
+        if ( evt instanceof NexusInitializedEvent )
+        {
+            bcManager.registerProvider();
+        }
+        else if ( evt instanceof NexusStoppedEvent )
+        {
+            bcManager.unregisterProvider();
+        }
+    }
+}
```
jdillon: Should prefer new event handling stuff to use the nicer EventBus api instead of the legacy appevents stuff.

jdillon:
I wonder if maybe we want to make this just a generic crypto plugin and have it depend on goodies-crypto to pick up its CryptoHelper and support classes?

nabcos:
I would prefer this to become a general 'crypto plugin' as well.

jdillon:
We may still need to put BC directly into core, as there are bits like the license extension bits which need this... and IIUC that isn't possible with this isolated into a plugin?

cstamas:
I might be wrong (Ben would need to confirm this), but: BC is used by lic-creator only. And lic-creator is _unused_ in lic-ext (it was, to create "temp" eval licenses). Also, just checked Nx Pro 2.3, and BC is not in core (while nx-lic-ext is), that also makes me think this.

By reading license-creator, only Key and license creator comps are using BC classes, and those two are referenced from CLI tool only. Hence, even if license-creator is needed by nx-lic-ext (I doubt), those two classes that need BC seems unused on runtime (again, BC was never in core of Nx).

@nabcos any comment?

nabcos:
license-bundle (which is consumed by licensing-extension) does not depend on bc. Only license-creator does, and only because of https://issues.sonatype.org/browse/PRDSARA-15, IIRC. It's not needed to read or write the Nexus license keys.

jdillon:
@nabcos okay, thx for clarification... I was only looking for the bcprov artifact and didn't pay too much attention that its only used in an exclusion in nexus-licensing-extension.  Its actually a proper dep too, not enough coffee I guess, though its marked as provided, which is odd ish...

I'm still not sure however that this is best in a plugin though... the handling of the registration of the provider via events for example, creates a window where plugins might not get the provider installed.

IMO this would probably be better just added to the core, and install the provider very early in an servlet context listener before any of the component/plugin machinery gets started.

cstamas:
Maybe the registration should happen in the contructor of the BCManagerImpl then? (and we can leave the "tear down" part for NexusStoppedEvent).

Even today's PM loads plugins in "proper order", the window (or expectation of having BC not yet registered) might happen only for plugins not declaring dependency on BC plugin. Meaning, PM guarantees that a plugin depending on BC plugin will be loaded after BC plugin is loaded. I guess (guess!) something similar is possible in OSGi land too...

Also, Core today does not use BC. If we do plan to use BC in core, then yes, moving it into core is the way to go.

cstamas:
Modified BC registration, it now happens in component constructor that is eager singleton.

If no core needs BC, this should be good enough. As we tend to skim down (not fatten up) Core, I think this is okay, as any BC consumer will (should) live in a plugin, hence no lack of BC should be happen to them.

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF7-1

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF7-2

jdillon:
-0.9057 I think this is just too complicated for its purpose.

intf BCManager
class BCManagerImpl
class BCPluginEventHandler
class BCProviderImpl

There is IMO no reason that any of plugins should be using these components.  Standard Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) should always be used to find BC.  The components you've added here indicate that its okay or recommended to use those to get these components and I don't think we need to provide an alternative api to find these, when they are all bound to the JVM and the JRE provides accessors for this already.

This has potential harm to leak and potentially cause the BCManagerImpl.getProvider() and Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) to be come out of sync.

I think only the event handler to register the provider is needed... I'm still on the fence if this should be in a plugin or not honestly.

cstamas:
I believe you meant to say "only the event handler to UNregister the provider is needed", as you were objecting (see above) to use eventing to _register_ it in the first place? O_o

I completely agree that an event handler is needed to unregister it (think WAR). Registering should happen in constructor. And I think this would need to be a plugin, to not clobber core with things like BC, goodies crypto and their deps, as I assume we do want to keep core slim as slim possible.

cstamas:
Applied comments and renamed the plugin to reflect it's main purpose better.

plugins/crypto/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <artifactId>nexus-plugins</artifactId>
+    <version>2.4-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nexus-crypto-plugin</artifactId>
+  <packaging>nexus-plugin</packaging>
+
+  <properties>
+    <nexus.plugin.name>Nexus Crypto Services Provider Plugin</nexus.plugin.name>
+    <nexus.plugin.description>Provides Bouncy Castle JARs and registers it as a JCE provider and provides crypto goodies.</nexus.plugin.description>
```
jdillon: This should be normalized like all other plugins <pluginName> <pluginDescription>

PR #714, opened 2013-01-09
bring plexus-rule classes into nexus-compat

jdillon:
https://issues.sonatype.org/browse/NEXUS-5475


jdillon:
There is still some issues to sort out how to get staging + procurement to use the same "rule" api, as right now staging uses this stuff (from plexus-rule) and procurement uses its own "rule" api.

jdillon:
Will not bring this into nexus-compat, its only used by staging.

PR #715, opened 2013-01-12
Common parent for OSS plugins

jdillon:
https://issues.sonatype.org/browse/NXCM-4924

Still pending how this might work in our pro plugins, or in thirdparty plugins... inheritance of a single parent to provide shared configuration is a real limitation... which means that some configuration will always have to be duplicated.


jdillon:
Some hacks to sort out some issues with deps... might need a few more.  Trying to treat plugins/\* as isolated (in case they are pulled out into separate projects) but some of the DM configuration presently causes some duplication.

Also the provided-dependencies-maven-plugin only pays attention to deps defined in a project, not anything in DM which the project has access too, so in a few places had to add hard deps for things which were already in DM.

jdillon:
Once this is basically working in OSS, then we can update an external OSS plugin (likely capabilities) to use this parent for its project.  Then look at how to sort out the PRO side.  ATM I think for PRO we will be forced to duplicate most of plugin-parent/pom.xml content so that the PRO version of this can extend from the corporate pom and pick up all that configuration... so we should try to keep the plugin-parent/pom.xml as small as possible to make it manageable, and there is still more work to be done here once the basic concept is vetted and proven will work.

jdillon:
3 examples of use of nexus-plugin-parent: capabilities, yum, site-repository

plugin-parent/pom.xml
```diff
@@ -0,0 +1,522 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.sonatype.forge</groupId>
+    <artifactId>forge-parent</artifactId>
+    <version>30</version>
```
peterlynch: 31 is avail if you want

jdillon: Okay, I'll update it.

peterlynch:
looked over pretty close and didn't see anything to complain about

jdillon:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF10-14 :white_check_mark:

PR #716, opened 2013-01-14
NEXUS-5418 lock checksum items during delete and read

ifedorenko:
Bamboo says +1 https://bamboo.zion.sonatype.com/browse/NXO-OSSF11-1

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/AbstractMavenRepository.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/AbstractMavenRepository.java
@@ -27,11 +27,13 @@ import org.sonatype.nexus.proxy.ItemNotFoundException;
 import org.sonatype.nexus.proxy.RemoteAccessException;
 import org.sonatype.nexus.proxy.ResourceStoreRequest;
 import org.sonatype.nexus.proxy.StorageException;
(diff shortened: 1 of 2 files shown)


cstamas:
While it looks fine, IMO this is deadlock prone.

I wanted to implement this issue in different way: by introducing a "subordinate" (or just hiding this from API user by some logic in created method of repository) for UID locks. Simply put, in case of Maven2 repositories, `path.jar` and `path.jar.sha1` should use the _same key_ for their UID locks, to really have them locked/unlocked as one. In this case, no deadlock would be possible either, as both, artifact and it's checksum access would happen using same lock.

So, while these changes here are fine, more would be needed in UID lock factory, that would "consult" the given repository for the passed in path, and in this case a special case would happen: UID path and it's UIDLock would use different "keys"  (UID path would be `path.jar.sha1` but UIDLock would use key `path.jar`)...

ifedorenko:
Can you describe deadlock scenario?

cstamas:
+1 after chatting with Igor

peterlynch:
can you exercise this code in a test with multiple threads - or is there already such a test you are confident this change works?

ifedorenko:
No, no tests that cover this scenario nor I know how to implement such tests reliably.

PR #717, opened 2013-01-15
Drop security-rest bits, which were moved into restlet1x plugin already

jdillon:
https://issues.sonatype.org/browse/NXCM-4937


jdillon:
Related: https://github.com/sonatype/security/pull/21

cstamas:
Thanks!

PR #718, opened 2013-01-15
Request flags

cstamas:
ResourceStoreRequest flags extended with new "asExpired" (see below) and covered with tests. They were used in some tests spread across multiple UTs but I did not found any UT specially concentrating on them (and covering many cases).

Introduced "asExpired" flag. This new flag aligns with existing "asLocal", "asRemote" flags. In contrary to "asRemote" flag, this one makes Nexus process the request as the item in question was expired (even if it is fresh actually), doing the HEAD then GET (if remote is newer). This is different than "remoteOnly" as in this case unconditional GET is _always_ done, content is always fetched, even if it is _same as the local cached one_. This new logic will be used by whitelist implementation to keep WL prefix file "in sync" with remote in much simpler way than expire+refetch (as expire is actually "walker" + retrieve when another retrieve is about to happen).

While experimenting with these, some interesting cases were discovered: in some cases Nexus does not behave as "accumulating proxy" but rather removes local cache content when  remote is removed (asRemote case, see UT). Added UT that covers asLocal, asRemote and the newly added asExpired cases for now, to have their functionality covered.

The new UT covers following:
- localOnly flag
- remoteOnly flag
- asExpired flag
- both localOnly and remoteOnly flags combined

Note: SystemPropertiesHelper.java changes went in by mistake (also used by WL). Is unrelated to these changes but will be also used in WL.


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF12-1

cstamas:
CI +1'd (one unstable and unrelated test failed: Nexus570IndexArchetypeIT)
https://bamboo.zion.sonatype.com/browse/NXO-OSSF12-2

nabcos:
I would rename the new flag because I find 'asExpired' too cryptic ('remoteIfNewer'? Makes more sense between 'localOnly', 'remoteOnly', IMO)

+1 otherwise.

cstamas:
Yes, the "localOnly" and "remoteOnly" (internally named, but "?asLocal" and "?asRemote" on REST!) were unfortunate names too. I decided with "asExpired" as this flag really tells nexus "handle this request's item -- if in cache -- as expired, so do check remotely if exists newer version of it, and if exists, give me that one" (if not in cache, this flag has no effect at all, does not change behaviour). I'll leave to natives to properly name this thing, as my reputation of naming things is, heh, as is.

ifedorenko:
What is the usecase for the new flag?

peterlynch:
Yes good question - use case?

cstamas:
Main use case is use of this flag in proxy404 implementation, but here are some other examples:
1. _Troubleshooting a Maven build:_ you know (as you just deployed or CI did) that Nexus that proxies repo where deployed have stale maven metadata (or your build just broke and you figured this out, as new timestamped snapshot was not picked up or such). Today, you'd have to go to Nexus UI to run "Expire Cache" against given proxy (now, you could be run it just from the grid, resulting of cache expiration of whole repo -- as lazy user would do, or you can navigate in the tree deeply down to the artefact and click on it, does not matter). "Expire Cache" is a heavy weight beast, that creates background task, invokes Walker, yada yada yada. With "asExpired" flag you can resolve this problem not even leaving your console:

```
$ wget http://nexus/...path.../maven-metadata.xml?asExpired
```

And instantly you can even assure that Nexus did get the new file from remote.
1. _"Expire Cache" does not expire all files in cache:_ The task uses Walker configured with a WalkerFilter that _avoids_ meta and hidden files. You cannot expire stuff like `/.meta/foo.txt` from UI either.
2. _Control or use different expiration for some files in a proxy repository:_ If you want to apply your own "update interval" to certain file(s), today you'd need to `expireProxyCaches(itemPath)` and then `retrieveItem(itemPath)`. But, the 1st invocation is very heavy weight: it uses Walker (even if it is about to expire one single file), that needs context created, processor and filter, then walking starts, yada yada yada (produces at least 2 or 3 lines in nexus log). The "asExpired" flag gives you "surgical precision" and simplicity: to ensure you really want to update (or check for update better) NOW -- as opposite to "when ItemMaxAge would expire", or to run expire machinery against it -- is just to perform a simple `retrieveItem(request)` where request have this flag set.

As use case 1 shows, by exposing this flag over REST (similarly to "asLocal" and "asRemote" exposed flags), you gain better control of your caches far away from UI.

But 2. and 3. point are the keys for the use in proxy404: prefix file will be need to be kept in sync with remote in intervals defined by proxy404 config, not proxy repo's MaxAges.

ifedorenko:
Only one of localOnly, remoteOnly and asExpired flags should be allowed to set simultaneously on a request. Any other combination should lead to IllegalArgumentException, not item not found.

cstamas:
Applied Igor comment: RequestContext now allows only one of the 3 flags to have set.

cstamas:
2nd CI with change above included started
https://bamboo.zion.sonatype.com/browse/NXO-OSSF12-3

cstamas:
2nd CI +1'd this

ifedorenko:
+1

PR #723, opened 2013-01-18
Warn when magic annotations (@ExtensionPoint @Managed) kick in

jdillon:
Logs warnings when components are reliant on the ExtensionPoint and Managed "magic" instead of properly using Named and Singleton.

Logs detected Plexus components as DEBUG too, once we are close to killing Plexus completely we can flip this to WARN.


peterlynch:
+1 for the idea

peterlynch:
How much debug logging will this spill out?

Want to avoid filling up customer logs with DEBUGs with this when they are trying to find some other problem.

jdillon:
It will spit out a bit of DEBUG on bootup warning about Plexus components (only once at bootup, not constantly), otherwise it shouldn't spit out anything as I've been fixing components as I find them.

jdillon:
I wonder if I should also warn about use of legacy com.google.inject.\* annotations.

mosabua:
I am sure it would help some devs.

PR #724, opened 2013-01-18
Add guice AOP support

jdillon:
Adds guice AOP support, and an example aspect which capturing timing information for methods adorned with @Timed.

Reference: http://code.google.com/p/google-guice/wiki/AOP

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
@@ -466,7 +466,14 @@ public class DefaultNexusPluginManager
         final ClassSpace annSpace = new URLClassSpace( pluginRealm, scanList.toArray( new URL[scanList.size()] ) );
         beanModules.add( new NexusAnnotatedBeanModule( annSpace, variables, exportedClassNames, repositoryTypes ) );
 
-        container.addPlexusInjector( beanModules, resourceModule );
+        final Module[] modules = {
+            resourceModule,
+
+            // Add support for @Timed
+            new TimingModule()
+        };
+
+        container.addPlexusInjector( beanModules, modules );
 
         for ( final RepositoryTypeDescriptor r : repositoryTypes )
         {
--- a/nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
@@ -54,6 +54,8 @@ import org.sonatype.nexus.proxy.maven.version.Version;
 import org.sonatype.nexus.proxy.maven.version.VersionParser;
 import org.sonatype.nexus.proxy.registry.RepositoryTypeDescriptor;
 import org.sonatype.nexus.proxy.registry.RepositoryTypeRegistry;
+import org.sonatype.nexus.timing.Timed;
+import org.sonatype.nexus.timing.TimingModule;
 import org.sonatype.nexus.util.AlphanumComparator;
 import org.sonatype.plexus.appevents.ApplicationEventMulticaster;
 import org.sonatype.plexus.appevents.Event;
@@ -132,6 +134,7 @@ public class DefaultNexusPluginManager
         return new HashMap<GAVCoordinate, PluginResponse>( pluginResponses );
     }
 
+    @Timed("nexus.pluginManager.activatePlugins")
     public Collection<PluginManagerResponse> activateInstalledPlugins()
     {
         final List<PluginManagerResponse> result = new ArrayList<PluginManagerResponse>();
@@ -266,6 +269,7 @@ public class DefaultNexusPluginManager
         return getActivatedPluginGav( gav, strict ) != null;
     }
 
+    @Timed
     protected PluginManagerResponse activatePlugin( final GAVCoordinate gav, final boolean strict,
                                                     final Set<GAVCoordinate> installedPluginsFilteredByGA )
     {
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeSource.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeSource.java
@@ -16,7 +16,7 @@ import java.util.Map;
 
 import org.sonatype.plugin.ExtensionPoint;
 
-import com.google.inject.Singleton;
+import javax.inject.Singleton;
 
 /**
  * RepositoryItemUid Attribute source that is contributing attributes to core.
--- a/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/NexusSpecific.java
+++ b/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/NexusSpecific.java
@@ -21,13 +21,13 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import com.google.inject.BindingAnnotation;
+import javax.inject.Qualifier;
 
-@BindingAnnotation
+@Qualifier
 @Target( { TYPE, FIELD, PARAMETER, METHOD } )
 @Retention( RUNTIME )
 public @interface NexusSpecific
 {
-
+    // empty
 }
 
--- a/nexus-web-utils/src/main/java/org/sonatype/nexus/web/PlexusContainerContextListener.java
+++ b/nexus-web-utils/src/main/java/org/sonatype/nexus/web/PlexusContainerContextListener.java
@@ -39,6 +39,7 @@ import org.sonatype.appcontext.source.PropertiesFileEntrySource;
 import org.sonatype.appcontext.source.StaticEntrySource;
 
 import com.google.inject.Module;
+import org.sonatype.nexus.timing.TimingModule;
 
 /**
  * This ServeletContextListener boots up Plexus in a webapp environment, if needed. It is safe to have it multiple times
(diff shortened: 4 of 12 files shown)


nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
```diff
@@ -132,6 +134,7 @@ public DefaultNexusPluginManager( final RepositoryTypeRegistry repositoryTypeReg
         return new HashMap<GAVCoordinate, PluginResponse>( pluginResponses );
     }
 
+    @Timed("nexus.pluginManager.activatePlugins")
```
jdillon, line 137: Example of explicit name.  Will add timing metrics when plugins are activated on NX boot.

nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
```diff
@@ -266,6 +269,7 @@ protected boolean isActivatedPlugin( final GAVCoordinate gav, final boolean stri
         return getActivatedPluginGav( gav, strict ) != null;
     }
 
+    @Timed
```
jdillon, line 272: Example of dynamic/detected name... pending perf/optimizations as to if this form is good or bad.  Will add timing metrics for activation of a single plugin.

jdillon:
ATM only logs the timing, which is something like this:

```
... TimedInterceptor - Simon Stopwatch: total 4.70 s, counter 15, max 1.51 s, min 7.00 us, mean 313 ms [org.sonatype.nexus.plugins.DefaultNexusPluginManager.activatePlugin-3 INHERIT]
... TimedInterceptor - Simon Stopwatch: total 4.86 s, counter 16, max 1.51 s, min 7.00 us, mean 304 ms [org.sonatype.nexus.plugins.DefaultNexusPluginManager.activatePlugin-3 INHERIT]
... TimedInterceptor - Simon Stopwatch: total 4.86 s, counter 17, max 1.51 s, min 7.00 us, mean 286 ms [org.sonatype.nexus.plugins.DefaultNexusPluginManager.activatePlugin-3 INHERIT]
... TimedInterceptor - Simon Stopwatch: total 4.89 s, counter 18, max 1.51 s, min 7.00 us, mean 272 ms [org.sonatype.nexus.plugins.DefaultNexusPluginManager.activatePlugin-3 INHERIT]
... TimedInterceptor - Simon Stopwatch: total 2.62 s, counter 1, max 2.62 s, min 2.62 s, mean 2.62 s [nexus.pluginManager.activatePlugins INHERIT]
```

Javasimon has integrated JMX support, so we could not log anything and simply capture these timings to mbeans for dynamic inspection via jconsole, etc.

jdillon:
The main implication here, is that tests that need to verify functionally based on a guice-aop aspect _must_ be injected, plain non-guice pojo tests won't get the weaved aspect.  I think this is acceptable.

That and the main guice-aop limitations:
- Classes must be public or package-private.
- Classes must be non-final
- Methods must be public, package-private or protected
- Methods must be non-final
- Instances must be created by Guice by an @Inject-annotated or no-argument constructor

... and I also think this is probably acceptable.

nexus-web-utils/src/main/java/org/sonatype/nexus/web/PlexusContainerContextListener.java
```diff
@@ -81,6 +82,9 @@ public void contextInitialized( final ServletContextEvent sce )
                 modules.add( new NexusWebModule( sce.getServletContext() ) );
                 modules.add( new AppContextModule( appContext ) );
 
+                // Add support for @Timed
+                modules.add( new TimingModule() );
```
jdillon, line 86: Add @Timed support to core.

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/status/StatusPlexusResource.java
```diff
@@ -70,6 +71,7 @@ public PathProtectionDescriptor getResourceProtection()
     /**
      * Get the status of the nexus server.
      */
+    @Timed
```
jdillon, line 74: Test @Timed on plugin.  Any access to /service/local/status will have timing metrics captured.

nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
```diff
@@ -462,7 +466,14 @@ protected void configure()
...
 
-        container.addPlexusInjector( beanModules, resourceModule );
+        final Module[] modules = {
+            resourceModule,
+
+            // Add support for @Timed
+            new TimingModule()
+        };
+
+        container.addPlexusInjector( beanModules, modules );
```
jdillon, line 476: Add @Timed support to plugins

nexus-core/src/main/java/org/sonatype/nexus/proxy/item/uid/RepositoryItemUidAttributeSource.java
```diff
@@ -16,7 +16,7 @@
 
 import org.sonatype.plugin.ExtensionPoint;
 
-import com.google.inject.Singleton;
+import javax.inject.Singleton;
```
jdillon, line 19: Unrelated to feature in this pull, just fixing to use proper annotations.

nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/NexusSpecific.java
```diff
@@ -21,13 +21,13 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import com.google.inject.BindingAnnotation;
+import javax.inject.Qualifier;
 
-@BindingAnnotation
+@Qualifier
```
jdillon, line 26: Unrelated to feature in this pull, just fixing to use proper annotations.

pom.xml
```diff
@@ -806,6 +806,12 @@
         <version>1.2</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.javasimon</groupId>
+        <artifactId>javasimon-core</artifactId>
+        <version>3.3.0</version>
+      </dependency>
```
jdillon, line 813: Core library to handle timing metric collection: http://code.google.com/p/javasimon/

This library is already in use by smartproxy to collect event timings, introduce here can simplify the smartproxy implementation.

nexus-web-utils/src/main/java/org/sonatype/nexus/web/PlexusContainerContextListener.java
ifedorenko, line 86: Adding @Named to TimingModule does the same thing, only in more scalable manner, i.e. we will not have to explicitly add each interceptor module here.

nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
ifedorenko, line 476: I would be nice not to have to explicitly enable all interceptor modules here, imho. Simply adding @Named to TimingModule did not work, unfortunately, and I don't know enough about guice/sisu to tell why.

jdillon, line 476: I'm not super concerned about it... we won't have too many of these... I think if we have a lot we are doing something wrong.

ifedorenko:
I am generally +1 for enabling guice-aop for nexus core and nexus plugins. It would be nice not to have to explicitly bind interceptor modules in two places (see my line comments), but this is not a blocker.

I am -0 for adding @Timed interceptor. Although I do not see anything particularly wrong neither with the idea nor with implementation, I think more planning is needed before we introduce monitoring capabilities into nexus. Otherwise we risk getting random/inconsistent instrumentation.

nexus-core/src/main/java/org/sonatype/nexus/plugins/DefaultNexusPluginManager.java
ifedorenko, line 476: I think ability to introduce plugin-specific aspects would be nice. Otherwise, yes, not a big concern.

jdillon:
I still like @Timed and its dead simple to use, and as mentioned I'm already doing this same thing via aspectj in smartproxy.  I think its pretty low risk, and if/when we have a better plan, we can easily replace it... either by adapting the aspect's advise, or replacing annotations, etc.

There are at least 2 areas in NX core that could use @Timed now to simplify implementations... and I'm talking about the guava Stopwatch use in remote store impls.

adreghiciu:
+1

ifedorenko:
My concern with @Timed is that it will likely get applied inconsistently throughout the code, i.e. there will be patches of (over) instrumented code and large areas of the code that has not been instrumented. I am fine updating existing SP implementation to use guice-aop, but I'd like to have little upfront planning before we start using @Timed elsewhere. Also, no @Timed refactoring outside of SP in 2.4 please... unless we really run out of stuff to do.

jdillon:
Same could be said about guava Stopwatch use :-P  Only it has a much larger impact on the codebase.

Anyways, this is about getting general AOP into NX and I really don't understand why you are making such a fuss over this one example/poc annotation+aspect :-\

ifedorenko:
All I am saying, for 2.4 only apply @Timed where we already have timers and maybe for the new code we write for 2.4. Do not introduce @Timed in other places until we understand what we want to achieve in the end.

jdillon:
NO WAY... I'm gonna add @Timed to everything :-P :-P

ifedorenko:
merge this then

jdillon:
I'll merge a bit later today, trying to get staging build back to normality first.

jdillon:
FTR:

https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-1 :white_check_mark:

PR #725, opened 2013-01-21
Safer handling of mbean registration

jdillon:
In test env, this fails often in tests (with loud logs).

Simple change to unregister before registration if already exists.

Though opened pull here to see if anyone else has issues with the components in the test-env are only getting lifecycle operations (dispose in this case) once per-test-class... which is the root cause of this problem.

ie. this could be a symptom of a larger problem.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/events/DebugEventInspector.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/events/DebugEventInspector.java
@@ -76,7 +76,11 @@ public class DebugEventInspector
         {
             try
             {
-                ManagementFactory.getPlatformMBeanServer().unregisterMBean( jmxName );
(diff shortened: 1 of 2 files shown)


cstamas:
PlexusTestCase (the oldie one) did dispose the container (hence, dispose was called on Disposable components) per method. Did something change wrt this behavior? Or better, are we creating container per-method (as JMX beans are tried to be registered per method) while we dispose it per class?

I have these errors too, but aside of noise, they should be harmless as far as test outcome is regarded.

jdillon:
Legacy integration-tests do not use PlexusTestCase they are all sub-classes of AbstractNexusIntegrationTest which has no super-class.  AFAICT stopping of NX and the disposal of components is only done at the end of test-class execution, after each test-method... and its been like that for a long time it seems.

jdillon:
I'm gonna merge this I can't stand seeing these MBean exceptions in the test logs.

PR #726, opened 2013-01-22
[REVIEW] NEXUS-5511: Improve not found reasoning

cstamas:
Nexus might response with 404 Not Found in for many reasons, and proxy404 will be "just another" one of them.

Nexus since 2.2 (or 2.3?) bubbles up the actual not found message to REST layer and HTML content page even renders the message to user. But, unfortunately, the message was a constant ("The item not found in ...").

These changes now changes critical places like NFC, local storage, remote storage, processors to give more hints why response is 404.

Aside improving messages, this pull also gives ability to "hint" reason upfront, like the case with localOnly or groupLocalOnly flags is (by setting these flags you still don't know the outcome of the response). A similar technique would be used in proxy404 to tell user that "Whitelist rejected to perform a remote request".


cstamas:
CI:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF22

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundReasons.java
```diff
@@ -0,0 +1,179 @@
...
+import org.sonatype.sisu.goodies.common.FormatTemplate;
+import org.sonatype.sisu.goodies.common.SimpleFormat;
+
+/**
+ * Class with some handy static methods to contruct reasoning for {@link ItemNotFoundException}.
+ * 
+ * @author cstamas
+ * @since 2.4
+ */
+public class ItemNotFoundReasons
```
peterlynch: abstract?

nexus-core/src/main/java/org/sonatype/nexus/proxy/router/DefaultRepositoryRouter.java
```diff
@@ -519,13 +528,14 @@ public RequestRoute getRequestRouteForRequest( ResourceStoreRequest request )
                 if ( !repository.isExposed() )
                 {
                     // this is not the main facet or the repo is not exposed
-                    throw new ItemNotFoundException( request );
+                    throw new ItemNotFoundException( reasonFor( request, "Repository %s exists but is not exposed!",
```
peterlynch: I could live without the exclamation marks ;)

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
```diff
@@ -482,7 +496,7 @@ protected void handleRenameOperation( File hiddenTarget, File target )
             if ( success )
             {
                 getLogger().info( "Rename operation succeeded after {} retries on {} --> {}",
-                                  new Object[]{ i, hiddenTarget.getAbsolutePath(), target.getAbsolutePath() } );
+                    new Object[] { i, hiddenTarget.getAbsolutePath(), target.getAbsolutePath() } );
```
peterlynch: I think var args make new Object[] not required if memory serves. Same in several other places.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -25,41 +25,89 @@
...
+     * @since 2.4
      */
-    @Deprecated
-    public ItemNotFoundException( String path, Throwable cause )
+    public ItemNotFoundException( final ItemNotFoundReason reason, final Throwable cause )
     {
-        super( "Item not found on path " + path, cause );
-        this.repository = null;
-        this.request = null;
+        super( reason.getMessage(), cause ); // will NPE with null reason
```
peterlynch: should add @throws NullPointerException to related javadoc imo

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
cstamas: These lines were not touched by me, it was only reformat doing it. Updating the SLF4J use to the level where new method was introduced on it would require more effort that this here anyway.

nexus-core/src/main/java/org/sonatype/nexus/proxy/router/DefaultRepositoryRouter.java
cstamas: Fixed

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundReasons.java
cstamas: Added private constructor to prevent instantiation.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
cstamas: Reworking this a bit, initial solution was awkward.

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
jdillon: This is generally pretty easy to clean up... and we probably should try to clean this stuff up as we go along if we don't plan to fix all logging usage in one go (which I'm not recommending).

cstamas:
Updated logging to new SLF4J api in DefaultFSPeer.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -25,41 +29,129 @@
 {
     private static final long serialVersionUID = -4964273361722823796L;
 
-    private final ResourceStoreRequest request;
+    /**
+     * Reason of item not found when no repository is involved. Usually ther IS one, so you should use
+     * {@link ItemNotFoundInRepositoryReason} instead. This one is used in places like {@link RepositoryRouter}, where
+     * the "targeted" repository is still unknown or similar places.
+     */
+    public static class ItemNotFoundReason
```
ifedorenko, line 104: This is subjective, but I find it odd that standalone ItemNotFoundReasons creates instances of nested ItemNotFoundException.ItemNotFoundReason. I would probably collapse ItemNotFoundReasons onto ItemNotFoundException, but there are other sensible alternatives too.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -1320,6 +1329,9 @@ protected boolean checkPostConditions( final ResourceStoreRequest request, final
             {
                 if ( !processor.shouldRetrieve( this, request, item ) )
                 {
+                    request.addItemNotFoundReason( reasonFor( request, this,
+                        "Retrieval of request %s in repository %s prevented.", request.getRequestPath(),
+                        RepositoryStringUtils.getHumanizedNameString( this ) ) );
```
ifedorenko: What additional information does the message above provide? Either add info about processor, if it is relevant to the user, or don't provide additional info.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepository.java
```diff
@@ -1184,6 +1195,9 @@ protected StorageItem doRetrieveItem0( ResourceStoreRequest request, AbstractSto
                 if ( !shouldProxy )
                 {
                     // escape
+                    request.addItemNotFoundReason( ItemNotFoundReasons.reasonFor( request, this,
+                        "Proxying of path %s in repository %s prevented", request.getRequestPath(),
```
ifedorenko: Same as above, this does not provide any additional info. Either make it useful or remove it.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepository.java
```diff
@@ -263,11 +267,8 @@ public boolean expireProxyCaches( final ResourceStoreRequest request, final Walk
 
             // fire off the new event if crawling did end, so we did flip all the bits
             eventBus().post(
-                new RepositoryEventExpireProxyCaches(
-                    this, request.getRequestPath(), request.getRequestContext().flatten(),
-                    expireCacheWalkerProcessor.isCacheAltered()
-                )
-            );
+                new RepositoryEventExpireProxyCaches( this, request.getRequestPath(),
+                    request.getRequestContext().flatten(), expireCacheWalkerProcessor.isCacheAltered() ) );
```
ifedorenko, line 270: In the future, please push formatting changes as separate commits. When part of a larger unrelated pull, formatting changes make reviews harder than they should be.

ifedorenko:
-0

I agree with the general approach, but I am concerned about using ResourceStoreRequest to capture ItemNotFoundReason. Ideally, requests should be immutable objects that describe original client requests to the Repository/Storage subsystem. We already use requests to track request execution state, i.e. push/pop path, applied mappings, etc, which I believe is wrong. With this proposed change ResourceStoreRequest will also track request execution results, which will make the problem even worse. 

As far as I can tell AbstractRepository.checkPostConditions and AbstractRepository.checkRequestProcessors are the only two places where the code really needs to collect info about conditions not met, everywhere else we can just throw ItemNotFoundException directly. I suggest we find a cleaner way to communicate info about failing conditions.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
cstamas, line 104: INFReasons class removed.

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF22-4

cstamas:
@ifedorenko I'd like you to give this pull a 2nd pass (as your comments were applied while you were on vacation)

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -1303,6 +1311,8 @@ protected boolean checkRequestProcessors( final ResourceStoreRequest request, fi
             {
                 if ( !processor.process( this, request, action ) )
                 {
+                    // let's not state anything about reason, and allow processor say why it rejected
+                    // by throwing appropriate exception
                     return false;
```
ifedorenko: This is not good enough. RequestProcessor needs a way to provide specific reasons why the request was rejected all the way back to the user. And no, request object is not the right place to store this additional information ;-)

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -1320,6 +1330,8 @@ protected boolean checkPostConditions( final ResourceStoreRequest request, final
             {
                 if ( !processor.shouldRetrieve( this, request, item ) )
                 {
+                    // let's not state anything about reason, and allow processor say why it rejected
+                    // by throwing appropriate exception
```
ifedorenko: Same as above, RequestProcessor needs a way to communicate reasons back to the user.

ifedorenko:
-1. RequestProcessor's need a way to communicate reasons request were rejected back to the user.

rseddon:
If we can't cover the  cases mentioned by Igor before the 2.4 release is done I'd suggest merging this pull request anyhow, and filing a separate bug for those.  What we have currently makes it practically impossible for users to self-diagnose 404's, and this causes a lot of support tickets.

cstamas:
CI job:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF38-1

nexus-core/pom.xml
```diff
@@ -58,6 +58,10 @@
     </dependency>
     <dependency>
       <groupId>org.sonatype.sisu.goodies</groupId>
+      <artifactId>goodies-common</artifactId>
+    </dependency>
```
jdillon, line 62: This may have a bit more impact than this change.  Pluginkit and staging, maybe some others (have to check) also include goodies-common, so we'll have to update those too.

But FTR I'm pro adding goodies-common to core.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -24,42 +29,206 @@
...
+        return new ItemNotFoundReason( SimpleFormat.template( message, params ), request );
+    }
+
+    /**
+     * Creates a new instance of {@link ItemNotFoundInRepositoryReason}.
+     * 
+     * @param request
+     * @param repository
+     * @param message
+     * @param params
```
jdillon: Why do you feel you need to put empty javadoc tags which do not provide any other information?  This is pointless and only increases the chances that when a parameter is renamed that the docs will be out of sync (which I've seen in many many many places inside of NX).

General rule to follow, if you aren't going to actually add any documentation... then do not add empty javadoc tags.  Its not like w/o these tags that javadoc won't actually report this method has parameters.  This is just extra bytes/lines of code which add zero value.  Please avoid.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -24,42 +29,206 @@
...
+            this.message = checkNotNull( message );
+            this.resourceStoreRequest = checkNotNull( resourceStoreRequest ).cloneAndDetach();
+        }
+
+        /**
+         * Returns the reason message.
+         * 
+         * @return the reason message.
+         */
+        public String getMessage()
```
jdillon, line 125: This is a good example of bad documentation.  The docs here add zero value, as any sane developer should assume a "String getMessage()" method would return a message.  Documentation would really only serve here to explain any reason why it wouldn't return a string.  And either way, there is no reason you document @return here as the method description covers it.  This is just extra wasted bytes to maintain which add zero value.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -94,6 +265,7 @@ public ItemNotFoundException( final ResourceStoreRequest request, final Reposito
      * @param request
      * @param repository
      * @param cause
+     * @deprecated Use constructor with {@link ItemNotFoundReason} instead.
      */
     public ItemNotFoundException( final ResourceStoreRequest request, final Repository repository, final Throwable cause )
```
jdillon, line 264: @Deprecate and @deprecate should come in pairs... the later to explain why the former was used.

jdillon:
-0.333 for lack of @since on new public api

-0.333 for the poor use of javadoc

-0.334 for the formatting changes mixed in which made this even more difficult to comprehend the actual changes

:-P

nexus-core/pom.xml
cstamas, line 62: Given that introducing goodies to core is not quite simple (as it has implications for aforementioned plugins at least), maybe it deserves a separate (cleaner and simpler to review) pull first?

jdillon, line 62: yes probably

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepository.java
```diff
@@ -976,19 +975,6 @@ protected DownloadMirrorSelector openDownloadMirrorSelector( ResourceStoreReques
...
-        boolean shouldCache = true;
-
-        // ask request processors too
-        for ( RequestProcessor processor : getRequestProcessors().values() )
-        {
-            shouldCache = processor.shouldCache( this, item );
-
-            if ( !shouldCache )
-            {
-                return item;
```
ifedorenko, line 985: Did you remove this on purpose or by accident? If on purpose, why?

ifedorenko:
Still not there but definitely getting closer. 

Before I can agree with this change,
- I want to understand if behaviour change in AbstractProxyRepository was on purpose or not (see inline question above)
- Decide if we want to maintain backwards compatibility for RequestProcessor, I personally think it's not worth it, but lets see what others say about it
- Undo all formatting-only changes to make this pull smaller and easier to review. The easiest way to do this is probably squash all commits into one, then undo all formatting-only changes and amend, but feel-free to do compensating commit.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractProxyRepository.java
cstamas, line 985: Yes, as newly introduced RequestStrategy does not have this method anymore, and I did not want to leave RequestProcessors (those deprecated ones) still littered here.

Also, as I checked, no implementation of RequestProcessor was doing anything else with this method, they never returned `false` here.

Finally, even if they _would return false_ here, I am completely sure it would just cause problems, similar to https://github.com/sonatype/nexus/pull/820 HTTP connection leak, as the RRS items (those wrapping HTTP response body) would bubble up all way to the REST layer, where in case of a `?describe` requests -- and sure for any non-200 OK response too -- ContentLocators are just tossed away ( = leak same as NEXUS-5468 in fixed in pull above).

ifedorenko:
-1

As discussed in email, there is no need to provide backwards compatibility for existing RequestProcessor implementations. Please replace  RequestProcessor with RequestStrategy and remove code related to RequestProcessor backwards compatibility.

nexus-core/src/main/java/org/sonatype/nexus/proxy/ItemNotFoundException.java
```diff
@@ -25,42 +30,200 @@
...
+        {
+            return repository;
+        }
     }
 
+    private final ItemNotFoundReason reason;
+
     /**
-     * Do not use this constructor!
+     * Constructor.
```
ifedorenko: I think the comment should read something like

   Constructor of class ItemNotFoundException written in java programming language

(seriously, though, please either provide meaningful javadoc or don't provide any at all)

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -267,14 +275,44 @@ public RepositoryTaskFilter getRepositoryTaskFilter()
...
+    }
+
+    @Override
+    public Map<String, RequestStrategy> getRegisteredStrategies()
     {
-        if ( requestProcessors == null )
+        synchronized ( requestStrategies )
         {
-            requestProcessors = new HashMap<String, RequestProcessor>();
+            return Maps.newHashMap( requestStrategies );
```
ifedorenko, line 304: Need to return unmodifiable map

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -267,14 +275,44 @@ public RepositoryTaskFilter getRepositoryTaskFilter()
...
         }
+    }
 
-        return requestProcessors;
+    /**
+     * This method is used to get the map of {@link RequestStrategy}s to evaluate.
+     * 
+     * @return
+     */
+    protected Map<String, RequestStrategy> getEffectiveRequestStrategies()
```
ifedorenko: This method is now redundant, please remove it.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -1287,40 +1310,33 @@ protected boolean checkConditions( ResourceStoreRequest request, Action action )
...
     }
 
-    protected boolean checkPostConditions( final ResourceStoreRequest request, final StorageItem item )
-        throws IllegalOperationException, ItemNotFoundException, AccessDeniedException
+    protected void checkPostConditions( final ResourceStoreRequest request, final StorageItem item )
+        throws IllegalOperationException, ItemNotFoundException
     {
-        if ( getRequestProcessors().size() > 0 )
+        final Map<String, RequestStrategy> effectiveRequestStrategies = getEffectiveRequestStrategies();
+        if ( !effectiveRequestStrategies.isEmpty() )
```
ifedorenko: Why do you need !isEmpty check here? It only adds noise, for-each loop handles empty collections already.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+     * with reason why. Otherwise, a clean return from the method is needed.
+     * 
+     * @param repository from which the item is about to be attempted retrieval (not null)
+     * @param request retrieval request (not null)
+     * @param action
+     * @throws ItemNotFoundException
+     * @throws IllegalOperationException
+     */
+    void onHandle( Repository repository, ResourceStoreRequest request, Action action )
+        throws ItemNotFoundException, IllegalOperationException;
```
ifedorenko, line 50: It is not clear in what cases this method will be called and it expected behaviour if "item should not be handled", i.e. the difference between ItemNotFoundException and IllegalOperationException needs to be explained.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+     * the client, it should throw some exception with reason why. Otherwise, a clean return from the method is needed.
+     * 
+     * @param repository from which the item is retrieved (not null)
+     * @param request retrieval request (not null)
+     * @param item item to be retrieved (not null)
+     * @throws ItemNotFoundException
+     * @throws IllegalOperationException
+     */
+    void onServing( Repository repository, ResourceStoreRequest request, StorageItem item )
+        throws ItemNotFoundException, IllegalOperationException;
```
ifedorenko, line 69: It is not clear in what cases IllegalOperationException should be thrown. Are there reasons to have this exception here at all? This method either allows serving of the artifact or throws ItemNotFoundException that explains the reason why the artifact should not be served.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+     * current request.
+     * 
+     * @param repository
+     * @param request
+     * @param context
+     * @throws ItemNotFoundException
+     * @throws IllegalOperationException
+     */
+    void onRemoteAccess( ProxyRepository repository, ResourceStoreRequest request )
+        throws ItemNotFoundException, IllegalOperationException;
```
ifedorenko, line 98: Same as above, IllegalOperationException appears redundant.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+import org.sonatype.nexus.proxy.access.Action;
+import org.sonatype.nexus.proxy.item.StorageItem;
+
+/**
+ * A Request Strategy that is able to process/modify the request, and change how Nexus core handles the request.
+ * 
+ * @author cstamas
+ * @since 2.5
+ */
+public interface RequestStrategy
```
ifedorenko, line 29: Why did you decide to rename RequestProcessor to RequestStrategy? Both names are too generic to explain the purpose of this interface, so I am not sure why change one poor name with another equally poor name.

ifedorenko:
-1

See few smaller/cosmetic comments inline

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
cstamas, line 304: It returns a _copied_ (modifiable) map, a detached "snapshot" of strategies in a moment of method invocation. Javadoc clearly states that:
https://github.com/sonatype/nexus/pull/726/files#L14R381

Unsure why you think it need unmodifiable map...

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
cstamas, line 29: Well, I did not rename RequestProcessor (at least not in events sequence) on purpose, it is the result of recent email exchanges we had (implicitly "gained"). 

The decision to _remove/break_ RequestProcessor happened today morning (my time), but before that decision, I had introduced RequestProcessor2 (initially very badly named, on Wed I renamed it to RequestStrategy) in parallel with keeping but deprecating ReqProcs.

As there was multiple problems with RP interface (badly named methods, inconsistent method signatures and such), but also how they got registered and unregistered to repository (was not thread safe and too error prone), I decided to introduce a "clean slate" _new_ RequestProcessor2 (again, later renamed to Strategy), while keeping backward compat (again, all this happened earlier this week, way before the email for 1/2/3 options was fired on list).

Now, that I applied the option 3, and removed RequestProcessors and support for backward compat, I can rename the RequestStrategy back to RequestProcessor if we see that change as really needed... I did not see it important. Either way, we broke backward compat.

ifedorenko:
+1

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RequestStrategy.java
jdillon, line 29: names are not horrible... not sure there is anything better, at least I can't think of anything better atm.

PR #728, opened 2013-01-23
WIP: Integrate requirejs into Ext.define

jdillon:


nabcos:
Already pushed into master.

PR #734, opened 2013-02-01
Add common place for core + plugin guice modules

jdillon:
... help keep these modules in sync by having them very close to each other.


adreghiciu:
+1

cstamas:
+1

PR #740, opened 2013-02-06
Allow modern and legacy component lookup

jdillon:


jdillon:
ci passed, this should be pretty safe:  https://bamboo.zion.sonatype.com/browse/NXO-OSSF5-1

PR #742, opened 2013-02-09
WIP: Hc4 conn pool per repo

ifedorenko:
Just a place to keep discussion for now. DO NOT MERGE YET.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
@@ -12,45 +12,69 @@
  */
 package org.sonatype.nexus.apachehttpclient;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.http.conn.ClientConnectionManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
 
 /**
  * Low priority daemon thread responsible to evict connection manager pooled connections/
- * 
+ *
  * @author cstamas
  * @since 2.2
  */
 class EvictingThread
     extends Thread
 {
+
     private static final Logger LOGGER = LoggerFactory.getLogger( EvictingThread.class );
 
-    private final ClientConnectionManager clientConnectionManager;
+    private final Collection<ClientConnectionManager> clientConnectionManagers;
 
     private final long idleTimeMillis;
 
     private final long delay;
 
-    EvictingThread( final ClientConnectionManager clientConnectionManager, final long idleTimeMillis, final long delay )
+    EvictingThread( final long idleTimeMillis, final long delay )
     {
         super( "HC4x-EvictingThread" );
         Preconditions.checkArgument( idleTimeMillis > -1, "Keep alive period in milliseconds cannot be negative." );
-        this.clientConnectionManager = Preconditions.checkNotNull( clientConnectionManager );
+        clientConnectionManagers = Lists.newCopyOnWriteArrayList();
         this.idleTimeMillis = idleTimeMillis;
         this.delay = delay;
         setDaemon( true );
         setPriority( MIN_PRIORITY );
     }
 
-    EvictingThread( final ClientConnectionManager clientConnectionManager, final long idleTimeMillis )
+    EvictingThread( final long idleTimeMillis )
     {
-        this( clientConnectionManager, idleTimeMillis, 5000 );
+        this( idleTimeMillis, 5000 );
+    }
+
+    /**
+     * Registers a connection manager.
+     *
+     * @param clientConnectionManager to be registered
+     */
+    public void register( final ClientConnectionManager clientConnectionManager )
+    {
+        clientConnectionManagers.add( checkNotNull( clientConnectionManager ) );
+    }
+
+    /**
+     * Unregisters a connection manager.
+     *
+     * @param clientConnectionManager to be unregistered
+     */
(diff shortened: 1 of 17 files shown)


nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
```diff
@@ -12,45 +12,69 @@
...
+    }
+
+    /**
+     * Unregisters a connection manager.
+     *
+     * @param clientConnectionManager to be unregistered
+     */
+    public void unregister( final ClientConnectionManager clientConnectionManager )
+    {
+        clientConnectionManagers.remove( checkNotNull( clientConnectionManager ) );
```
ifedorenko, line 77: This relies on undocumented ClientConnectionManager#equals implementation. It may be possible to replace #remove with iterator and == comparison, but I am not sure it is possible to get proper synchronization behaviour with CopyOnWriteArrayList in this case. I'd probably go with ArrayList and synchronized {} blocks around use to keep things simple.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
```diff
@@ -12,45 +12,69 @@
...
+    }
+
+    /**
+     * Registers a connection manager.
+     *
+     * @param clientConnectionManager to be registered
+     */
+    public void register( final ClientConnectionManager clientConnectionManager )
+    {
+        clientConnectionManagers.add( checkNotNull( clientConnectionManager ) );
```
ifedorenko, line 67: I'd check for duplicate managers

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/RemoteStorageContextUpdatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.sonatype.nexus.proxy.repository.ProxyRepository;
+import org.sonatype.plexus.appevents.AbstractEvent;
+
+/**
+ * The event that is fired every time remote storage context is updated.
+ *
+ * @since 2.4
+ */
+public class RemoteStorageContextUpdatedEvent
```
ifedorenko, line 25: What's the purpose of this event? I don't see it's used anywhere.

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/AbstractRemoteRepositoryStorage.java
```diff
@@ -152,6 +157,7 @@ protected synchronized RemoteStorageContext getRemoteStorageContext( ProxyReposi
                 updateContext( repository, ctx );
                 ctx.putContextObject( CONTEXT_UPDATED_KEY, Boolean.TRUE );
                 repositoryContexts.put( repository.getId(), ctx.getGeneration() );
+                eventBus.post( new RemoteStorageContextUpdatedEvent( this, repository ) );
```
ifedorenko, line 160: It sucks we have to change entire AbstractRemoteRepositoryStorage class hierarchy to add single injected member. Guess constructor injection has downsides too.

ifedorenko:
I had a very cursory look at PoolingClientConnectionManager implementation. It appears to have necessary hooks to plug custom ClientConnectionOperator implementation, which I think allows per repository keystore and other customizations. This will allow us cap total number of connection in the pool, which is probably not a big deal, given that we don't expect nexus to have more then one or two connections per repository. The bigger question if this will make implementation simpler and easier maintain over time. Did you look at using ClientConnectionOperator?

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
adreghiciu, line 67: Evicting thread is an internal class not supposed to be called outside of HC4Provider. And this providers only calls the add when a new conn manager is created.
Even if someone would call and we get a connection manager more times in the list, the effect would be that the eviction is called more often on that conn pool.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
adreghiciu, line 77: CopyOnWriteArrayList should be sufficient looking for this as we do not really care if during one traversal to evict connections, new conn managers will be added. They will be evicted next time.

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/RemoteStorageContextUpdatedEvent.java
adreghiciu, line 25: In this event we react in ssl plugin and we change the connection manager scheme for https.

adreghiciu:
I saw that class while I was looking for alternative solutions to changing the HC4Provider. Did not look in depth, yet I will do that now.

adreghiciu:
I could not figure out how ClientConnectionOperator would help us. The idea is that we need to link the socket factory to a repository. The ClientConnectionOperator is one / connection manager so when using a shared connection manager there is no way to relate a connection to the repository making the connection.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/EvictingThread.java
ifedorenko, line 77: My point was, clientConnectionManagers.remove uses ClientConnectionManager#equals and we don't know how that method is implemented for all ClientConnectionManager we'll have to work with. Depending on ClientConnectionManager#equals implementation the code above may be removing wrong connection manager from the list, it removes first equal instance found, IIRC, which may not be the instance we want to remove.

ifedorenko:
This pull is no longer relevant, closing.

PR #743, opened 2013-02-10
Some improvements to the core NX.* js bits

jdillon:
Adds support for mixins.
Fixes issue with NX.define in a complex hierarchy.
Class and class prototype are now $className and $simpleClassName aware.
Objects get a slightly better toString()
Re-implemented NX.log, has better control over how levels are filtered.
Added LogAwareMixin which provides log{Debug|Warn|Info|Error} functions to objects to replace the $log usage.
LogAwareMixin has some runtime control over the format includeName, includeLevel, simpleName, etc.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/access/DefaultNexusItemAuthorizer.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/access/DefaultNexusItemAuthorizer.java
@@ -131,49 +131,32 @@ public class DefaultNexusItemAuthorizer
         return perms;
     }
 
-    protected boolean isPermitted( final List<String> perms )
-    {
-        if ( securitySystem.isSecurityEnabled() )
-        {
-            // Get the current user
-            final Subject subject = securitySystem.getSubject();
-            if ( subject != null ) // TODO: isn't this missing? "&& subject.isAuthenticated()"
-            {
-                if ( getLogger().isDebugEnabled() )
-                {
-                    getLogger().debug( "Checking isPermitted() with perms: " + perms.toString() );
-                }
-                // And finally check each of the target permissions and see if the user
-                // has access, all it takes is one
-                for ( String perm : perms )
-                {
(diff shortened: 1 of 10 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/access/DefaultNexusItemAuthorizer.java
```diff
@@ -131,49 +131,32 @@ public boolean isViewable( final String objectType, final String objectId )
...
+            if (subject.isPermitted(perm)) {
+                // TODO: we should remember/cache these decisions per-thread and not re-evaluate it always from Security
+                getLogger().debug("Subject '{}' has permission: {}; allowing", subject.getPrincipal(), perm);
+                return true;
+            }
+        }
+
+        getLogger().debug("Subject '{}' is missing required permissions; rejecting", subject.getPrincipal());
+        return false;
     }
```
jdillon, line 161: Eh, not JS related but snuck into the pull... I'll weed it out.

plugins/nexus-timeline-plugin/src/main/java/org/sonatype/nexus/feeds/record/NexusAuthenticationEventInspector.java
```diff
@@ -65,7 +65,7 @@ public void inspect( Event<?> evt )
         final ClientInfo ai = nae.getClientInfo();
 
         final String msg =
-            String.format( "%s user [%s] from IP address %s", ( nae.isSuccessful() ? "Sucessfully authenticated"
+            String.format( "%s user [%s] from IP address %s", ( nae.isSuccessful() ? "Successfully authenticated"
                 : "Unable to authenticate" ), ai.getUserid(), StringUtils.defaultString( ai.getRemoteIP(), "[unknown]" ) );
```
jdillon, line 69: This snuck through too.

nabcos:
+1

PR #744, opened 2013-02-13
[NXCM-4895] Use a strategy for ClientConnectionOperator based on http host, http context and http parameters

adreghiciu:
Used by nexus ssl plugin

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
@@ -120,7 +120,7 @@ public class HttpClientRemoteStorage
     /**
      * Http context key of repository making a request.
      */
-    public static final String HTTP_CTX_KEY_REPOSITORY_ID = PROVIDER_STRING + ".repositoryId";
+    public static final String HTTP_CTX_KEY_REPOSITORY = PROVIDER_STRING + ".repository";
 
     private final QueryStringBuilder queryStringBuilder;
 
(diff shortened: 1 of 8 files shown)


adreghiciu:
CI: https://bamboo.zion.sonatype.com/browse/NXO-OSSF37-1

adreghiciu:
CI :+1:

cstamas:
:+1:

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
```diff
@@ -486,7 +492,10 @@ private HttpResponse doExecuteRequest( final ProxyRepository repository, final R
         HttpResponse httpResponse = null;
         try
         {
-            httpResponse = httpClient.execute( httpRequest );
+            final BasicHttpContext httpContext = new BasicHttpContext();
+            httpContext.setAttribute( HTTP_CTX_KEY_REPOSITORY_ID, repository.getId()  );
```
ifedorenko: Really minor, but do you think it makes more sense to pass repository object instead of reposito id in the context?

ifedorenko:
+1

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
adreghiciu: Makes sense. I'm yet not sure if we should still be putting it at all in the context, because we do not use it anymore in SSL. I left it there just in case. 
Yet I agree is better to have the repository so we do not have to look it up again.

PR #745, opened 2013-02-13
Catch-All fix

cstamas:
Removal of "special catch-all" permission, and doing checks instead of PlexusResources on the fly.

### DIFF
--- a/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/NexusApplication.java
+++ b/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/NexusApplication.java
@@ -14,6 +14,7 @@ package org.sonatype.nexus.rest;
 
 import java.util.List;
 
+import org.apache.shiro.util.AntPathMatcher;
 import org.codehaus.plexus.component.annotations.Component;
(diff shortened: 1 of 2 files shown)


adreghiciu:
+1

jdillon:
Looks okay to me :sake:

cstamas:
Just FTR: checked Shiro Filter implementation and here what happens: resources are "handled" as they are discovered (by Plexus collection lookup) as this:

```
NexusApplication#handlePlexusResourceSecurity(PlexusResource) calls
org.sonatype.security.web.DefaultProtectedPathManager calls
org.apache.shiro.web.filter.mgt.DefaultFilterChainManager.class calls
org.apache.shiro.web.filter.PathMatchingFilter
```

Finally, Shiro in `PathMatchingFilter` keeps path patterns from descriptors as plain String in a map. On incoming request, it uses this same `org.apache.shiro.util.AntPathMatcher` to match the pattern in a map against `request.getRequestURI()` (with some spice). Meaning, we should get the very same results wrt this sanity check as Shiro.

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF38-1

nabcos:
+1

cstamas:
CI +1'd it, merging it.

PR #747, opened 2013-02-14
[NEXUS-5434] Update to ExtJS 3.4.1

nabcos:
Nexus Pro Dist built from this branch:
https://docs.google.com/a/sonatype.com/file/d/0B4rXJy46Zo2WcldjZldQMjA3em8/edit


jdillon:
Are their updates for our marriage of requirejs w/define/create/etc ?

nabcos:
It's possible to use the stock Ext.define in 3.4.1 in our NX.define. That does not give us much, though. callParent is mostly broken, because the parent a class inherits from has to be defined with Ext.define as well (AFAICT none of the classes in Ext have been updated to do that).

I quickly tried to lift code from them to give methods in NX.defined classes and used superclasses an $owner (which is used to determine the parent for a method), but the scope wasn't right when it was actually used. Apart from replacing NX.Base with Ext.Base, I don't see much value by using Ext.define inside NX.define.

jdillon:
Okay, fair enough.  I think we will try to keep NX.define Ext.define v4 compatible anyways.

So callParent() in 3.4.1 is not working properly?

nabcos:
IIUC callParent only works when you define the method you override with Ext.define:

```
Ext.define('A', {
    extend : 'Ext.Panel',
    aMethod : function() {
        console.log('A');
    },
    doLayout : function() {
        this.callParent(arguments);
    }
}

Ext.define('B', {
    extend : 'A',
    aMethod : function() {
        this.callParent(arguments);
    }
}

// log "A"
new B().aMethod();

// won't work because Ext.Panel was not defined with Ext.define
new A().doLayout();
```

nabcos:
I assume we could make it work by doing the proper analysis in NX.define, adding the proper $owner for every method in the prototype chain - not sure it's worth it for our use cases, though

nabcos:
Already merged to master.

PR #748, opened 2013-02-15
[NEXUS-5554] Reorganize NX namespace, allow plugins to specify require.js module/dependencies directly

nabcos:
Use requirejs for most of what was in lib and move to a different namespace.
Introduce support for plugins to specify their js dependencies directly.


jdillon:
The namespace took me by surprise a little, but I think after looking this over for a bit that its all good ;-)

+1

nexus-core/src/main/java/org/sonatype/nexus/plugins/rest/RequireJsContributor.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * where plugin-all.js contains require.js module definitions (e.g. `define("Nexus/plugin/bootstrap", ["Nexus/plugin/dep1"], function() {}` ).
+     */
+    public static class RequireJsContribution {
+
+        private List<String> dependencies;
+
+        private String module;
+
+        public RequireJsContribution( final String module, final List<String> dependencies )
```
jdillon: If this is expected to be coded up like you have done here:

https://github.com/sonatype/nexus-ssl/commit/a55033a35420b34f60f4ce794fdc32d46d2bfd53

Then you should probably make String... dependencies helper to avoid needing to have the implementation make lists.

Though really I think a builder model here is better, since I'm guessing a lot of plugins will want to have something like you put in nexus-ssl:

```
return new RequireJsContribution( "Nexus/ssl/bootstrap", ImmutableList.of(String.format("static/js/%s-all.js", ARTIFACT_ID)) );
```

And it would be good to simply just have a method which put a dependency on static/js/%s-all.js w/ an artifact id.  Leaving this up to the plugin will invariably leave us with plugins putting sources in the wrong place.  I recommend we hide all of that from the implementation of a require contributor.

Also, keep in mind ExtJS 4.1, whatever new plugin support we add should also be able to translate into ExtJS 4.1 w/o any requirejs.  So I might recommend a better name for this, and cook in support for requirejs and probably add in support to just require a js class (and then translate it to requirejs semantics under the covers).

jdillon:
+1 once its mergable

PR #749, opened 2013-02-15
Stable timestamp of generated js/css urls

ifedorenko:
Derive timestamp from jar file lastModified (for packaged plugins)
or actual resource file lastModified (for exploded plugins). Fall
back to current time if referenced resource file cannot be detected.


jdillon:
looks okay to me :tangerine:

kellyrob99:
+1
Should using this be the default for new plugins, and is it worth updating existing plugins to use?

ifedorenko:
Yes, I think IndexHtmlSnippetBuilder is useful and we should eventually update existing plugins to use it.

ifedorenko:
Merged the commit directly, but kept backreference to this pull request in the commit message. I think I like it. It gives us clean commit history on master and the discussion place pull requests provide. https://github.com/sonatype/nexus/commit/3045969daf031d24ee0f52b63b8fc47998cad52e

PR #751, opened 2013-02-19
Fix for NPE during version range resolution.

thallgren:
The NexusWorkspaceReader returned null on findVersions() instead of
an empty collection. This is in violation with the javadoc and
resulted in NPE's during version range resolutions.


cstamas:
+1

adreghiciu:
+1

PR #753, opened 2013-02-20
Subordinate locking and improvement for NEXUS-5418

cstamas:
This change introduces a possibility left to Repository implementors to interfere (and probably modify) the UID key to UIDLock key mapping. Before this change, UID Key was used for UIDLock key as is, meaning, no different UID was using same UIDLock instance.

In this change, Maven repositories would lock artefacts and their checksums "as one", irrespective do you operate with artefact or checksum UID's (or requests or items). Simply put, this change makes Maven repositories always use the artefact lock for both, for artefacts and for their checksums.

Since this change reduces the involved locks (in case of Maven repositories) from 2 to 1, this also avoids possible deadlocks for concurrent requests tampering with same artefact and it's checksum.

Note: sorry for reformat in AbstractRepository.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/item/AbstractRepositoryItemUidFactory.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/item/AbstractRepositoryItemUidFactory.java
@@ -12,6 +12,8 @@
  */
 package org.sonatype.nexus.proxy.item;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.lang.ref.WeakReference;
 import java.util.WeakHashMap;
 
@@ -49,7 +51,7 @@ public abstract class AbstractRepositoryItemUidFactory
             path = RepositoryItemUid.PATH_ROOT;
         }
 
-        return new DefaultRepositoryItemUid( this, repository, path );
+        return new DefaultRepositoryItemUid( repository, path );
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


cstamas:
Another use of subordinate locking might be P2 repository, where `artifacts.xml` and `content.xml` is handled "as one" (updated as one), and both might be using UIDLock with key "repoId:p2-repo-metadata" or such.

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF16-1

cstamas:
After fixing forgotten UT, CI +1'd this:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF16-2

cstamas:
New CI with M2 related fix and UT:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF16-3

ifedorenko:
Do not merge this until I have the time to review this.

ifedorenko:
-1

This change does not appear to fix anything. I rebased subordinate-locking branch on current master and ran a simple stress tests when one set of threads were deploying pom/sha1/md5 to a snapshot repository and another were downloading the pom/sha1/md5. With or without this patch, I see see large number of FileNotFoundException for sha1/md5 checksums and download threads get HTTP 500 response from the server.

And generally, given overall state of item uid locking implementation, I am extremely uncomfortable synchronizing access to related but separate repository items. The code is way too convoluted to be able to assess all implications of this change.

PR #754, opened 2013-02-21
[NXCM-4995] Minimal fix to make logging work in IE

nabcos:
Created [NEXUS-5557](https://issues.sonatype.org/browse/NEXUS-5557)  to track logging framework adoption.


peterlynch:
I'd say go for it

jdillon:
I'd like to see some more comments on these blocks to explain briefly what its doing, more for general comprehension, as I already understand the problem.  But someone new looking at this might not have any idea... and the '"IE mode", pretty messed up' comment, which is the only comment I see probably won't help too much :-P

Otherwise +1 if its working in IE again.

PR #756, opened 2013-02-22
[NXCM-4895] Introduce SmtpSessionParametersCustomizer for allowing customisation of SMTP session parameters

adreghiciu:

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
@@ -12,8 +12,10 @@
  */
 package org.sonatype.nexus.email;
 
+import org.codehaus.plexus.PlexusContainer;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Startable;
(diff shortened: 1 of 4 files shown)


adreghiciu:
CI build: https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-1

nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
```diff
@@ -55,6 +58,9 @@
     @Requirement
     private EMailer eMailer;
 
+    @Requirement
+    private PlexusContainer plexus;
```
nabcos: better convert to `@Named` and use method injection or `Provider<EMailerConfiguration>`?

ifedorenko:
I really dislike plexus.lookup/catch(ComponentLookupException). I think NexusEmailerConfiguration does not need to be a component, it should be plain class with package visibility.

Otherwise +1.

adreghiciu:
CI :+1:

jdillon:
-1

We should not be introducing anymore more use of PlexusContainer... we should _ONLY_ be removing usage of this.

adreghiciu:
@ifedorenko Changed to use "new"

adreghiciu:
@jdillon did not want to convert this yet to sisu, but maybe we should do it

jdillon:
@adreghiciu if we come across simple plexus components when adding/fixing features we should try and convert them.  Or we'll never finish the transition from plexus to guice.

Anyways, my main beef here was with use of PlexusContainer directly.  We _MUST NOT_ use this anymore.

PR #761, opened 2013-03-04
[NXCM-4999] Use in-repo path for matching repo targets

nabcos:
This code path was still using "/repositories/$name/$path" to match
repository target patterns. Repository Browsing worked fine, but archive browser triggered this
bug and lead to not using repo targets the same way:

`(?!/com/.*-sources.jar).*` would deny access to the source jar itself but
would allow access to its contents, while `(?!.*/com/.*-sources.jar).*`
would also deny content access.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/router/DefaultRepositoryRouter.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/router/DefaultRepositoryRouter.java
@@ -729,7 +729,7 @@ public class DefaultRepositoryRouter
                 // if this repository is contained in any group, we need to get those targets, and tweak the TargetMatch
                 try
                 {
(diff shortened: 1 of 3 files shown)


nabcos:
CI: https://bamboo.zion.sonatype.com/browse/NXO-OSSF25-1

ifedorenko:
Is this hard to unit test? (and I mean "unit", I don't want full blown integration test here).

cstamas:
Change is correct, added UT verifying it.

cstamas:
2nd CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF25-2

cstamas:
CI passed (with two WL related ITs failing, that was fixed in https://github.com/sonatype/nexus/pull/765 )

PR #768, opened 2013-03-08
NEXUS-5542: Update index task fails if one member gets a SocketTimeoutException

cstamas:
Similar to NEXUS-5249 but not same. This is about CASCADE operations, when one single group is being processed.

Just applied same stuff already in place from NEXUS-5249 to places doing cascade. Currently the task will fail, if any member fails, but the processing will not be interrupted, all the members will be processed. If other behaviour needed, this pull will need more.

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF3-1


ifedorenko:
+0

I generally agree with the solution, but I think the code can be improved by pulling perform-same-operation-on-all-group-members logic to a template method, see untested proposed change in https://github.com/sonatype/nexus/commit/6cba5fd7ba9119d5a5eacd17b3bbf30486b73ba6

jdillon:
agree looks like some duplication could be avoided with some simple abstraction.

cstamas:
Consumed igor's proposed change.

New CI (with template method):
https://bamboo.zion.sonatype.com/browse/NXO-OSSF3-2

cstamas:
CI +1'd template method too
https://bamboo.zion.sonatype.com/browse/NXO-OSSF3-2

PR #769, opened 2013-03-09
(NEXUS-5594) handle RepositoryNotAvailableException and avoid error log,...

jdillon:
... as this could happen normally.

https://issues.sonatype.org/browse/NEXUS-5594


jdillon:
This appears to resolve the nasty logging and I'm pretty sure this has no other effects.

ifedorenko:
+1

The change looks safe.

kellyrob99:
+1

PR #770, opened 2013-03-10
Nexus 5418 store checksum in item attributes

ifedorenko:
I could not figure out how to convince bamboo to build this branch :-(

PR #771, opened 2013-03-10
23x nexus 5418 store checksum in item attributes

ifedorenko:
This is 2.3.x version of https://github.com/sonatype/nexus/pull/770 . I decided not to remove item uid #getAttributeLock method, but changed implementation to delegate to #getLock. There are also minor differences due to changes in RepositoryRequest, but otherwise the two pulls are equivalent.

### DIFF
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUid.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUid.java
@@ -86,16 +86,11 @@ public class DefaultRepositoryItemUid
 
         return lock;
     }
-    
+
     @Override
     public synchronized RepositoryItemUidLock getAttributeLock()
     {
-        if ( lock == null )
-        {
-            lock = factory.createUidAttributeLock( this );
-        }
-
-        return lock;
+        return getLock();
     }
 
     @Override
--- a/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/RepositoryItemUid.java
+++ b/nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/RepositoryItemUid.java
@@ -95,6 +95,8 @@ public interface RepositoryItemUid
      * As you can see on {@link RepositoryItemUidLock#unlock()} method javadoc, last of the boxed unlocks (if any boxing
      * at all, simply last invocation) also releases the lazily created lock.
      * 
+     * @deprecated this method is fully equivalent to {@link #getLock()}
+     * 
      * @return
      */
     RepositoryItemUidLock getAttributeLock();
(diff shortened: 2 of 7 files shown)


ifedorenko:
Bamboo is happy with this branch https://bamboo.zion.sonatype.com/browse/NXO-OSSFET2429-1

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUid.java
```diff
@@ -86,16 +86,11 @@ public synchronized RepositoryItemUidLock getLock()
...
     @Override
     public synchronized RepositoryItemUidLock getAttributeLock()
     {
-        if ( lock == null )
-        {
-            lock = factory.createUidAttributeLock( this );
-        }
-
-        return lock;
+        return getLock();
```
cstamas, line 93: Word of warning: using item lock instead of attribute lock _may_ serialize serving up items even from a hosted repository, as Nexus does _write_ to attributes things like last requested etc even for RO requests (ie. HTTP GET). How locks are "upgraded" (from shared to exclusive) means that when attribute is updated would result in upgraded lock, held until the very end of request. See NEXUS-4628

This could be a non-issue (I'd imagine it could affect a "swarm" building same thing, requesting same things like a CI), but still, I'd like to see some proof this does not affect (very badly) Nexus perf.

cstamas:
How is this change meant to cope with "Rebuild Maven Metadata Files" task? It will try to explicitly save checksum file(s), and I see no store handling of checksums.....

nexus/nexus-api/src/main/java/org/sonatype/nexus/proxy/item/DefaultRepositoryItemUid.java
ifedorenko, line 93: As I mentioned in the commit message, both #getLock and #getAttributeLock were already using the same lock (notice the same #lock member is returned by both methods). There is no change in lock escalation behaviour.

It maybe possible that some operations were only touching attribute lock, but in this case only one lock is involved and using item lock in these operations does not introduce new deadlock.

ifedorenko, line 93: What kind of "proof" there is no performance regression do you have in mind? The stress test I used to reproduce original FileNotFoundException covered deploy to hosted snapshot repository, smartproxy with preemptive fetch and download from snapshot proxy, and showed comparable request-per-minute performance, with or without this change. Anything else we need to test?

ifedorenko:
Added the logic necessary to intercept storeItem request for Maven proxy repositories. This should take care of  "Rebuild Maven Metadata Files Task"  use case. Note that I plan to squash together the two commits that implement "store sha1/md5 checksums in item attributes" change before I merge this pull request. I keep them separately to simplify code review.

Bamboo is happy with this branch https://bamboo.zion.sonatype.com/browse/NXO-OSSFET2429-2

cstamas:
This with store support now looks good. +1

PR #775, opened 2013-03-14
Close InputStream and Reader where needed

gaul:


ifedorenko:
-1

What is the problem you are trying to solve? Are you getting out-of-file handles or similar exceptions indicating open stream leaks? Although I agree Xpp3DomBuilder API is far from clear, Xpp3DomBuilder.build actually closes Reader passed in as the parameter. I am actually surprised you are not getting "stream already closed" exceptions with this change.

PR #776, opened 2013-03-15
PENDING: add local.storage.timing logger

kellyrob99:
https://issues.sonatype.org/browse/NXCM-4855
NXCM-4855 - detailed logs about outgoing HTTP request and requests to local storage
Follows the logic already in place for remote.storage.timing logger

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/AbstractLocalRepositoryStorage.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/AbstractLocalRepositoryStorage.java
@@ -21,6 +21,9 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 
+import com.google.common.base.Stopwatch;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.sonatype.nexus.logging.AbstractLoggingComponent;
 import org.sonatype.nexus.mime.MimeSupport;
 import org.sonatype.nexus.proxy.ItemNotFoundException;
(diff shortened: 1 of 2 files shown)


jdillon:
+0 generally I think almost all timing related activities are better handled via AOP... instead of mudding up the codebase with lots of stopwatch related code which is ancillary to the components logic.

kellyrob99:
@jdillon So in a Guice world that would be implemented using org.aopalliance.intercept.MethodInterceptor for AOP?

ifedorenko:
+1

I agree this adds some unwanted noise to the code, but I am not sure method interceptors will result in much cleaner implementation. Specifically, log should contain information about request repository and path (and execution time, of course), which I think implies that interceptor needs a way to access method parameters, which are different for different methods, i.e. Repository/ResourceStoreRequest vs Repository/StorageItem. Also, I am not sure how reusable interceptors like this can be, which I believe significantly reduces AOP appeal.

jdillon:
I think we can probably figure out a fairly simple and elegant solution here.

I also think javasimon should be used over guava stopwatch, since its jmx aware and provides more metrics (avg times, etc).

I believe before we go adding anymore timing related stuff that we regroup and come up with the basic requirements and design a proper solution around them.  I have some ideas on how an AOP mechanism might work, though its only based on some rough thoughts I've got on what the requirements are ATM.

I only see the need for timing various parts of NX to grow, so I think its worth the time to put together something more sustainable now before we end up with more complex stopwatch blocks throughout the codebase.

ifedorenko:
Ok, lets wait with this change, it is not critical for 2.4 and it is better to come up with clean(er) longer term solution if we can.

jdillon:
May want to see if we can make use of this: 

http://metrics.codahale.com/manual/guice/

... which means integrating yammer metrics, which has lots of other nice goodies like healthchecks and such.

Blah, looks like this (and some other support) was removed after 2.2.0: https://github.com/codahale/metrics/commit/e058f76dabf3f805d1c220950a4f42c2ec605ecd

Though its still up here: https://github.com/palominolabs/metrics-guice

PITA... still looks like what we want/need though, so maybe we have to Sisu this up (if we haven't already).

jdillon:
Related to yammer metrics use: https://github.com/sonatype/nexus/pull/779

PR #777, opened 2013-03-18
NEXUS-5525: Custom packaging types do not show up in index after deployment

cstamas:
Tracking down the problem revealed a lack of changes when MI
indexer-artifact was moved out from core. Since 2.2 or 2.1,
when indexer-artifact (part of MI) was moved out from core (indexer-core
was moved even earlier), the package mapper had to be moved into Maven
Support of Nexus as other parts of Nexus did rely on it. But, MI internal components
still relied on the package mapper from MI, but since that move, nothing inited it.

But since the move, nothing was setting it up properly, hence
all the packaging mappings were left unused (not picked up)
in the process of indexing. Due to this, artifact locator
was unable to locate the artifact if POM file was given.

This change now introduces a "wrapper" plexus component that
redirects MI package mapper to Nexus package mapper (they
are basically same but in different packages).

Note: the new component had to be plexus to _replace_ the one coming from MI. Having this as SISU did not resulting in wanted effect.


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSFET243-1

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF8-1

PR #778, opened 2013-03-18
[NEXUS-5478] Handle HTML entities in role and privilege IDs

nabcos:
I'm on the fence whether this change needs an IT. It's only "transport configuration" after all.

### DIFF
--- a/nexus-webapp/src/main/webapp/js/repoServer/RoleEditPanel.js
+++ b/nexus-webapp/src/main/webapp/js/repoServer/RoleEditPanel.js
@@ -430,6 +430,7 @@ Sonatype.repoServer.DefaultRoleEditor = function(config) {
         helpText : ht.name,
         name : 'name',
         allowBlank : false,
+        htmlDecode : true,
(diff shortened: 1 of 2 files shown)


jdillon:
Why was this not reviewed by anyone? So close to a release and this is IMO non-trivial... :poop:

nabcos:
I merged this without further ado because it has only a small scope, it only affects the resources explicitely mentioned in the change. I manually tested the code paths using the UI, and was under the impression there were no ITs for this I could reach with a feature build (actually nexus-client ITs indirectly exercise this change, and those are in core).

The problem is when XML variant is used, the generated XML changed from `<roles><role>foo</role></roles` to `<roles><string>foo</string></roles>`. Using JSON is fine.

ifedorenko:
I do not believe code review would have helped finding the problem earlier. Most likely we would just spend time discussing if the change was safe enough to go in and insisted on a unit test (which may or may not spotted the problem). Most likely the code review would have delayed the merge until later staged build and so we would have found the problem later.

PR #779, opened 2013-03-19
REVIEW: Add yammer metrics ( http://metrics.codahale.com ) to NX

jdillon:
https://issues.sonatype.org/browse/NEXUS-5676

ML thread: https://groups.google.com/a/sonatype.com/d/msg/sonatype-nexus-dev-group/_lHQMlZ84Yw/vz-OEm52FaIJ


adreghiciu:
+1. IF we go further with this we should protect the exposed servlets via a filter chain as "noSessionCreation,authcBasic,perms[some nexus permission]"

kellyrob99:
+1 Do we have any concrete ideas about what metrics/healthchecks might be beneficial to track?

jdillon:
@kellyrob99 nothing concrete, some ideas up here: https://docs.sonatype.com/display/~jdillon/Nexus+Usage+Statistics

jdillon:
depends on for security:

https://github.com/sonatype/nexus/pull/800

adreghiciu:
+1

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/InstrumentedHttpClient.java
```diff
@@ -0,0 +1,74 @@
+package org.sonatype.nexus.apachehttpclient;
```
cstamas, line 13: License is missing

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/InstrumentedRequestDirector.java
```diff
@@ -0,0 +1,116 @@
+package org.sonatype.nexus.apachehttpclient;
```
cstamas, line 13: License is missing

nexus-oss-webapp/src/main/resources/content/conf/jetty.xml
```diff
@@ -24,13 +24,17 @@
 <!-- ============================================================ -->
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
     <Set name="threadPool">
-        <New class="org.sonatype.sisu.jetty.thread.InstrumentedQueuedThreadPool"/>
+        <New class="com.yammer.metrics.jetty.InstrumentedQueuedThreadPool"/>
```
cstamas, line 27: Just FTR: With this change you are removing this QTP
https://github.com/sonatype/sisu-jetty8/commit/0cf9572896d4f0cf77bbaa57c41359db86cf9586

Some context about it here:
http://dev.eclipse.org/mhonarc/lists/jetty-users/msg02350.html

And (unrelated) issue describing Greg's stance wrt OOM here:
http://jira.codehaus.org/browse/JETTY-1194

Short summary: Jetty is silent about thread death, it does not inform you in any way that a thread was removed from a pool. While this problem with P2 was a special case (one single point acquired huge amounts of heap, that is immediately put out of scope, the P2 repo metadata DOM). But the point is, that the fact that QTP thread doing working on P2 related request was not logged anywhere, there was not signs at all, simply that one single request failed, and Nexus happily continued running, everything was seemingly working just fine. 

While I do share Greg's concerns about OOM, the problem is that there is no feedback at all about OOM happened. And due to the nature of it, Nexus just continued happily working.

jdillon, line 27: @cstamas thanks for pointing this out.  Is this still relevant, or does Jetty provide proper logging now?  If you think org.sonatype.sisu.jetty.thread.InstrumentedQueuedThreadPool should be used instead, which is fine, then we can make a custom impl which captures metrics and does this strange thread wizardry.

BTW org.sonatype.sisu.jetty.thread.InstrumentedQueuedThreadPool ASIS does not really instrument anything, it just appears to use a thread factory and install a background thread... yes?

---

Also, some of the stock metrics integration classes have some minor issues which make them hard to re-use (like the httpclient bits).  We may want to fix these upstream and not need to hack around them here.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/InstrumentedHttpClient.java
jdillon, line 13: this is some duplicated/massaged code would need special handling of license here

nexus-oss-webapp/src/main/resources/content/conf/jetty.xml
cstamas, line 27: @jdillon No, Jetty does not provide anything in this area yet.

To repeat, the problem is that there is _no feedback_ (in logs, visible) that OOM happened. This QTP modification was just a "best effort" try to make Nexus (well, Jetty) start nagging you if it detected "thread loss" from pool. Simply, you (the ops running Nexus) can be completely unaware of the fact that there was an OOM in system.

After reading a bit more, am leaning towards Greg's stance, I am fine with your change (removing this customized QTP and replacing it with Yammer one), and we should just recommend putting `-XX:+HeapDumpOnOutOfMemoryError` into wrapper.conf... as that's the best you could do so far.

cstamas:
+1

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/InstrumentedHttpClient.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.http.conn.ClientConnectionManager;
+import org.apache.http.conn.ConnectionKeepAliveStrategy;
+import org.apache.http.conn.routing.HttpRoutePlanner;
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.http.params.HttpParams;
+import org.apache.http.protocol.HttpProcessor;
+import org.apache.http.protocol.HttpRequestExecutor;
+
+// NOTE: Duplicated and augmented from original 2.2.0 source to change signature of CTOR ClientConnectionManager parameter
+// NOTE: Should get this changes into metrics-httpclient and avoid needing this
```
jdillon, line 31: Change to remove the need for this is currently up here:

https://github.com/jdillon/metrics/commit/d59bbe02704e372fd87366973e7d7ae857b40374

Pending any other changes to simplify our integration before I make a pull request and lobby to get this change in.  No clue how difficult that might be.

jdillon:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF13-6 :white_check_mark:

PR #780, opened 2013-03-19
WL update misbehaving

cstamas:
This pull contains fix for multiple WL related issues: two related to WL performance, one for polluting logs and one about maintaing WL for hosted reposes.

First, group WL's were double updated in some cases (like group member addition or removal). This resulted in correct results but the "wl update" operation was done twice, redundantly.

Second, on Nexus boot groups were "inited" multiple times, redundantly. The "init loop" of WL was traversing hosted and proxy reposes only, leaving group init to happen as "side effect" of cascade operation (when a repo WL changes, it cascades to all groups where repo in question is member). But this also meant, that one group WL was updated (published, event fired, but no WL content change after 1st pass) as many times as many members it had. Now, the boot process is modified, to not use "cascade" and explicitly init group repositories too.

Third, logs were polluted on boot with M1 reposes not supported by WL feature (if you had non shadow M1 reposes in your config).

Fourth, when a parent folder was deleted (parent of a WL entry for hosted repo), the WL was not properly maintained.

Related issues:
https://issues.sonatype.org/browse/NEXUS-5602
https://issues.sonatype.org/browse/NEXUS-5608
https://issues.sonatype.org/browse/NEXUS-5618
https://issues.sonatype.org/browse/NEXUS-5623

CI:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-1

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/EventDispatcher.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/EventDispatcher.java
@@ -18,11 +18,8 @@ import java.io.IOException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.sonatype.nexus.configuration.Configurable;
-import org.sonatype.nexus.configuration.ConfigurationChangeEvent;
 import org.sonatype.nexus.proxy.RequestContext;
 import org.sonatype.nexus.proxy.events.RepositoryConfigurationUpdatedEvent;
-import org.sonatype.nexus.proxy.events.RepositoryGroupMembersChangedEvent;
 import org.sonatype.nexus.proxy.events.RepositoryItemEvent;
 import org.sonatype.nexus.proxy.events.RepositoryItemEventCache;
 import org.sonatype.nexus.proxy.events.RepositoryItemEventDelete;
@@ -31,7 +28,6 @@ import org.sonatype.nexus.proxy.events.RepositoryItemEventStore;
 import org.sonatype.nexus.proxy.events.RepositoryRegistryEventAdd;
 import org.sonatype.nexus.proxy.item.StorageFileItem;
 import org.sonatype.nexus.proxy.item.uid.IsHiddenAttribute;
-import org.sonatype.nexus.proxy.maven.MavenGroupRepository;
 import org.sonatype.nexus.proxy.maven.MavenHostedRepository;
 import org.sonatype.nexus.proxy.maven.MavenRepository;
 import org.sonatype.nexus.proxy.maven.maven2.Maven2ContentClass;
@@ -310,50 +306,6 @@ public class EventDispatcher
         {
             final MavenRepository mavenRepository = evt.getRepository().adaptToFacet( MavenRepository.class );
             handleRepositoryModified( mavenRepository );
-        }
-    }
-
-    // == Handlers for Group changes (WL of group and groups of groups needs to be updated)
-
-    /**
-     * This subscription is disabled, as this event is fired BEFORE configuration is committed, so group members are NOT
-     * discovered properly and leading to bad WL content!
-     * 
-     * @param evt
-     */
-    // @Subscribe
-    // @AllowConcurrentEvents
-    public void onRepositoryGroupMembersChangedEvent( final RepositoryGroupMembersChangedEvent evt )
-    {
-        if ( isRepositoryHandled( evt.getRepository() ) )
-        {
-            final MavenRepository mavenRepository = evt.getRepository().adaptToFacet( MavenRepository.class );
-            handleRepositoryModified( mavenRepository );
-        }
-    }
-
-    /**
-     * Workaround for method above! This tricky subscriber actually listens for group member changes, but it does it by
-     * hooking to {@link ConfigurationChangeEvent} instead of {@link RepositoryGroupMembersChangedEvent}, since former
-     * is fired AFTER configuration is committed, while latter is fired BEFORE.
-     * 
-     * @param evt
-     */
-    // @Subscribe
-    // @AllowConcurrentEvents
-    public void onConfigurationChangeEvent( final ConfigurationChangeEvent evt )
-    {
-        for ( Configurable configurable : evt.getChanges() )
-        {
-            if ( configurable instanceof Repository )
-            {
-                final Repository repository = (Repository) configurable;
-                final MavenGroupRepository mavenGroupRepository = repository.adaptToFacet( MavenGroupRepository.class );
-                if ( mavenGroupRepository != null && isRepositoryHandled( mavenGroupRepository ) )
-                {
-                    handleRepositoryModified( mavenGroupRepository );
-                }
-            }
         }
     }
 }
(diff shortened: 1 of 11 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/EventDispatcher.java
```diff
@@ -339,8 +339,8 @@ public void onRepositoryGroupMembersChangedEvent( final RepositoryGroupMembersCh
      * 
      * @param evt
      */
-    @Subscribe
-    @AllowConcurrentEvents
+    // @Subscribe
+    // @AllowConcurrentEvents
```
jdillon: Please add some additional comments as to why these are commented out. If this is not temporary, then these should be removed not commented.

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -225,10 +254,41 @@ public void shutdown()
     @Override
     public void initializeWhitelist( final MavenRepository mavenRepository )
     {
-        doInitializeWhitelist( mavenRepository );
+        getLogger().debug( "Initializing white-list of newly added {}",
+            RepositoryStringUtils.getHumanizedNameString( mavenRepository ) );
```
jdillon: Curious, why isn't the default toString() of these objects rendering "humanized" forms?  This "humanization" of the str repr would be better done there and would allow for slf4j {} optimization to be used to avoid string concat for logs like this when the level is not enabled.  Otherwise stuff like this should be guarded.

I recommend you consider changing the toString() to render what is appropriate for logging, in this case whatever the {{getHumanizedNameString()}} is doing and avoid copying this code all over the place.

cstamas: I believe you are right. This is DEBUG log, and I already added toString for reposes (eases debugging but is handy also in DEBUG logs). Will change,

cstamas:
new CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-5

cstamas:
CI +1d (after tightening a correct but unstable UT)
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-6

This pull is ready to be merged.

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -225,13 +256,42 @@ public void shutdown()
...
+        getLogger().debug( "Initializing white-list of newly added {}", mavenRepository );
+        try
+        {
+            // mark it for noscrape if not marked yet
+            unpublish( mavenRepository );
+            // spawn update, this will do whatever is needed (and handle cases like blocked, out of service etc),
+            // and publish
+            updateWhitelist( mavenRepository );
+            getLogger().info( "Initializing non-existing white-list of newly added {}",
+                RepositoryStringUtils.getHumanizedNameString( mavenRepository ) );
```
kellyrob99, line 268: We don't need the RepositoryStringUtils call anymore do we?

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -225,13 +256,42 @@ public void shutdown()
...
+        }
+        catch ( Exception e )
+        {
+            getLogger().warn( "Problem during white-list initialisation of newly added {}",
+                RepositoryStringUtils.getHumanizedNameString( mavenRepository ), e );
```
kellyrob99, line 273: RepositoryStringUtils again

nexus-core/src/test/java/org/sonatype/nexus/proxy/maven/wl/internal/WLMaintenanceTest.java
```diff
@@ -224,11 +224,11 @@ public void smoke()
 
         {
             final List<String> entries = getEntriesOf( mavenRepository );
-            assertThat( entries.size(), equalTo( 5 ) );
+            assertThat( entries.size(), equalTo( 4 ) );
```
kellyrob99: can use hasSize(4) instead for better messaging in case of failure

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
cstamas, line 268: Yes, we had a discussion about this with @jdillon ... the Repository.toString is implemented since 2.3 and is fine to be used in logging.

Here, the RSU is used only -- as level is INFO and is among 1st things user sees when Nexus boots -- to have it "aligned" with other INFO messages, where repositories are read up from config, as those lines are using RSU also.

I don't have a strong stance about use or non-use of RSU in logging, but according to previous comments from Jason, all DEBUG and TRACE level logging (at least in WL) has removed RSU use to not have the "protected" (if logger.isDebugEnabled...), but in case of these INFO level logs I don't consider them as problem.

kellyrob99, line 268: Thanks for the explanation

kellyrob99:
+1

cstamas:
@kellyrob99 Thanks for comments, hasSize change applied.

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-7

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -169,26 +169,57 @@ public WLManagerImpl( final EventBus eventBus, final ApplicationStatusSource app
...
+            for ( MavenRepository mavenRepository : initableGroupRepositories )
+            {
+                if ( isMavenRepositorySupported( mavenRepository )
+                    && mavenRepository.getLocalStatus().shouldServiceRequest() )
+                {
+                    // groups will not be collected to needs-update list
+                    doInitializeWhitelistOnStartup( mavenRepository );
                 }
             }
         }
```
ifedorenko, line 215: Why do we need to iterate over hosted/proxy and group repositories separately here? doInitializeWhitelistOnStartup only loads existing prefix file and does not depend on group repository members as far as I can tell.

nexus-core/src/test/java/org/sonatype/nexus/proxy/maven/wl/internal/WLUpdatePropagationGroupUpdatesTest.java
```diff
@@ -0,0 +1,258 @@
...
+import org.sonatype.sisu.goodies.eventbus.EventBus;
+
+import com.google.common.eventbus.Subscribe;
+
+/**
+ * Testing issues NEXUS-5602 and NEXUS-5608
+ * 
+ * @author cstamas
+ */
+public class WLUpdatePropagationGroupUpdatesTest
```
ifedorenko, line 55: Can you add tests for group-of-groups?

jdillon:
VERY STRANGE src/test/java/core/whitelist/WhitelistWithHostedRepositoryIT.java

cstamas:
CI:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-9

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-9

cstamas:
Discussed over the phone with @ifedorenko, where he blessed this pull. Merging it.

PR #782, opened 2013-03-20
REVIEW: Enhance password hashing

scarlucci:
These changes are for NXCM-4361, which requested that we start salting password hashes. The bulk of the changes for this task are in sonatype/security, which has a corresponding pull request open

Salting:
Passwords are now salted prior to hashing. Salts are unique per user, per password. The salt is stored along side the password hash in security.xml. The model version of security.xml has been updated

Hashing algorithm:
Passwords are now hashed with SHA-512. Previously, passwords were hashed with SHA-1. The credentials matcher is backwards compatible with SHA-1 and MD5 hashes

Key Stretching:
The hashing algorithm is now applied N times, where N is defined in security-configuration.xml, in the new hashIterations element. The model version of security-configuration.xml has been updated.

Upgrading legacy users:
In order to convert old SHA-1 password hashes, we must wait for the user to login. When they do, we detect that their password hash is old, generate a unique salt, and regenerate a SHA-512 hash

### DIFF
--- a/nexus-core/src/test/java/org/sonatype/nexus/jsecurity/DefaultPasswordGeneratorTest.java
+++ b/nexus-core/src/test/java/org/sonatype/nexus/jsecurity/DefaultPasswordGeneratorTest.java
@@ -12,6 +12,11 @@
  */
 package org.sonatype.nexus.jsecurity;
 
+import java.util.HashSet;
+import java.util.Set;
+
+import junit.framework.Assert;
+
 import org.codehaus.plexus.ContainerConfiguration;
 import org.codehaus.plexus.PlexusConstants;
 import org.sonatype.nexus.test.PlexusTestCaseSupport;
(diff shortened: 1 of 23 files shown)


nexus-core/src/test/java/org/sonatype/nexus/jsecurity/DefaultPasswordGeneratorTest.java
```diff
@@ -75,4 +80,43 @@ public void testGeneratePassword()
...
+    	throws Exception
+    {
+    	String password = "test-password";
+    	String salt = pwGenerator.generateSalt();
+    	int hashIterations = 1024;
+    	
+    	String hash1 = pwGenerator.hashPassword(password, salt, hashIterations);
+    	String hash2 = pwGenerator.hashPassword(password, salt, hashIterations);
+    	
+    	Assert.assertEquals(hash1, hash2);
```
jdillon: Recommend all new tests use hamcrest matchers, ex:

```
 assertThat(hash2, is(hash1));
```

These produce more descriptive messages when the assertion fails.

nexus-core/src/test/resources/org/sonatype/nexus/configuration/upgrade/security-configuration-001-1.xml
```diff
@@ -10,4 +10,5 @@
     <realm>XmlAuthorizingRealm</realm>
   </realms>
   <securityManager>default</securityManager>
+  <hashIterations>1024</hashIterations>
```
jdillon, line 13: I'm a little confused about the changes to the upgrade test xml sources here.  IIUC these are to test upgrading... so shouldn't it have done the correct thing as part of the test execution to apply these changes to the configuration?

pom.xml
```diff
@@ -177,7 +177,7 @@
     <aether.version>1.13.1</aether.version>
     <jetty.version>8.1.8.v20121106</jetty.version>
     <maven.version>3.0.4</maven.version>
-    <plexus-security.version>3.0.2</plexus-security.version>
+    <plexus-security.version>3.0.2-SNAPSHOT</plexus-security.version>
```
jdillon: Addition of advanced hash/salting to security does not seem like its a bugfix.  The version of security would be bumped to 3.1-SNAPSHOT.

Also, you are changing a release 3.0.2 version back to its former 3.0.2-SNAPSHOT, if this was really just a bug fix it should have been the latest version which is 3.0.3-SNAPSHOT

nexus-core/src/test/java/org/sonatype/nexus/jsecurity/DefaultPasswordGeneratorTest.java
scarlucci: No problem. I'll go through all of the tests that I added and make the switch

nexus-core/src/test/resources/org/sonatype/nexus/configuration/upgrade/security-configuration-001-1.xml
scarlucci, line 13: I was a little confused too while going through this. We have these files all over the place, with varying naming conventions. In this case, this is actually one of the result files that is used to make sure a previous versions file was upgraded correctly

pom.xml
scarlucci: Yep, this was just my ignorance on our policy for handling version changes, and maven versioning in general. I agree that the version should be changed to 3.1-SNAPSHOT. I'll get that in

scarlucci:
Thanks for the feedback Jason. I believe I have incorporated all of your suggestions

PR #783, opened 2013-03-21
REVIEW: Move timeline ui

nabcos:
When the timeline was changed to be a plugin, the UI code was left in core. This change sets up timeline plugin to contribute its own UI.

This change is not for 2.4, and should be considered at the start of 2.4.1 or 2.5


adreghiciu:
+1 on the move

nabcos:
Rebased because of conflict with unrelated changes in master (r.js config).

kellyrob99:
+1

jdillon:
looks okay, be sure to update @since before merging

+1

PR #786, opened 2013-03-22
TRIVIAL: clean up httpclient logging

jdillon:


cstamas:
+1

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
```diff
@@ -181,7 +171,7 @@ protected void configureAuthentication( final DefaultHttpClient httpClient,
                 authScope = proxyHost.toHostString() + " proxy";
             }
 
-            List<String> authorisationPreference = new ArrayList<String>( 2 );
+            List<String> authorisationPreference = Lists.newArrayListWithExpectedSize(3);
```
ifedorenko, line 174: Logging cleanup?

ifedorenko:
+1

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
jdillon, line 174: Obviously not :-P  I did notice when looking over the logging here that when ntlm was configured that the size is expected to be 3 not 2.  Its a minor and safe optimization which I didn't bother to track in a separate change.

adreghiciu:
+1

PR #790, opened 2013-03-27
[2.5] NexusClient UA should carry proper version

cstamas:
This pull is pending, not for merge yet.

Currently, NexusClient represents itself as `NexusClient/1.0` where the "1.0" part is basically
a constant.

This made us unable to distinguish properly between clients
(like in case of staging) to differentiate between different
versions of Staging client.


ifedorenko:
+1

nabcos:
+1

nexus-client-core/src/main/java/org/sonatype/nexus/client/rest/jersey/JerseyNexusClientFactory.java
```diff
@@ -123,6 +125,12 @@ protected ApacheHttpClient4 doCreateHttpClientFor( final ConnectionInfo connecti
         return client;
     }
 
+    protected String discoverClientVersion()
+    {
+        // TODO: Should default be "1.0" to behave as older clients, if version is not discovered
+        return Version.readVersion( "/META-INF/maven/org.sonatype.nexus.client/nexus-client-core/pom.properties", "1.0" );
```
jdillon: very odd default version, probably better to use "unknown" or something to make it clear there is a problem instead of pretending its an older version.

cstamas: Fixed, "unknown" string used instead of "1.0"

kellyrob99:
I'm late to the party on this one, but this is very similar to other code like https://github.com/sonatype/nexus/blob/master/nexus-core/src/main/java/org/sonatype/nexus/plugins/rest/AbstractNexusIndexHtmlCustomizer.java#L49 for instance. 
Seems like a cross-cutting concern that we should be sharing code for.

PR #791, opened 2013-03-27
[2.5] NEXUS-5612 File listing should fail in case of IOEx

cstamas:
Fix for
https://issues.sonatype.org/browse/NEXUS-5612

Currently DefaultFSPeer was hiding the null returned
from File.listFiles. According to javadocs, it returns null
in two cases, a) when the File instance does not denote
a directory, or b) in case of IO error.

As case a) is checked few lines above (if check failed,
we would not get here), the null return value should
be treated as IO Error.

By throwing a generic LocalStorageExcetion, we
make this case to be treated as error.


nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSPeer.java
```diff
@@ -348,6 +348,9 @@ public boolean accept( File pathname )
                 getLogger().warn( "Cannot list directory in repository {}, path \"{}\"",
                                   RepositoryStringUtils.getHumanizedNameString( repository ),
                                   target.getAbsolutePath() );
+
+                throw new LocalStorageException( "Cannot list directory in repository " + repository + ", path "
```
ifedorenko, line 348: This will result in double-logging of the same "Cannot list directory in repository" message, one from explicit log statement above, another from exception handler.

ifedorenko:
Why pull request commend says "related to NEXUS-5612" and commit comment does not mention NEXUS-5612 at all? Does the commit solve the problem described in NEXUS-5612?

cstamas:
Commit is a week old, before the issue was created. We had a phone talk with Rich where he described the problem to me, but the issue was still not created. I just wanted to save this change, resulting from that phone call.

Will remove the log and redo the commit with proper message.

cstamas:
Commit redone with Igor's proposed changes and proper message.

Code reading shows that FSPeer (and it's implementation DefaultFSPeer) is used only in FS LS (and corresponding tests) as expected.

Furthermore, code also shows that the newly thrown `LocalStorageException` will propagate to caller, and in case of Walker, will be caught and will stop the context (walk will stop).

Still, some tests proving the last (IOEx during listing + walker) are needed....

adreghiciu:
+1

nabcos:
+1

adreghiciu:
Started a CI build: https://bamboo.zion.sonatype.com/browse/NXO-OSSF20-1

adreghiciu:
CI :+1:

PR #792, opened 2013-03-27
REVIEW: NEXUS-5602: Using the periodic update thread for initial updates too

cstamas:
On 1st boot of updated instances, a lot of MavenRepository
instances (hosted and proxy) will be piled up in the array
needUpdateRepositories in WLManagerImpl#startup().

Currently, as the thread pool with fixed size of 5 and
blocking queue was deferring update.

With this change, the for-loop doing initial WL build
(walk of the content for hosted and discovery for proxies)
is deferred, and periodic updater thread is used.

This also means, that upon boot, the initial delay of
first run will be applied, and on that moment on, hosted
reposes will have WL, not before that.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/EventDispatcher.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/EventDispatcher.java
@@ -58,19 +58,15 @@ public class EventDispatcher
 
     private final Manager wlManager;
 
-    private final boolean active;
-
     /**
      * Da constructor.
      * 
      * @param wlManager
-     * @param active
      */
-    public EventDispatcher( final Manager wlManager, final boolean active )
+    public EventDispatcher( final Manager wlManager )
     {
         this.logger = LoggerFactory.getLogger( getClass() );
         this.wlManager = checkNotNull( wlManager );
-        this.active = active;
     }
 
     protected Logger getLogger()
@@ -156,11 +152,6 @@ public class EventDispatcher
 
     // == Filters
 
-    protected boolean isActive()
-    {
-        return active;
-    }
-
     protected boolean isRequestContextMarked( final RequestContext context )
     {
         return context.containsKey( Manager.ROUTING_INITIATED_FILE_OPERATION_FLAG_KEY );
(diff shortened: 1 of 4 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -213,31 +219,32 @@ public void startup()
                 }
             }
         }
-        // spawn all the needed updates as bg jobs
-        // these will maintaing groups too as needed
-        for ( MavenRepository mavenRepository : needUpdateRepositories )
+        // add the periodic thread only when feature is active
+        if ( config.isFeatureActive() )
```
ifedorenko: Why do anything during startup() if autorouting is disabled?

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/WLManagerImpl.java
```diff
@@ -213,31 +219,32 @@ public void startup()
...
-        }, TimeUnit.MINUTES.toMillis( 1 ), TimeUnit.HOURS.toMillis( 1 ), TimeUnit.MILLISECONDS );
-        // register event dispatcher, to start receiving events
-        eventBus.register( eventDispatcher );
     }
 
     @Override
     public void shutdown()
     {
-        eventBus.unregister( eventDispatcher );
+        if ( config.isFeatureActive() )
```
ifedorenko: Why does it matter if the feature is enabled or not? Shutdown should unregister eventDispatcher regardless.

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF42-3

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/wl/internal/EventDispatcher.java
```diff
@@ -170,7 +161,7 @@ protected boolean isRepositoryHandled( final Repository repository )
     {
         // we handle repository events after this isActive, is not out of service, and only for non-shadow repository
         // that are Maven2 reposes
-        return isActive() && repository != null
```
nabcos: comment does not seem to match the code anymore

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF42-4

jdillon:
sure

cstamas:
CI Started after rebase against master with renames:
https://bamboo.zion.sonatype.com/browse/NXO-OSSF42-5

cstamas:
Previous CI "succeeded" with one unrelated IT failure.

Next CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF42-6

PR #793, opened 2013-03-27
NEXUS-5641 honour lucene.fsdirectory.type during index update

ifedorenko:
This is meant for 2.4.1

### DIFF
--- a/plugins/indexer/nexus-indexer-lucene-plugin/pom.xml
+++ b/plugins/indexer/nexus-indexer-lucene-plugin/pom.xml
@@ -31,7 +31,7 @@
     <pluginName>Nexus Indexer Lucene Plugin</pluginName>
     <pluginDescription>Adds search capabilities for repository content.</pluginDescription>
 
-    <mavenIndexer.version>5.1.1</mavenIndexer.version>
+    <mavenIndexer.version>5.1.2-SNAPSHOT</mavenIndexer.version>
     <aether.version>1.13.1</aether.version>
   </properties>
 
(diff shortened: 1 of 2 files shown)


jdillon:
no clue what this does or what its for... but looks fine

ifedorenko:
Everyone knows what lucene.fsdirectory.type does, no? 

Lucene FSDirectory.open picks actual implementation based on specifics of jvm/os/arch combination, see (http://svn.apache.org/repos/asf/lucene/dev/tags/lucene_solr_3_6_1/lucene/core/src/java/org/apache/lucene/store/FSDirectory.java) and on 64 bit Linux JVMs it picks MMapDirectory. For larger indexes this results in potentially tens of gigs of mapped address space used by JVM and although I was never able to reproduce this locally it sometimes triggers heavy operating system swap, to the point that JVM itself becomes unresponsive and gets killed by jsw. (all systems I looked at that showed this problem has swap partition on SAN, so my theory this is required to trigger the heavy swapping)

To workaround this problem we introduced lucene.fsdirectory.type property in 2.3, which allows nexus admin force specific FSDirectory implementation, like NIOFSDirectory, which does not have the problem with swapping. 

Unfortunately, the original fix missed one case where default Lucene FSDirectory.open behaviour was used and this pull fixes that.

jdillon:
Thanks for the detail.  I know jack about lucene (maybe a tiny bit more now after reading this though).

Go for it.

nabcos:
+1

cstamas:
MI updated to the deployed pseudo release.

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF43-1

PR #794, opened 2013-03-28
[2.4 blocker] NEXUS-5639: Rename WL

cstamas:
This is a "shallow" rename change (just in case) that gets rid of "whitelist" string from UI and logs.

A "deep" change, with class name changes on the way, will be a separate pull, in a spin off of this branch.


adreghiciu:
+1

ifedorenko:
+1

jdillon:
plus one

PR #796, opened 2013-03-28
REVIEW: Add event for plugins to add to or change the server settings view

nabcos:


jdillon:
How is a plugin intended to use this event?  Looks like the plugins will have to know about the fields of this object, making the integration brittle.  Should consider adding some public methods to expose to publics the types of changes they can add.  ie. addFormField() or addFieldSet() etc.

nabcos:
(Only) usage of the event is here:
https://github.com/sonatype/nexus-ssl/pull/8/files#L2R38
https://github.com/sonatype/nexus-ssl/pull/8/files#L3R77

Note that the event was updated to use the formPanel as payload, so that handlers do not need to know about the wrapper object, and that the first event was removed which was used to mangle the configuration used to initialize the form panel.

nexus-webapp/src/main/webapp/js/Sonatype/repoServer/ServerEditPanel.js
```diff
@@ -687,6 +690,8 @@ Ext.define('Sonatype.repoServer.ServerEditPanel', {
 
     securityConfigField = this.formPanel.find('name', 'securityEnabled')[0];
     securityConfigField.on('select', this.securitySelectHandler, securityConfigField);
+
+    Sonatype.Events.fireEvent('serverConfigViewPostInit', this.formPanel);
```
jdillon, line 701: is this documented anywhere?  if not then should add jsduck /*\* @event ... */ muck here to explain what this event is.

PR #798, opened 2013-04-03
REVIEW: NEXUS-5615 update UI to indicate first/last name are required and add a last name t...

kellyrob99:
...o default admin user for new deployments

https://issues.sonatype.org/browse/NEXUS-5615

### DIFF
--- a/nexus-core/src/main/resources/META-INF/security/security.xml
+++ b/nexus-core/src/main/resources/META-INF/security/security.xml
@@ -16,59 +16,59 @@
 <!-- Default nexus security configuration -->
 <!-- used as default config source -->
 <security>
-	<version>2.0.5</version>
-	<users>
-		<user>
-			<id>admin</id>
-			<firstName>Administrator</firstName>
-            <lastName>Nexus</lastName>
-			<!-- admin123 -->
-			<password>f865b53623b121fd34ee5426c792e5c33af8c227</password>
-			<status>active</status>
-			<email>changeme@yourcompany.com</email>
-		</user>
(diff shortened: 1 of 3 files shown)


jdillon:
Please cross link github and jira... so its easier to move between them.

https://issues.sonatype.org/browse/NEXUS-5615

nexus-core/src/main/resources/META-INF/security/security.xml
```diff
@@ -21,6 +21,7 @@
 		<user>
 			<id>admin</id>
 			<firstName>Administrator</firstName>
+            <lastName>Nexus</lastName>
```
jdillon: probably should be:

```
<lastName>User</lastName>
```

... which matches the others closer.

indent is also off but so is this entire file :-\

jdillon:
This doesn't actually fix existing installations... but not sure we care too much about those, since hopefully they have deleted this user ;-)

kellyrob99:
I considered updating existing installations, but since there's no functionality that is broken as-s it seemed overkill for such a small issue.
If anyone feels strongly about that, I'd welcome the opportunity to start learning about how to upgrade the configuration xml files properly.

jdillon:
@kellyrob99 its not worth the trouble don't bother.

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF23-1

adreghiciu:
+1

jdillon:
+1

PR #799, opened 2013-04-04
[2.5] NXCM-5153: NXCM-5161: Core changes for upload resource

cstamas:
Refactoring and cleanup.

Issues solved (well, indirectly solves):
https://issues.sonatype.org/browse/NXCM-5153 
https://issues.sonatype.org/browse/NXCM-5161

Changes:
- doc'd how upload actually works
- cleanup: variable and their scope cleanup, formatting cleanup
- minor logging cleanup
- support for form field processing extension (in subclasses of abstract upload resource)
- and the "key change", introduction of new methods (one called once when all the coordinates of upload payload is available), to make subclasses introduce new parameters easier.

The key change is introduction of new method
`uploadGavParametersAvailable()` that is in this same
class implemented as nop. This method is called by
upload method only once, at the moment when
upload GAVP coordinates are available.

None of these changes modify functionality of upload (nor the Artifact Upload resource is changed in any way). Same thing on UI side. The new changes are used to implement NXCM-5153 and NXCM-5161 in staging plugin.


plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
```diff
@@ -288,153 +288,130 @@ protected Object getContent( Variant variant, boolean redirectTo, Request reques
...
-        String classifier = null;
-
-        String packaging = null;
-
-        String extension = null;
-
-        ArtifactCoordinate coords = null;
-
-        PomArtifactManager pomManager =
+        // Every file selected for upload will generate a SEPARATE upload HTTP POST request
```
cstamas: This spec should be moved to javadoc.

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF47-1

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF47-1

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
```diff
@@ -283,158 +283,139 @@ protected Object getContent( Variant variant, boolean redirectTo, Request reques
...
-                    coords.setVersion( version );
-                    coords.setPackaging( packaging );
+                    // parameters are first in "nibble"
+                    processFormField( request, uploadContext, fi );
                 }
                 else
                 {
-                    // a file
-                    isPom = fi.getName().endsWith( ".pom" ) || fi.getName().endsWith( "pom.xml" );
+                    // a file, this means NO parameters will income anymore
```
nabcos, line 336: FTR, I don't like that this may break when you pass in parameters in the wrong order (files first). Maybe move the upload out of the for-loop?
But it's been like that forever IIUC, so we can leave it like this.

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
```diff
@@ -465,15 +445,193 @@ else if ( "hasPom".equals( fi.getFieldName() ) )
...
+        {
+            return repositoryId;
+        }
+
+        public void setRepositoryId( String repositoryId )
+        {
+            this.repositoryId = repositoryId;
+        }
+
+        public boolean isHasPom()
```
nabcos: isHasPom? better rename that field to 'pomAvailable' ;)

nabcos:
+1

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
cstamas, line 336: Yes if you read the original (those three lines talking about "nibbles" that are removed with this change) or the new (in javadoc) "specs" how this resource works (and did work, as no change has been introduced in this area), it is a _requirement_ that params must arrive first, then the content(s). And if you think about it, it does make sense: if a huge JAR would arrive before parameters (like "r" for repository ID), where would Nexus put it?

This has been like this forever and was designed like this.

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
cstamas: Done

adreghiciu:
+1

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/artifact/AbstractArtifactPlexusResource.java
nabcos, line 336: My point was that I think if you send the parameter order "r,g,a,v,file,c" the artifact will be deployed without classifier, because it will be deployed inside the loop. If I send "file,r,g,a,v" it will fail altogether. The jar file is placed in a temp storage by fileupload-commons anyway - we can easily hold on to it until we've seen all parameters.

Minor point, as you said it's been like this forever. I still don't like it. ;)

PR #800, opened 2013-04-05
REVIEW: [NEXUS-5651] Siesta resources authorisation

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-5651
https://bamboo.zion.sonatype.com/browse/NXO-OSSF49

The fix will solve two issues:
1. Make browsers pop up the basic auth dialog in case that user is not authenticated
2. Make possible to access siesta resources via rest clients as wget/restclient/curl

The fix does not solve the issues related to accessing siesta resources when security is off. 

The fix requires that siesta resources (/service/siesta/**) to be filtered through Shiro "authBasic" filter. The pull request introduces a generic way to configure Shiro filter chains.

### DIFF
--- a/nexus-web-utils/src/main/java/org/sonatype/nexus/web/FilterChainInstaller.java
+++ b/nexus-web-utils/src/main/java/org/sonatype/nexus/web/FilterChainInstaller.java
@@ -10,11 +10,11 @@
  * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
  * Eclipse Foundation. All other trademarks are the property of their respective owners.
  */
-package org.sonatype.nexus.plugins.siesta;
+package org.sonatype.nexus.web;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static org.sonatype.nexus.plugins.siesta.SiestaModule.MOUNT_POINT;
 
+import java.util.List;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Provider;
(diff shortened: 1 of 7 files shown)


kellyrob99:
+1

nexus-web-utils/src/main/java/org/sonatype/nexus/web/FilterChainInstaller.java
```diff
@@ -0,0 +1,75 @@
...
+    @Subscribe
+    public void onEvent( final NexusStartedEvent event )
+    {
+        for ( final FilterChain filterChain : filterChains )
+        {
+            protectedPathManager.get().addProtectedResource(
+                filterChain.getPathPattern(), filterChain.getFilterExpression()
+            );
+        }
+    }
```
jdillon, line 67: Doesn't this introduce a small window where the server is unprotected (or rather has inconsistent production) during startup?

adreghiciu, line 67: It is consistent with how restlet resources are protected.
In siesta resources case is not actually any protection, as we are filtering just in order to be consistent with how Nexus behaves for restlet resources. The resource calls are always protected due to AOP.

cstamas, line 67: FTR, in both deployment variations (Nexus bundle or WAR), the NexusStartedEvent event is fired before Jetty opens HTTP Port (or WAR is deployed, hence have requests routed to it).

cstamas:
+1

ifedorenko:
Do we actually want browser basic auth window to popup when user is not authenticated? I thought that was some problems on some browsers, no?

nabcos:
The problem is when you authenticate via browser popup, you cannot log out from Nexus UI anymore (unless you clean up your browser session/restart browser).

plugins/siesta/nexus-siesta-plugin/src/main/java/org/sonatype/nexus/plugins/siesta/AuthorizationExceptionMapper.java
```diff
@@ -31,6 +35,38 @@
     extends ErrorExceptionMapperSupport<AuthorizationException>
 {
 
+    private static final String AUTH_SCHEME_KEY = "auth.scheme";
+
+    public static final String AUTH_REALM_KEY = "auth.realm";
+
+    private static final String ANONYMOUS_LOGIN = "nexus.anonynmous";
```
nabcos: typo

adreghiciu: That one is copied from Nexus filters (copied because those filters are in restlet1x). I will change it in both places.

adreghiciu:
@ifedorenko @nabcos about browser popup. I made it work as restlet1x resources to be consistent. Do we want that jersey based views to show 403 and restlet1x to show the popup?

jdillon:
@nabcos btw nexus resources already have this problem of showing the browser auth dialog if you've not previously gone to the UI to authenticate.

I think we have to take a holistic approach to solving this problem... which has existing for many many many years IIUC.

Rip security out, replace with something sane, simple and comprehensive.  Or else we'll be chasing problems like this for years to come.

nabcos:
I know, and I agree. My comment was just for clarification what exactly is wrong with browser auth. IMO for now, siesta and restlet should behave the same.

jdillon:
Is this waiting for any more changes?  Or does this function already?

PR #802, opened 2013-04-05
REVIEW: Enhanced password hashing

scarlucci:
These changes are for https://issues.sonatype.org/browse/NXCM-4361. There are corresponding changes for this work in sonatype/security

The goal of the changes is to use a more secure mechanism to hash user passwords. The changes leverage the new PasswordService feature in Shiro 1.2 to encapsulate all of the logic related to how we hash passwords.

Salting:
Passwords are now salted prior to hashing. Salts are unique per user, per password.

Hashing algorithm:
Passwords are now hashed with SHA-512. Previously, passwords were hashed with SHA-1. The password matcher is backwards compatible with SHA-1 and MD5 hashes

Key Stretching:
The hashing algorithm is now applied N times, where N is defined in security-configuration.xml, in the new hashIterations element. The model version of security-configuration.xml has been updated.

Upgrading legacy users:
In order to convert old SHA-1 password hashes, we must wait for the user to login. When they do, we detect that their password hash is old and regenerate the hash


ifedorenko:
+1

nexus-core/src/test/java/org/sonatype/nexus/proxy/maven/routing/internal/TextFilePrefixSourceMarshallerTest.java
```diff
@@ -143,7 +143,7 @@ public void roundtrip()
         assertThat( outputStream.size(), greaterThan( 15 ) );
 
         final String output = new String( outputStream.toByteArray(), UTF8 );
-        assertThat( output, equalTo( withStandardHeaders( prefixFile1( false ) ) ) );
+        assertThat( output.replace("\r", ""), equalTo( withStandardHeaders( prefixFile1( false ) ).replace("\r", "") ) );
```
kellyrob99: Any special reason we have to strip carriage returns now?

nexus-web-utils/src/test/java/org/sonatype/nexus/web/PlexusContainerContextListenerTest.java
```diff
@@ -1,66 +0,0 @@
-/*
```
kellyrob99, line 1: This doesn't have any relevance to the task at hand, does it?

kellyrob99:
Couple of minor questions, but nothing that should hold up merging
+1

nexus-core/src/test/java/org/sonatype/nexus/proxy/maven/routing/internal/TextFilePrefixSourceMarshallerTest.java
scarlucci: This change is obsolete now. I had fixed this awhile ago on this branch, but ended up fixing on master in a better way. I will remove this change

nexus-web-utils/src/test/java/org/sonatype/nexus/web/PlexusContainerContextListenerTest.java
scarlucci, line 1: This test was going to need some updates to work with these changes, and the comments in the test suggested it was obsolete. Before spending any time on this test, I spoke with Tamas, who said to kill it

PR #804, opened 2013-04-06
REVIEW: NEXUS-5660 add java.net.preferIPv4Stack system property via nexus.properties

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-5660

NEXUS-5660 - java.net.preferIPv4Stack=true should be set by default in case Java 7+ or JMX is used with Nexus

### DIFF
--- a/nexus-test/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
+++ b/nexus-test/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
@@ -164,6 +164,7 @@ public abstract class AbstractNexusIntegrationTest
 
         // guice finalizer turned OFF
         System.setProperty( "guice.executor.class", "NONE" );
+        // NEXUS-5660 test with ipv4 as app runs with this set now
         System.setProperty( "java.net.preferIPv4Stack", "true" );
     }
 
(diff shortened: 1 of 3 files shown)


jdillon:
Any reason you didn't put this into defaults.properties

https://github.com/sonatype/nexus/blob/master/nexus-bootstrap/src/main/resources/org/sonatype/nexus/bootstrap/default.properties

?

peterlynch:
Can we guarantee that PlainSocketImpl static block is not read before default.properties is loaded?

http://www.docjar.com/html/api/java/net/PlainSocketImpl.java.html

-1

Seems like something that could easily regress without us knowing. Also this change does not fix the trial edition either - since the trial bundle uses a non JSW launcher.

For this reason, I am suggesting that this property be set in wrapper.conf and the analogous ini file in the trial bundle instead as a -D property to the JVM. There it has no chance of regressing.

Also, to catch possible regressions if the launcher changes, add a test which verifies this property is set in the JVM on launch - possibly using a test component or something.

cstamas:
IMO, this should go into wrapper.conf, since JSW as first thing will open a port that is used to communicate between JSW wrapper native process and Java process. Hence, it is very likely that the above mentioned code is probably executed well before nexus.props (or default.props) are read up to set that property...

kellyrob99:
OK, I'll take another look at this and get back to you guys

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF53-5

nexus-test/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/integrationtests/AbstractNexusIntegrationTest.java
```diff
@@ -164,6 +164,7 @@
 
         // guice finalizer turned OFF
         System.setProperty( "guice.executor.class", "NONE" );
+        System.setProperty( "java.net.preferIPv4Stack", "true" );
```
peterlynch, line 168: a comment here would be useful, like a link to the issue

kellyrob99, line 168: Done

PR #805, opened 2013-04-06
REVIEW: NXCM-5181 expose ability to force an update of SystemStatus 

kellyrob99:
https://issues.sonatype.org/browse/NXCM-5181

NXCM-5181 - Can't install new license into expired trial.

Allow for forcing an update of SystemStatus. Needs to be triggered in plugin to solve associated issue.


adreghiciu:
Started CI build: https://bamboo.zion.sonatype.com/browse/NXO-OSSF48-1

adreghiciu:
+1

kellyrob99:
+1 by CI

nabcos:
Hm, seems like github ate my comment - in short, I would prefer a `void updateSystemStatus()` method instead of `SystemStatus getUpdatedSystemStatus()`, especially because the only caller is not interested in the system status at all.

But that's minor, +0.9 as it is right now ;)

kellyrob99:
@nabcos I can do that, was just trying to preserve the existing semantics for potential future callers, but if we don't have a use case to consume the return value I'm fine with that

nabcos:
IIUC calling update and then getSystemStatus does not have a performance penalty (same status calculation and locking ops) and emphasizes that 'update' is not free (getting license details from PLM means FS roundtrip and decrypting the results)

nexus-core/src/main/java/org/sonatype/nexus/AbstractApplicationStatusSource.java
```diff
@@ -71,19 +71,32 @@ protected ReadWriteLock getLock()
...
     {
-        updateSystemStatusIfNeeded( false );
+        return doGetSystemStatus(false);
+    }
 
-        Lock lock = getLock().readLock();
+    /**
+     * Returns the SystemStatus, forcing an update of possibly cached information, and guaranteeing its consistent state.
+     */
+    public SystemStatus getUpdatedSystemStatus()
```
nabcos: should add `@since` on new public API

PR #806, opened 2013-04-08
Checkum policy blocks prefix list

cstamas:
Fixes
http://issues.sonatype.org/browse/NXCM-5188

Multiple fixes:
- NPE fixed that is apparent in log in the issue (is general, not prefix file related!)
- Implemented a fix: override per-request the ChecksumPolicy in Maven repositories
- making use of the core change in `RemotePrefixFileStrategy`, added UT
- added comments to remove temptation (like I had)

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF1-1

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/RemotePrefixFileStrategy.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/RemotePrefixFileStrategy.java
@@ -29,6 +29,7 @@ import org.sonatype.nexus.proxy.access.Action;
 import org.sonatype.nexus.proxy.item.RepositoryItemUid;
 import org.sonatype.nexus.proxy.item.StorageFileItem;
 import org.sonatype.nexus.proxy.item.StorageItem;
+import org.sonatype.nexus.proxy.maven.ChecksumPolicy;
 import org.sonatype.nexus.proxy.maven.MavenProxyRepository;
 import org.sonatype.nexus.proxy.maven.routing.Config;
 import org.sonatype.nexus.proxy.maven.routing.Manager;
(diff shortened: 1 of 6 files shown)


cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF1-1

nabcos:
+1

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/RemotePrefixFileStrategy.java
```diff
@@ -147,8 +148,14 @@ protected StorageFileItem retrieveFromRemoteIfExists( final MavenProxyRepository
         throws IOException
     {
         final ResourceStoreRequest request = new ResourceStoreRequest( path );
-        request.getRequestContext().put( Manager.ROUTING_INITIATED_FILE_OPERATION_FLAG_KEY, Boolean.TRUE );
         request.setRequestRemoteOnly( true );
+        request.getRequestContext().put( Manager.ROUTING_INITIATED_FILE_OPERATION_FLAG_KEY, Boolean.TRUE );
+        if ( ChecksumPolicy.STRICT == mavenProxyRepository.getChecksumPolicy() )
```
ifedorenko: Why not unconditionally use ChecksumPolicy.IGNORE? Not a big deal, but extra remote request for SHA1/MD5 files that almost certainly will not be there and do not add much even if they are, seems pointless.

ifedorenko:
+1 if you agree that ChecksumPolicy.IGNORE is the right checksum policy to use for prefix files.

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/RemotePrefixFileStrategy.java
cstamas: They will be there if remote is Nexus 2.4+, and they are certainly not there if remote is "static repo" like Central is, unless... we modify the already changed specs and make `prefixes.txt.sha1` a requirement for example. For transport corruption the checksum is still the best and simplest, and if it's already available, why not use it?

ifedorenko: because this complicates the code and repo-to-repo interactions ever so slightly, and I highly doubt this improves reliability of prefix file processing in any meaningful ways (all transport errors will most likely be caught by standard tcp/ip logic). anyways, I don't feel strong enough about this, merge it as is if you think checksum validation is useful here.

/me shrugs

PR #807, opened 2013-04-08
NXCM-5194: G level metadata merge creates duplicated entries

cstamas:
Reason is that entry equality was checked by comparing all
three properties of the Plugin: prefix, artifactId and name.

This commit "improves" this, by performing comparison
on prefix and artifactId only. This means that
duplication will not happen anymore on new
deploys, and will be corrected if maven metadata
rebuild task is run against repository (dupe entries will be removed). Still,
the one (edge) case "changing prefix" mentioned
in related issue is NOT fixed by this change.

Related:
https://issues.sonatype.org/browse/NEXUS-4448

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/AddPluginOperation.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/AddPluginOperation.java
@@ -12,7 +12,7 @@
  */
 package org.sonatype.nexus.proxy.maven.metadata.operations;
 
-import static org.sonatype.nexus.proxy.maven.metadata.operations.MetadataUtil.isPluginEquals;
+import static org.sonatype.nexus.proxy.maven.metadata.operations.MetadataUtil.isPluginPrefixAndNameEquals;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -88,8 +88,9 @@ public class AddPluginOperation
         {
             if ( p.getArtifactId().equals( plugin.getArtifactId() ) )
             {
-                if ( isPluginEquals( p, plugin ) )
+                if ( isPluginPrefixAndNameEquals( p, plugin ) )
                 {
+                    p.setName( plugin.getName() );
                     // plugin already enlisted
                     return false;
                 }
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/AddPluginOperation.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/AddPluginOperation.java
@@ -12,8 +12,6 @@
  */
 package org.sonatype.nexus.proxy.maven.metadata.operations;
 
-import static org.sonatype.nexus.proxy.maven.metadata.operations.MetadataUtil.isPluginPrefixAndNameEquals;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/MetadataUtil.java
```diff
@@ -56,4 +56,15 @@ public static boolean isPluginEquals( Plugin p1, Plugin p2 )
         return false;
     }
 
+    public static boolean isPluginPrefixAndNameEquals( Plugin p1, Plugin p2 )
```
ifedorenko: Name is misleading, should be isPluginPrefixAndArtifactIdEquals, unless, of course, the method is meant to check names, not artofact ids.

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/AddPluginOperation.java
```diff
@@ -88,8 +88,9 @@ public boolean perform( Metadata metadata )
         {
             if ( p.getArtifactId().equals( plugin.getArtifactId() ) )
             {
-                if ( isPluginEquals( p, plugin ) )
+                if ( isPluginPrefixAndNameEquals( p, plugin ) )
                 {
+                    p.setName( plugin.getName() );
```
ifedorenko, line 93: I don't know if there is a good way to merge plugins with the same prefix:artifactId but different names, but why did you choose to use new name instead of the original one?

cstamas, line 93: Rationale was to use "latest deployed" name, as Name might change during time. See comments in [NEXUS-4448](https://issues.sonatype.org/browse/NEXUS-4448). So, in short, latest deployed plugin "wins" (or when rebuild maven metadata happens, as versions become sorted, "greatest version" wins), as plugin name change is obviously "legal thing" to do.

Now, from comments in NEXUS-4448, currently we do not know about any code that actually _uses_ the name field. If it's a historic construct or meant just to help manual inspection of metadata, this change should be fine. Anyway, this change has minimal if no impact at all.

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/MetadataUtil.java
cstamas: Method renamed

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/metadata/operations/MetadataUtil.java
```diff
@@ -56,4 +56,15 @@ public static boolean isPluginEquals( Plugin p1, Plugin p2 )
         return false;
     }
 
+    public static boolean isPluginPrefixAndArtifactIdEquals( Plugin p1, Plugin p2 )
+    {
+        if ( StringUtils.equals( p1.getArtifactId(), p2.getArtifactId() )
```
adreghiciu: Why not just:

return StringUtils.equals( p1.getArtifactId(), p2.getArtifactId() ) && StringUtils.equals( p1.getPrefix(), p2.getPrefix() );

adreghiciu:
+1

ifedorenko:
merge it

cstamas:
Just for the record, the referenced JIRA contains _another_ problem, and fix is in maven-plugins project:
https://github.com/sonatype/nexus-maven-plugins/pull/34

Am merging this pull as @ifedorenko recommended, but this not makes the referenced JIRA issue testable yet.

PR #809, opened 2013-04-10
[NEXUS-5420] Prepend base URL to all emails

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-5420
https://bamboo.zion.sonatype.com/browse/NXO-OSSF28

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
@@ -14,6 +14,7 @@ package org.sonatype.nexus.email;
 
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Startable;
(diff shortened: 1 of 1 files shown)


nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
```diff
@@ -105,10 +111,48 @@ public MailRequestStatus sendMail( MailRequest request )
...
+
+        prependNexusBaseUrl( request );
         
         return eMailer.sendMail( request );
     }
 
+    /**
+     * Prepend to message body a link to this Nexus instance (base server URL).
+     */
+    private void prependNexusBaseUrl( final MailRequest request )
```
cstamas, line 123: Just FTR, this could be achieved simply by introducing (and redirecting) to new MailType, one that might pick up _real_ Velocity template from somewhere...

See the (default) MailTypes used currently by nexus, but these might be contributed by Nexus itself:
https://github.com/sonatype/sisu-mailer/blob/master/src/main/java/org/sonatype/micromailer/imp/DefaultMailType.java
https://github.com/sonatype/sisu-mailer/blob/master/src/main/java/org/sonatype/micromailer/imp/HtmlMailType.java

adreghiciu:
@cstamas You like this one better? sonatype/nexus/pull/812
I think is better as it gives the user the choice to have the url prepended or not

adreghiciu:
Closing in favor of sonatype/nexus#812

ifedorenko:
+1

adreghiciu:
See also sonatype/nexus/pull/828

PR #812, opened 2013-04-10
REVIEW: [NEXUS-5420] Prepend base URL to all emails (Take 2)

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-5420
https://bamboo.zion.sonatype.com/browse/NXO-OSSF31

Beside solving the issue, the pull introduces two new Nexus specific Mail types. When used those will prepend the email with Nexus base url. User still can use the micromailer default/html types that will work as before (no prepend).

The templates are read from VM files.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/email/DefaultNexusEmailer.java
@@ -14,6 +14,7 @@ package org.sonatype.nexus.email;
 
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Startable;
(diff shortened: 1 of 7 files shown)


kellyrob99:
+1

jdillon:
Does this do the same thing as https://github.com/sonatype/nexus/pull/809 ?  If so its change is a _lot_ smaller.  I feel like if we want to introduce templates we should do it properly and not half ass it, which is what I think this change does.

adreghiciu:
@jdillon 
It has the same end result as #809 but that one I closed in favour of this pull.

Why are you saying that is half ass done? 
The body/subject are already using velocity templates right now as in for example for body "$body". As in order to get the link back to nexus this template becomes more complicated, what this pull does is to load the template form a file instead of passing in a string which would be harder to read/maintain.

jdillon:
@adreghiciu I'm saying its half done with respect to how NX will use templates generally for all emails.  I don't think we should be attempting to solve this problem, or half solve it, until there is a vision for how templates in NX will work and a plan for how it will be implemented across the board.

So until that happens... I recommend the vastly simpler change in https://github.com/sonatype/nexus/issues/809

This change will likely only become cruft that adds no direct value and could potential make it more work to implement a general template email solution down the road, especially if these classes start to leak out in non uniform ways.

+1 on simple solution in https://github.com/sonatype/nexus/issues/809

-1 on this complex change which does the same given lack of vision and plan for general email tempting in NX.

adreghiciu:
I can go either ways. This solution also has the advantage of letting one to not add the "Message from:" text in top of emails by using the email types that were in place till now.

If #809 is in place then also the staging emails for example will have that prepended. With current pull that is an option.

jdillon:
IMO prepend or append is fine, as long as the detail is there.  These emails suck asis so no point trying to optimize them until we have a general solution.

ifedorenko:
-1, Lets go with https://github.com/sonatype/nexus/pull/809 instead. 

I don't have anything particular against this pull, but I don't know how proper template support should look like yet so can't tell if this changes takes us in the right direction or not (I can tell this change does not introduce templates support everywhere we'll likely need them). In the mean time, the simpler pull looks like much safer short-term solution to the problem at hand.

adreghiciu:
Closed in favor of sonatype/nexus/pull/828

PR #813, opened 2013-04-11
[REVIEW] NEXUS-2450: [M1 to M2] EJB artifacts are served from **/jars/ directory instead of **/ejbs/

cstamas:
Fixes
https://issues.sonatype.org/browse/NEXUS-2450

Use case: using Maven3, and you need to access some legacy Maven1 layouted repository with EJBs. As Nexus has not enough information (especially the "type" of the artifact in question) only the path, this change simply makes Nexus to iterate one extra M1 folder when looking for artifact. This happens in doRetrieveItem method.

No other change, despite introduction on List<String> as return params on coversion methods, as all the effected metods (naturally except the doRetrieveItem) just use the 1st element of the list, retaining same behaviour as before (when only single string was returned).

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/LayoutConverterShadowRepository.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/LayoutConverterShadowRepository.java
@@ -14,7 +14,10 @@ package org.sonatype.nexus.proxy.maven;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 
 import org.codehaus.plexus.component.annotations.Requirement;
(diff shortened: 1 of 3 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/LayoutConverterShadowRepository.java
```diff
@@ -446,30 +449,39 @@ protected String transformM2toM1( final String path )
...
-        sb.append( RepositoryItemUid.PATH_SEPARATOR );
-        sb.append( gav.getName() );
-        return sb.toString();
+        final ArrayList<String> exts = new ArrayList<String>();
+        exts.add( gav.getExtension() + "s" );
+        if ( extraFolders != null && !extraFolders.isEmpty() )
+        {
+            exts.addAll( extraFolders );
+        }
+        final ArrayList<String> result = new ArrayList<String>( exts.size() );
```
jdillon: minor: Any reason why ArrayList is used specifically here?  And related why isn't guava Lists.\* helpers being used?  That is the direction the rest of the codebase is heading, not sure why this is any different.

cstamas: No special reason, updated

PR #814, opened 2013-04-11
[REVIEW] NEXUS-5621: Blocking routing rules are not enlisted in describe

cstamas:
Fixes:
https://issues.sonatype.org/browse/NEXUS-5621

the "appliedMappings" contains both INCLUSIVE and EXCLUSIVE rules
when asked for describe a given request, but BLOCKING mappings
were not enlisted

This simple change makes sure those are enlisted too, making
it easier to detect why something is not served up (in case
BLOCKING rules are used at all).


jdillon:
I'm guessing this is ready for review?

+1

cstamas:
Yes, it is meant for review. For me it makes more sense to mark "exceptional" pulls with some extra prefix than unconditionally _mark all_, since 90% of pulls _is meant for review_. Marking the rest 10% with "WIP" or whatnot sounds more reasonable to me :)

PR #815, opened 2013-04-11
REVIEW [NXCM-5147] allow for persistent tooltips

kellyrob99:
https://issues.sonatype.org/browse/NXCM-5147
Move code originally implemented as a one-off in nexus-clm to core for reuse.

NXCM-5147 - Resolve Technical Debt introduced by borrowing of helpText code in CLM plugin


jdillon:
+1

PR #816, opened 2013-04-12
REVIEW: Add servlet filter to add userId to MDC

jdillon:
https://issues.sonatype.org/browse/NEXUS-5662


adreghiciu:
+1

cstamas:
+1

Btw, what about threads running as scheduled tasks (actually all non-jetty HTTP related threads)?

jdillon:
@cstamas scheduled tasks specifically change user ctx to the system "task user".  For other cases I can not say atm if there is user ctx or not.  I can guess however that given the large number of threads we have w/o a system-wide thread-mgt strategy that changing this for all threads would be non-trivial.

I mentioned this a few times, and you are the first to say anything.  I'm not sure its worth the trouble to add this, but if you know of any specific places where a non-jetty thread actually carries user context please let me know so I can look at it closer.

cstamas:
For scheduled tasks with the ctx you mention:
https://github.com/sonatype/nexus/blob/5ff39ca62d0b999ef5774f2da0e92d6d927a0857/nexus-core/src/main/java/org/sonatype/nexus/scheduling/AbstractNexusTask.java#L171

For other pools in general (like Auto-Routing, Proxy Remote Status Checker, etc), they all use same thread factory: `org.sonatype.nexus.threads.NexusThreadFactory`. So, if we want, we _could_ set those here.

jdillon:
```
public Thread newThread( Runnable r )
```

would have to wrap the runnable to set/unset the mdc key when the thread is started.  but I can't say if the context will be correct if the threads are used in a pool and reused by whatever system created them. so it is still not simple IMO to simply do this generically.  If there was however an abstraction to "do some work" and not "create a thread" then it could have been generic to set/unset MDC keys around the execution of "do some work".

PR #817, opened 2013-04-15
[REVIEW] Routing fixes

cstamas:
Cumulative fixes for automatic routing.

Changes:
- improve logging of prefix updates (log only when repo transitions itself from published to unpublished or other way around, not ALWAYS)
- UT and fix to make proxy repositories not change downloaded prefix file, hence preserving them as-is for downstream consumption.

Already merged to master:
- fix NPE (affecting mostly routing related UTs and ITs) that makes tests unstable. There was a "hole" that made tests able to kick in too early, when booting was not finished (yet, test assertions assumed it was done).

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF17-1


adreghiciu:
+1

nexus-core/src/main/java/org/sonatype/nexus/proxy/maven/routing/internal/ManagerImpl.java
```diff
@@ -550,20 +570,34 @@ else if ( mavenRepository.getRepositoryKind().isFacetAvailable( MavenHostedRepos
...
                 getLogger().info( "Repository {} unsupported by automatic routing feature",
                     RepositoryStringUtils.getFullHumanizedNameString( mavenRepository ) );
                 return;
             }
+
+            // this is never null
+            final PrefixSource oldPrefixSource = getPrefixSourceFor( mavenRepository );
+            // does repo goes from unpublished to published or other way around?
+            final boolean stateChanged =
+                ( oldPrefixSource.supported() ) != ( prefixSource != null && prefixSource.supported() );
```
nabcos, line 583: copy/paste from above, for proxy and hosted? code seems to do exactly the same

nabcos:
+1 except for DRY

And IIUC the fix for the tests is only 4 lines long, why not push that directly?

cstamas:
Ok, the 4 lines pushed to master, but this pull also remains to fix other problems, like annoying logs or prefix updates and such.

nabcos:
Sure, just wondering why you rush this pull because of 4 lines that could be pushed directly or have their own pull request or something.

cstamas:
No rush. NPE relevant parts in master, description updated.

PR #818, opened 2013-04-15
[POC] Exposing attributes to deployer.

cstamas:
With this change, item attributes might be contributes at
deploy, sourced from HTTP headers and/or from query parameters.

One possible use case with current maven (where it's impossible
to add query parameters to PUT requests AFAIK) is following:

Modify your settings.xml server entry with something like
this:

```
    <server>
      <id>nexus</id>
      <username>deployment</username>
      <password>deployment123</password>
      <configuration>
        <httpHeaders>
          <property>
            <name>X-Nexus-Deploy-Attribute-BuildId</name>
            <value>NXOSS-1234</value>
          </property>
        </httpHeaders>
      </configuration>
    </server>
```

Note1: the part AFTER "X-Nexus-Deploy-Attribute-" prefix is
used as attribute key, so you can add "X-Nexus-Deploy-Attribute-Foo"
and also "X-Nexus-Deploy-Attribute-Bar" etc. As many entries you want.

Note2: on CI it is worth to filter the header value with something meaningful,
like job ID, or current SCM commit, etc...

If you have settings.xml set up and used in Maven as above, and
you deploy, all deployed artifacts will have an attribute(s)
added like this (for details read javadoc):

deploy.BuildId=NXOSS-1234

Now, if you combine this with the example plugin
https://github.com/sonatype/nexus-example-plugins/tree/master/selectionactors

TagSelector (where you could have `?attributeKey=nexus.BuildId&attributeValue=NXOSS-1234`)
you can act on whole set of artifacts originating from same CI job or
similar.

Extended NexusClient.Content subsystem with two new methods:
- deployWithAttributes to upload _with_ user attributes
- getAttributes that reads file item attributes (for now it uses describe under the hud)

Added IT covering this change.

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF32-1


jdillon:
why do we care about this?

cstamas:
Simplest use case: "decorate" the given deploy with CI job ID that deployed it. So, provide mapping between timestamped snapshots and CI Jobs deploying them. Next, you can even add SCM commit ID to the attributes.

As an example, using this solution above with CI builds, to figure out changes you are testing would be brain dead simple (or changes contained in a snapshot). With request like this:
https://repository.sonatype.org/content/repositories/snapshots/org/sonatype/appcontext/appcontext/3.3-SNAPSHOT/appcontext-3.3-20120606.140136-1.jar?describe

you would immediately know (the attributes map in JSON response would contain those values) where to go back to a CI job building and deploying that artifact, and CI would have a chance to tell you what _changes_ are in this snapshot. But same is true for release deploys.

Anyway, exposing this as you see in change has minimal impact, as if nothing of those (headers or query params) are used, ie. if you simply neglect this feature, nothing changes on server side. Impact is minimal.

jdillon:
@cstamas I still don't understand why you took the time to write this... was this a requested feature?  Is this something you needed?  Or did you just code this up for fun?

jdillon:
@cstamas also could use some unit tests to ensure this works and doesn't break

kellyrob99:
Do we have clients looking for this functionality now?

PR #819, opened 2013-04-15
[POC] Preemptive auth test

cstamas:
As it turns out, HC4 RRS does not authenticate preemptively.

This pull contains a UT that confirms that.

Also, there is a change in RRS that fixes the problem but _only for hosts doing Basic authentication scheme_. Unsure how would this behave if remote does Digest or NTLM auth scheme (unknown before 1st request done to it). More about this in RRS line comments.

Best we could do here is similar to the "S3 trick" (flagging is remote S3 bucket or not, known and remembered after 1st remote access), and fix this for Basic protected remote peers, but do nothing for others... if we want to handle this at all.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/httpclient/HttpClientRemoteStorage.java
@@ -23,8 +23,10 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.apache.http.Header;
+import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
+import org.apache.http.client.AuthCache;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpDelete;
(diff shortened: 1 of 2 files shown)


ifedorenko:
I think the right solution is to enable auth cache, so nexus would make one unauthenticated request, get 401 challenge, then do required auth without being further challenged. See http://hc.apache.org/httpcomponents-client-ga/tutorial/html/authentication.html

cstamas:
Right, that's what is written in line comments here:
https://github.com/sonatype/nexus/pull/819/files#L0R499

But as I wrote in comment, since HttpContext cannot be shared across multiple threads, and "used" instances needs to be reused (to make use of pre-populated auth cache), we'd need some sort of pool of Contexts. 

Or do I miss something?

kellyrob99:
What actual problem are we trying to solve here?

cstamas:
This pull does not solves the problem, so here's the backing story.

RRS (repo remote storage) using HC3, when set up with authentication info, was using that info preemptively. Preemptive auth in case of proxy repository is okay (while preemptive auth globally in HTTP world is not always okay), as proxy repo hits one single URL (well, the repo base URL appended with artifact path, but this means always same host is hit, that has same auth set up). But it seems RRS with HC4 does not preemptively authenticate.

Without preemptive authentication, a "ping-pong" happens: Nexus asks for artifact, remote server responds with 401 and a challenge, Nexus again asks for same artifact but with request equipped with challenge response, and this repeats for _every_ request from that repo, even if Nexus proxy repo hits same URL over and over again. 

But, since Nexus always asks for hashes too, one client request (like a maven asking for a JAR from Nexus) results in 4 HTTP exchanges on nexus-remote server part coz of this (best case). Worst case, when JAR exists but have no hashes results in 6 HTTP exchanges between Nexus and remote server, while client is waiting for the JAR. Here is what happens:

```
 maven/client            nexus           remote repo
     GET a ---------------->
                         GET a -------------->
                           <--------------- 401
                         GET a -------------->
                           <--------------- 200
                         GET a.sha1 --------->
                           <--------------- 401
                         GET a.sha1 --------->
                           <--------------- 404
                         GET a.md5 ---------->
                           <--------------- 401
                         GET a.md5 ---------->
                           <--------------- 404
       <------------------ 200
```

So, finally, this pull has:
- UT that verifies RRS is preemptively authenticating (it fails currently, at least without commit below)
- the commit https://github.com/sonatype/nexus/commit/8bc4f3657c3c1274f7182ec24f3c48e777a80287 that is NOT MEANT FOR PRODUCTION, see the inline comments why, but makes the build pass

PR #820, opened 2013-04-15
[REVIEW] NEXUS-5468: HTTP connection leak in case of LSEx

cstamas:
Fix for LRS to always clean up PreparedContentLocator instances
and a UT verifying this happens in both case, in "norma" but
also in "early unexpected" exception case.

Issue:
https://issues.sonatype.org/browse/NEXUS-5468

CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF2-1

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSLocalRepositoryStorage.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSLocalRepositoryStorage.java
@@ -52,6 +52,7 @@ import org.sonatype.nexus.proxy.wastebasket.Wastebasket;
 import org.sonatype.nexus.util.ItemPathUtils;
 
 import com.google.common.base.Strings;
+import com.google.common.io.Closeables;
 
 /**
  * LocalRepositoryStorage that uses plain File System (relies on {@link File}) to implement it's functionality.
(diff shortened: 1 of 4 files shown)


kellyrob99:
+1

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/local/fs/DefaultFSLocalRepositoryStorage.java
```diff
@@ -328,42 +330,67 @@ public AbstractStorageItem retrieveItem( Repository repository, ResourceStoreReq
...
+            getFSPeer().storeItem( repository, getBaseDir( repository, item.getResourceStoreRequest() ), item, target, cl );
+        } 
+        finally
+        {
+            // NEXUS-5468: Ensure that in case of file item with prepared content
+            // (typically those coming from RRS, as the content is actually wrapped HTTP response body, hence not reusable)
+            // get closed irrelevant of the actual outcome. If all went right, stream was already closed,
+            // and we will be "punished" by one extra (redundant) call to Closeable#close().
+            if ( originalContentLocator != null
+                && ( originalContentLocator instanceof PreparedContentLocator ) )
```
ifedorenko: fyi, instanceof implies !=null

ifedorenko:
-1

The code is specific to PreparedContentLocator, however there is no guarantee this other ContentLocator locators do not also leak opened InputStream instances, either directly or through AbstractWrappingContentLocator. 

One way is to completely rework LocalRepositoryStorage/FSPeer API, which I believe we'll need to do in the long term, but this is way too big a change for 2.5.

In the short term, I suggest 1) change DefaultFSLocalRepositoryStorage#storeItem to close any Closable ContentLocator implementation and 2) make AbstractWrappingContentLocator implement Closeable and close wrapped location if it implements Closeable.

cstamas:
Agreed, will change this. I was also unease about being tied to `PreparedContentLocator`.

cstamas:
Igor's comments applied.

PR #821, opened 2013-04-15
REVIEW [NEXUS-5672] Create an ExtJS3 UI plugin

nabcos:
https://issues.sonatype.org/browse/NEXUS-5672

This pull request collects most scattered UI resources into one plugin. Only UiContributor/IndexHtmlCustomizer remains in nexus-core and is changed to honor the `?debug` URL parameter. Plugin javascript will then be loaded in the single (dev) files instead of the minified version. By using XmlPluginRepository + symlinking resources/static/js to target/classes/static/js you can edit the js in place and only need to reload the Nexus UI, no need to restart Nexus for UI dev anymore.

This pull changes setup for yuicompressor plugin, so every plugin contributing UI will need to move their js/css files to src/main/resources/static/(js|css) and needs a dependency on the new UI plugin to pick up UiContributor/IndexHtmlCustomizer.

Next step in terms of UI maintenance/improvements would be to add something like https://github.com/pguedes/maven-jsduck to produce js documentation. That plugin is working, but it seems like that particular plugin is not published into a public repository ATM, so I refrained from adding it to the build for now (there's not much documentation in the current sources anyway).


jdillon:
Curious why UiContributor/IndexHtmlCustomizer left in core?

plugins/nexus-ui-extjs3-plugin/src/main/css/Sonatype.css
```diff
@@ -197,7 +197,7 @@ button.st-icon-refresh {
 
 .form-label-helpmark-combo {
     /* large margin needed to place helpmark next to combobox - the box is position:absolute, and helpmark was covered by dropdown arrow */
-    margin-left: 19px;
```
kellyrob99, line 200: Appear to have overwritten a change here IIRC

plugins/nexus-ui-extjs3-plugin/src/main/java/org/sonatype/nexus/plugins/ui/RedirectingNexusApplicationCustomizer.java
```diff
@@ -0,0 +1,71 @@
...
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sonatype.nexus.rest.NexusApplication;
+import org.sonatype.nexus.rest.NexusApplicationCustomizer;
+import org.sonatype.plexus.rest.RetargetableRestlet;
+
+/**
+ * Installs a filter before the NexusApplication root that will redirect /images to /static/images.
```
kellyrob99, line 28: So this is to handle plugins that still load from /images path?

kellyrob99:
I could use a little more context as to why this change is beneficial please, as it seems a bit counter-intuitive to be moving all of our js/css out of the actual webapp module?

plugins/nexus-ui-extjs3-plugin/src/main/css/Sonatype.css
nabcos, line 200: I will probably rebase and redo the "move js/css" commits before merge, so that the latest changes from master are picked up here.

(I don't think git can handle a merge from master directly)

plugins/nexus-ui-extjs3-plugin/src/main/java/org/sonatype/nexus/plugins/ui/RedirectingNexusApplicationCustomizer.java
nabcos, line 28: Yes, for compatibility reasons.

nabcos:
@kellyrob99 
- unify build (behaves just like any other plugin)
- easier UI development (just drop the plugin into any dist/use XmlPluginRepository instead of rebuild dist/unzip war contents/copy files)
- separate from unrelated plugins (UI needed parts in core, webapp, restlet plugin)
- make way for alternative implementation

kellyrob99:
Thanks for the explanation Ben.
+1

jdillon:
+1 this seems to be steps in the right direction for current UI framework to support ?debug better.

PR #830, opened 2013-04-19
[REVIEW] Turn Velocity template caching on.

cstamas:
As turned out, Templates are always re-read from classpath
causing congestion.

Using SISU Velocity's VelocityConfigurator to configure
enging to Nexus needs (as it was using defaults) set in
VelocityImpl of sisu-velocity.

This NX specific Configurator overrides the defaults from here
https://github.com/sonatype/sisu-velocity/blob/master/src/main/java/org/sonatype/sisu/velocity/internal/VelocityImpl.java#L74

These defaults were used by NX so far.

Verified that caching works (as this is Velocity concern
not Nexus) by debugging. Method (line 294):

org.apache.velocity.runtime.resource.ResourceManagerImpl.getResource(String, int, String)

Verified that template is loaded up once, and subsequent calls
are retrieving it from "globalCache".

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
@@ -1,0 +1,83 @@
+/*
+ * Sonatype Nexus (TM) Open Source Version
+ * Copyright (c) 2007-2012 Sonatype, Inc.
+ * All rights reserved. Includes the third-party code listed at http://links.sonatype.com/products/nexus/oss/attributions.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse Public License Version 1.0,
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.velocity;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.apache.velocity.app.VelocityEngine;
+import org.apache.velocity.runtime.RuntimeConstants;
+import org.sonatype.sisu.velocity.internal.VelocityConfigurator;
+
+/**
+ * Nexus specific {@link VelocityConfigurator} implementation, that configures {@link VelocityEngine} for Nexus, and
+ * supports "development" or "production" mode, depending on configuration (system property
+ * {@code nexus.velocity.production} that defaults to {@code true}). By default, it configures the "production" mode,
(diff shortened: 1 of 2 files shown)


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF11-1

nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
```diff
@@ -0,0 +1,83 @@
...
+@Named
+public class NexusVelocityConfigurator
+    implements VelocityConfigurator
+{
+    private final boolean production;
+
+    @Inject
+    public NexusVelocityConfigurator( @Named( "${nexus.velocity.production:-true}" ) final boolean production )
+    {
+        super();
```
kellyrob99: no need to call super here is there?

nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
```diff
@@ -0,0 +1,83 @@
...
+        this.production = production;
```
kellyrob99: checkNotNull?

kellyrob99:
Nice! Was this change sparked by the thread dumps we saw yesterday on OSSRH blocking on reading templates out of jars?

kellyrob99:
+1 BTW :)

peterlynch:
_looks_ fine, although I would rather a test that verifies all the expected properties are set with expected values within the Nexus used VelocityEngine at runtime, for production and development.

For example, supposed there are two velocitiy configurators that set the same property - which one wins if a List is given to https://github.com/sonatype/sisu-velocity/blob/master/src/main/java/org/sonatype/sisu/velocity/internal/VelocityImpl.java#L45 or createEngine - looks like last one set - is the order predictable?

nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
jdillon: boolean is never null

nexus-core/src/main/java/org/sonatype/nexus/velocity/NexusVelocityConfigurator.java
```diff
@@ -0,0 +1,83 @@
...
+        engine.setProperty( "runtime.references.strict", "false" );
+    }
+
+    protected void configureForDevelopment( final VelocityEngine engine )
+    {
+        // caching OFF
+        engine.setProperty( "class.resource.loader.cache", "false" );
+        // strict mode ON for early error detection
+        engine.setProperty( "runtime.references.strict", "true" );
+    }
```
jdillon: why do we care about this production/dev muck?

we use velocity for almost nothing... and begs to question why we have velocity at all

kellyrob99:
There are several places in code where we create a VelocityRepresentation directly prior to rendering - does this caching prevent those calls from reloading templates as well?

cstamas:
@kellyrob99  Yes, as all of restlet related representations uses SISU Velocity (sticked into Context). Another beneficiary component is SISU Mailer (used for mails). So, yes, this turns on caching for all those spots, affecting "index page", "status page" (error response rendering), emails and such.

adreghiciu:
+1

PR #831, opened 2013-04-19
[REVIEW] Caching MimeUtil2 responses

cstamas:
Calling into MimeUtil2 hides a point where
congestion might happen, as all these calls
are hitting a single j.u.Properties instance.

This change adds a cache a top of MimeUtil2
(actually in Nexus component wrapping it),
and should reduce actual calls to MimeUtil2.

The calls to guess MIME types are happening often, at high rate.
It happens at every creation of StorageFileItem.

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/mime/DefaultMimeSupport.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/mime/DefaultMimeSupport.java
@@ -16,7 +16,6 @@ import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF40-1

kellyrob99:
+1

nexus-core/src/main/java/org/sonatype/nexus/mime/DefaultMimeSupport.java
```diff
@@ -88,11 +132,18 @@ public String guessMimeTypeFromPath( final MimeRulesSource mimeRulesSource, fina
...
+        // see constructor how it is configured
+        // Note: using same method to get extension as MimeUtil2's MimeDetectors would
+        final String pathExtension = MimeUtil2.getExtension( path );
+        try
+        {
+            return extensionToMimeTypeCache.get( pathExtension );
+        }
+        catch ( ExecutionException e )
+        {
+            throw new RuntimeException( e );
```
adreghiciu: Use throw Throables.propagate(e);

adreghiciu:
+1

nexus-core/src/main/java/org/sonatype/nexus/mime/DefaultMimeSupport.java
```diff
@@ -88,11 +133,20 @@ public String guessMimeTypeFromPath( final MimeRulesSource mimeRulesSource, fina
...
+        // Note: using same method to get extension as MimeUtil2's MimeDetectors would
+        final String pathExtension = MimeUtil2.getExtension( path );
+        try
+        {
+            return extensionToMimeTypeCache.get( pathExtension );
+        }
+        catch ( ExecutionException e )
+        {
+            Throwables.propagate( e );
+            // only to make compiler happy, execution will never get here
```
nabcos: JFYI you can `throw Throwables.propagate(e);`

PR #832, opened 2013-04-19
[REVIEW] Remove ShadowEventInspector

cstamas:
Changes:
- removed nasty ShadowRepositoryEventInspector that put load
  on repo registry, instead made shadows itself "listen" and
  act on their own.
- made shadow repository "cache" it's master instance,
  instead using registry to look it up per every getMasterRepo
  invocation.

These changes above does not disrupt how shadows work at all,
but have a huge benefit that Registry is not used anymore
during event processing shadows are interested in. Registry
is used only once, if master is set using it's ID (that
happens only during boot and/or shadow creation).


cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF60-1

kellyrob99:
Is there a Jira to go with this?

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractShadowRepository.java
```diff
@@ -128,6 +130,7 @@ public LocalStatus getLocalStatus()
     }
 
     @Override
+    @Subscribe
     public void onRepositoryItemEvent( final RepositoryItemEvent ievt )
```
jdillon, line 134: where is the corresponding EventBus.register() done for this to work?

adreghiciu, line 134: This class inherits from AbstractConfigurable that does the registration

jdillon, line 134: k, thx

adreghiciu:
+1

nabcos:
+1

PR #834, opened 2013-04-19
REVIEW: [NEXUS-5214] Make possible to configure separate general (default) HTTP proxies for http and https

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-5214
https://issues.sonatype.org/browse/NEXUS-5690
https://bamboo.zion.sonatype.com/browse/NXO-OSSF61

The changes required:
- HC4: Hc4ProviderBase, NonProxyHostsAwareHttpRoutePlanner, UrlSchemeAwareHttpRoutePlanner 97a04dc0cab950dca3aa529947cc94995325378f
- nexus configuration to accomodate the new globalHttpsProxySettings  section 9bffe21dd0e8800cac35
- add GlobalHttpsProxySettings that is the iface to interact with nexus config regarding the new configuration option. As this is similar to GlobalHttpProxySettings, in order to reuse code I extracted superclasses and inherit them in both http/https 97a04dc0cab950dca3aa529947cc94995325378f and  3e3bc028dc4edce1b41f70c6cb8dcf0df45344cc
- REST (value objects/resource) d44da1738b223427137bdca04ec72781ab305807
- UI 22f088b5c1b63980f50d044d44d3ade52b72acfd and 9ba3e2273e5358475a1df042bfe80a25d0965a5b
- IT testing what proxy is used based on combination of defined proxies at global http/https/repository level and used url scheme (http/https) 01c64f719158b966dadd492b54f97282b03e5f67
- add support for setting https proxy via nexus client b88d07d8428b9a7dcc49564f9da52802ce444678
- add support for setting web proxy on repository via nexus client 1d994349f5de779b2ee93f64b2120cbaa9259da8


nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
```diff
@@ -214,41 +217,70 @@ else if ( ras instanceof UsernamePasswordRemoteAuthenticationSettings )
         }
     }
 
-    protected void configureProxy( final DefaultHttpClient httpClient, final RemoteProxySettings remoteProxySettings )
+    /**
+     * @since 2.5
```
kellyrob99, line 219: Is this appropriate for a pre-existing method that we're refactoring?

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
```diff
@@ -214,41 +217,70 @@ else if ( ras instanceof UsernamePasswordRemoteAuthenticationSettings )
...
+     */
+    protected void configureProxy( final DefaultHttpClient httpClient, 
+                                   final RemoteProxySettings httpProxySettings,
+                                   final RemoteProxySettings httpsProxySettings)
     {
-        if ( remoteProxySettings.isEnabled() )
+        if ( httpProxySettings != null && httpProxySettings.isEnabled() )
         {
-            getLogger().debug("proxy setup with host '{}'", remoteProxySettings.getHostname());
+            if ( httpsProxySettings != null && httpsProxySettings.isEnabled() )
```
kellyrob99: Looks like we're only setting proxies now if we have BOTH http and https settings enabled?

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
```diff
@@ -0,0 +1,288 @@
...
+        if ( isEnabled() )
+        {
+            try
+            {
+                return authenticationInfoConverter.convertAndValidateFromModel( getCurrentConfiguration( false )
+                                                                                    .getAuthentication() );
+            }
+            catch ( ConfigurationException e )
+            {
+                // FIXME: what here??
```
kellyrob99: Minimally we should be logging the error here, but is there a particular reason why this getter requires the try/catch and others do not?

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
```diff
@@ -0,0 +1,288 @@
...
+
+        getCurrentConfiguration( true ).setAuthentication(
+            authenticationInfoConverter
+                .convertToModel( proxyAuthentication ) );
+    }
+
+    public RemoteProxySettings convertAndValidateFromModel( CRemoteHttpProxySettings model )
+        throws ConfigurationException
+    {
+        ( (AbstractCGlobalProxySettingsCoreConfiguration) getCurrentCoreConfiguration() ).doValidateChanges( model );
```
kellyrob99: Does this call handle possible nulls or should this be within the if block?

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    public CRemoteHttpProxySettings convertToModel( RemoteProxySettings settings )
+    {
+        if ( settings == null )
+        {
+            return null;
+        }
+        else
```
kellyrob99: else clause really isn't required here

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/upgrade/Upgrade220to250.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        catch ( XmlPullParserException e )
+        {
+            throw new ConfigurationIsCorruptedException( file.getAbsolutePath(), e );
+        }
+        finally
+        {
+            if ( fr != null )
+            {
+                fr.close();
```
kellyrob99: Prefer call to closeQuietly() here if possible, eliminates the null check

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/DefaultRemoteStorageContext.java
```diff
@@ -141,6 +142,88 @@ public void removeRemoteProxySettings()
...
+            final String keyHttp = RemoteProxySettings.class.getName();
+            if ( hasContextObject( keyHttp ) )
+            {
+                // case 1
+                return (RemoteProxySettings) getContextObject( keyHttp );
+            }
+            else
+            {
+                // case 2
+                return (RemoteProxySettings) getContextObject( key );
```
kellyrob99: Haven't we just established what we don't have have this key in the if check above ^^^ ?

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/HttpsProxyServer.java
```diff
@@ -0,0 +1,186 @@
...
+     * @return the server
+     */
+    public Server getServer()
+    {
+        return server;
+    }
+
+    public String getUrl( String context )
+    {
+        return "http://localhost:" + getPort() + "/" + context;
```
kellyrob99: Shouldn't this be https?

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/HttpsProxyServer.java
```diff
@@ -0,0 +1,186 @@
+/*
```
kellyrob99: Not sure why we're copying this class wholesale - can't we extend and override just the methods(I can only see 2 including the http VS https comment in getUrl() below) that differ?

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/Nexus5214HttpAndHttpsProxyIT.java
```diff
@@ -0,0 +1,379 @@
...
+
+    @Before
+    public void initWebProxies()
+        throws Exception
+    {
+        repoProxy = new HttpsProxyServer();
+        repoProxy.setPort( repoProxyPort );
+        repoProxy.initialize();
+
+        globalHttpProxy = new HttpsProxyServer();
```
kellyrob99: Should this be https?

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
adreghiciu: You have to look down in the source code to see that we are not.

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
adreghiciu: This is an extracted super class from an existing already class. I did not touch/look for changing existing code if it was not for the purpose of implementing the task at hand.

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
adreghiciu: See comment above about preexisting code.

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
adreghiciu: See comment above about preexisting code.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
adreghiciu, line 219: This is a mew method that has a signature similar with a method that already exists. The other method is still there.

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/upgrade/Upgrade220to250.java
adreghiciu: You are right. My bad that I copied preexistent classes without reviewing them.

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/DefaultRemoteStorageContext.java
adreghiciu: No, getContextObject will use a parent context to get the key if key is not present, where hasContextKey will look just in current context.

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/HttpsProxyServer.java
adreghiciu: Actually I could drop this methods as they are not used. I kept them just in case that we are moving this class in jetty test suite.
Anyhow the url should not be https because the client->proxy server communication is still http, not https.
The proxy server will facilitate tunnelling HTTPS.

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/HttpsProxyServer.java
adreghiciu: We cannot override the initialize() method as it needs to access private fields.
getUrl() is not necessary to be overridden (see above comment)

nexus-test/nexus-test-harness-its/src/test/java/org/sonatype/nexus/integrationtests/nexus5214/Nexus5214HttpAndHttpsProxyIT.java
adreghiciu: No. The HttpsProxyServer (maybe name is confusing) is not about client -> proxy communication via HTTPS but about a proxy that is able to tunnel HTTPS, which the current ProxyServer is not able to do.

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/GlobalHttpsProxySettings.java
```diff
@@ -0,0 +1,22 @@
...
+package org.sonatype.nexus.configuration.application;
+
+/**
+ * @since 2.5
+ */
+public interface GlobalHttpsProxySettings
+    extends GlobalProxySettings
+{
+
+}
```
jdillon: why do we need this interface?  what is is for?

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RemoteHttpsProxySettings.java
```diff
@@ -0,0 +1,22 @@
...
+package org.sonatype.nexus.proxy.repository;
+
+/**
+ * @since 2.5
+ */
+public interface RemoteHttpsProxySettings
+    extends RemoteProxySettings
+{
+
+}
```
jdillon: why?

nexus-webapp/src/main/webapp/js/Sonatype/repoServer/ServerEditPanel.js
```diff
@@ -499,11 +499,20 @@ Ext.define('Sonatype.repoServer.ServerEditPanel', {
...
+              }),
               scope : this
             },
             'collapse' : {
-              fn : this.optionalFieldsetCollapseHandler,
+              fn : Ext.createSequence(this.optionalFieldsetCollapseHandler, function() {
+                var fieldset = self.find('name', 'globalHttpsProxySettings')[0];
+                fieldset.collapse();
+                // we need to use the DOM here because the 'checkboxToggle' is not a real Ext component
+                fieldset.checkbox.dom.disabled = true;
```
jdillon: I don't understand what this is all about :-(

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public void setBlockInheritance( boolean val )
+    {
+        if ( !isEnabled() )
+        {
+            initConfig();
+        }
+
+        getCurrentConfiguration( true ).setBlockInheritance( val );
+    }
```
jdillon: what is block inheritance code here used for?

nexus-webapp/src/main/webapp/js/Sonatype/repoServer/ServerEditPanel.js
adreghiciu: (spec discussed with rich)
Https proxy should be enabled only when http proxy is enabled.
When http proxy is disabled, the https proxy should be automatically disabled.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RemoteHttpsProxySettings.java
adreghiciu: The RemoteProxySettings.getClass().getName() is just used as a key to find the settings in RemoteStorageContext. So, I did the same for RemoteHttpsProxySettings.
I could have changed both and use different keys but that could brake existing code that looks up the RemoteProxySettings from RemoteStorageContext

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/GlobalHttpsProxySettings.java
adreghiciu: That one is similar to GlobalHttpProxySettings and is used through nexus code

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
adreghiciu: This is legacy code (I just extracted the class from DefaultGlobalHttpProxySettings in order to inherit it also in DefaultGlobalHttpsProxySettings). My guess, Tamas should know more about it) was meant to be able to inherit settings from a parent. E.g. Repository level proxy settings inherit the global level proxy settings. This means that in case that repository level ones are not present, the global ones will be used. 
Indeed it sounds a bit like over-engineering but my guess was they were meant for more then two levels.

nexus-webapp/src/main/webapp/js/Sonatype/repoServer/ServerEditPanel.js
jdillon: okay... if it works ;-)

hopefully all this js will get replaced anyway with something simple and sane one of these days.

nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/RemoteHttpsProxySettings.java
jdillon: okay, probably should add javadoc to these buggers with this details

nexus-core/src/main/java/org/sonatype/nexus/configuration/application/AbstractGlobalProxySettings.java
jdillon: k.  iirc per-repo proxy settings are something @rseddon wanted to kill and when we do that this could also go away.  though i wonder if that is the case, and I'm not imagining it, if we should consider killing that legacy now given this refactoring touches that muck?

adreghiciu: +1 on that. This will simplify quite some code.

nexus-core/src/main/java/org/sonatype/nexus/proxy/storage/remote/DefaultRemoteStorageContext.java
kellyrob99: Aaah, thanks for the clarification.

adreghiciu:
CI :+1: https://bamboo.zion.sonatype.com/browse/NXO-OSSF61-5

rseddon:
Yes, lets get rid of the repository level http proxy settings, they are broken for way too many bits of Nexus.
I've put an issue in 2.5 for this:

https://issues.sonatype.org/browse/NEXUS-5690

ifedorenko:
-1

There should be only one global non-proxy list, having two separate lists for http and https is redundant and confusing. Check OSX network configuration options, for example, or Eclipse or Windows for that matter, they all provide single non-proxy list for all protocols.

Also, as I already mentioned on hipchat, the amount of change required to introduce https proxy is way way much than I expected. In order to help me understand the reason for all this change, can you provide what hi-level changes were required? I can see changes in ui, configuration including upgrade, remote storage and tests. Anything else that I miss?

adreghiciu:
I specifically discussed the non proxy hosts with @rseddon and the issue is implemented according to that.
The pull description already mentions the high level changes. Will detail.

jdillon:
@adreghiciu can you share the reasons @rseddon wanted to have non-proxy host configuration specific to http or https proxies (and not a single configuration to be used by both)?  or maybe @rseddon can explain?  I don't understand why we would need this either.

adreghiciu:
CI :+1: https://bamboo.zion.sonatype.com/browse/NXO-OSSF61-8

nexus-configuration-model/src/main/mdo/nexus.xml
```diff
@@ -298,13 +313,50 @@
...
+          <version>2.5.0+</version>
+          <description>The global HTTPS Proxy settings are settings applied across Nexus.</description>
+          <association>
+            <type>CRemoteHttpProxySettings</type>
+          </association>
+          <required>false</required>
+        </field>
+        <field>
+          <name>nonProxyHosts</name>
+          <version>1.4.1+</version>
```
jdillon: this should be 2.5.0+ since its on a class which was just introduced as version 2.5.0+ yes?

nexus-core/src/test/filtered-resources/org/sonatype/nexus/configuration/upgrade/nexus-101.xml.result
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <nexusConfiguration>
-  <version>2.2.0</version>
+  <version>${nexus.configuration.version}</version>
```
jdillon, line 3: why are you using a version here?  this will make the tests using these very brittle when this changes and the xml files are not also updated to have the correct content a the target version.

FTR I hate all of these damn xml files to feed tests (which I also hate).  But until we can actually fix this problem I think introducing a variable here will only make maintenance of this more problematic and cumbersome.

IMO it would be better to live with the duplication of version information as it was, instead of making it trivial to break lots of tests unknowingly.

ifedorenko:
-0. This change is way too big, I can't do meaningful codereview. I think it should be split at least in two, one to remove per-repository http proxy, the other to introduce https proxy support.

To be clear, this is not a veto, the change can be merged if at least two other developers vouch for it, as is usually required for any nontrivial change.

nexus-configuration-model/src/main/mdo/nexus.xml
adreghiciu: Fixed

nexus-core/src/test/filtered-resources/org/sonatype/nexus/configuration/upgrade/nexus-101.xml.result
adreghiciu, line 3: I did this change only to avoid changing all of this xml's every time we increase nexus configuration version.
By using a variable there is only one place to change, in the pom. Using hardcoded version you have to change about 16 files.

I'm not strong on this, whatever we decide.

nexus-core/src/main/java/org/sonatype/nexus/apachehttpclient/Hc4ProviderBase.java
```diff
@@ -214,21 +215,50 @@ else if ( ras instanceof UsernamePasswordRemoteAuthenticationSettings )
...
+    /**
+     * @since 2.5
+     */
+    protected void configureProxy( final DefaultHttpClient httpClient,
+                                   final RemoteProxySettings remoteProxySettings)
     {
-        if ( remoteProxySettings.isEnabled() )
+        if ( remoteProxySettings != null
+            && remoteProxySettings.getHttpProxySettings() != null
+            && remoteProxySettings.getHttpProxySettings().isEnabled() )
```
cstamas, line 226: This IF makes impossible to set HTTPS proxy if HTTP proxy is not set. Is this intentional?

adreghiciu, line 226: Yes. 
If there is no HTTP proxy then there is no HTTPS proxy.
If there is HTTP proxy and no HTTPS proxy, the HTTP proxy is also used for HTTPS.

The UI also reflects this.

cstamas:
Looks okay, but the amount is huge. Went thru relevant parts, while huge chunk of changes (like those UT related) left to CI to sort out :)

+1

adreghiciu:
Another +1 from CI after latest changes: https://bamboo.zion.sonatype.com/browse/NXO-OSSF61-10

rseddon:
I think we've missed the window of opportunity for 2.5, this will need to wait for 2.6.

But I'm going to test this on an isolated network right now anyhow.  If I don't see any problems I'll ask Anders if he wants to try it out at the site that needs this.

PR #836, opened 2013-04-22
REVIEW [NEXUS-5599] Warn if browser is <IE8

nabcos:
Opens a modal dialog so that the user has to acknowledge that they are using an unsupported browser.

Link goes to https://support.sonatype.com/entries/22926682-Nexus-2-3-Internet-Explorer-Compatibility, this should probably be a redirect from links.sonatype.com

### DIFF
--- a/plugins/restlet1x/nexus-restlet1x-plugin/src/main/resources/templates/index.vm
+++ b/plugins/restlet1x/nexus-restlet1x-plugin/src/main/resources/templates/index.vm
@@ -79,6 +79,10 @@
                         }
                         alreadyDone = true;
 
+                        if (Ext.isIE6 || Ext.isIE7) {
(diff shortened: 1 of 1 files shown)


kellyrob99:
+1

jdillon:
minor, "knowledge base" probably should be "knowledge base article"

+1

jdillon:
tested with winxp ie6 and I see the dialog.

link should probably use a different target frame.

links.\* url does not actually take me to the kb article.

related... the article here isn't super helpful to non-technical users (which I'd expect would be the case for folks who run into this).

+1 on this change, though we need to follow through and get the link + kb content updated.

nabcos:
I pinged @jswank about setting up the redirect and documented it on https://docs.sonatype.com/display/ProdMgmt/External+Links+and+Redirects+for+Sonatype+Commercial+Products

If the redirect is not in place yet, this might be the right time to update the target URL.

adreghiciu:
+1

PR #837, opened 2013-04-22
[NEXUS-267] default to sortable columns

nabcos:
This change will make grid columns sortable for those views in Nexus which extend the GridViewer (e.g. security/user, sec/privileges, plugin console, ...)


jdillon:
cursory testing of oss build shows nothing broken.

+1

adreghiciu:
+1

PR #838, opened 2013-04-22
REVIEW [NEXUS-5094] Purging snapshot task should take into account classifier

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-5094

Just partition the files we're considering by a (possible) classifier before applying the algorithm which decides what to delete


nexus-core/src/main/java/org/sonatype/nexus/maven/tasks/DefaultSnapshotRemover.java
```diff
@@ -13,10 +13,10 @@
 package org.sonatype.nexus.maven.tasks;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.commons.lang.StringUtils.*;
```
jdillon, line 16: should generally avoid using star imports so that its clear what is being imported

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-1

kellyrob99:
+1 by CI again after merging in trunk

ifedorenko:
-1. This chance have to have some sort of regression test.

kellyrob99:
+1 by CI with new IT https://bamboo.zion.sonatype.com/browse/NXO-OSSF39-3

ifedorenko:
+1 I don't see any unexpected changes.

PR #839, opened 2013-04-23
Nexus Kenai Realm Plugin

cstamas:
While this realm is not shipped with any of distros
it's OSS plugin for Nexus, hence it built with Nexus OSS.

Related:
https://issues.sonatype.org/browse/NXCM-4884

To buy out "yet another git repo" (that actually holds few lines only):
https://github.com/sonatype/nexus-kenai


cstamas:
Note: this pulls only adds the nexus-kenai-plugin to the build, but it DOES NOT USE IT (packs into any bundle). That's intentional, as we never shipped this Realm, it's "more optional" than those "optional" plugins we already have.

jdillon:
why don't you move the muck from security in here too?  I don't know of any other _real_ use of kenai realm muck except for use in java.net's nexus.

plugins/kenai/nexus-kenai-testsuite/src/test/java/org/sonatype/nexus/kenai/testsuite/KenaiIT.java
```diff
@@ -0,0 +1,32 @@
+/*
```
kellyrob99, line 1: Not sure what the point of keeping this empty class is...

jdillon, line 1: agree... either add real its for this, or defer adding this module until someone makes them.  No point in making a module unless you plan to use it.

cstamas, line 1: I intend to implement the "teting doco" (see wiki for Kenai testing) as a simple IT. Will come to it later, was distracted by looking for proper place for this plugin first... :)

cstamas:
@jdillon I might move the muck from security here, but then, why not move _all of the security_ into nexus? And then we come to the question you raised last week, about "build minification". We should probably discuss this over the phone with whole team involved.

plugins/kenai/nexus-kenai-testsuite/src/test/java/org/sonatype/nexus/kenai/testsuite/KenaiIT.java
cstamas, line 1: IT implemented

cstamas:
Realm moved in

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF41-1

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF41-2

plugins/kenai/nexus-kenai-testsuite/src/test/java/org/sonatype/nexus/kenai/testsuite/KenaiIT.java
kellyrob99, line 1: Thanks much!

kellyrob99:
+1

jdillon:
@cstamas btw I'm still working out how we can merge git repos best w/o loosing too much history.  its more of a pita than I had hopped, but I hope to have some sort of plan soon.

cstamas:
@jdillon sure, but this realm is 3 classes only (with the Realm being almost rewritten for GET vs POST), so I don't see the point of keeping history for Kenai at least.

For other security, yes.

jdillon:
@cstamas there is always a point to retaining history of some sort.  I don't care about the kenai realm code though, but my answer was in response to your question "why not move all of the security into nexus"

PR #842, opened 2013-04-24
REVIEW [NEXUS-5037] single quote password does not work

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-5037

Just apply existing unescape mechanism for password when creating a new user

### DIFF
--- a/nexus-client-core-testsuite/src/test/java/org/sonatype/nexus/client/testsuite/UserIT.java
+++ b/nexus-client-core-testsuite/src/test/java/org/sonatype/nexus/client/testsuite/UserIT.java
@@ -1,0 +1,36 @@
+package org.sonatype.nexus.client.testsuite;
+
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.sonatype.nexus.client.core.NexusClient;
+import org.sonatype.nexus.client.core.subsystem.security.Users;
+
+/**
+ * @since 2.5
+ */
+public class UserIT
+    extends NexusClientITSupport
+{
+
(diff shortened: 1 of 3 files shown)


jdillon:
+1 ( i hate that we have to do this, but IIUC we don't have anything better atm )

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF64-3

adreghiciu:
UT/IT checking that problem was solved?

kellyrob99:
I spent a fair bit of time trying to get a working test for this, but it's proving hard to inspect the password value during testing. Any suggestions you might have @adreghiciu would be appreciated.

jdillon:
I do not think we need a IT for this.  We've done this in the past for each tiny change and its produced the testsuite we have now that takes forever.

We should be able to know that use of HtmlUnescapeStringConverter as used in this context (since lots of other stuff doing the same thing) works properly.  We don't need to bloat the testsuite with ~30s additional time to execute with trivial stuff like this.

@kellyrob99 if you want to make a UT that is fine, though what would the UT be testing? the abstract resource or the converter used in the model?  I suggest its the converter and how its generally used in models like this.  I pray there are already UT for it.

adreghiciu:
I wold like the IT not proving that currently the solution works but more that we do not introduce a regression.
@kellyrob99 I would use teh new IT support, create a user with a password containing a single quote and then create a new nexus client using the new user + make a request to nexus using that client. if password would not work it will fail.

adreghiciu:
+1 in rest

jdillon:
/me shrugs ... this code will get thrown away in the not so distant future so way your time spent on this in that light.

kellyrob99:
New IT added and +1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF64-5

PR #845, opened 2013-04-29
[REVIEW] NXCM-5067: Setting LDAP connection stuff in JVM props

cstamas:
Setting the Sun JNDI default implementation related properties.

Also, removing the dead code, as according to Sun documentation, the pooling might be set _only by system properties_ (unlike putting key/values in env map as done here), _affects globally all contexts_ (unlike code here expected to set per context).

Still, setter for polling left (and does nop), to adhere javadoc comments and to retain API compatibility.


adreghiciu:
+1

Why not just drop setter/getter and field? Is just confusing...

nexus-oss-webapp/src/main/resources/content/bin/jsw/conf/wrapper.conf
```diff
@@ -27,11 +27,13 @@ wrapper.java.library.path.1=bin/jsw/lib
 
 # Additional JVM parameters (tune if needed, but match the sequence of numbers!)
 wrapper.java.additional.1=-Djava.net.preferIPv4Stack=true
-#wrapper.java.additional.2=-Xdebug
-#wrapper.java.additional.3=-Xnoagent
-#wrapper.java.additional.4=-Djava.compiler=NONE
-#wrapper.java.additional.5=-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
-#wrapper.java.additional.6=-XX:+HeapDumpOnOutOfMemoryError
+wrapper.java.additional.2=-Dcom.sun.jndi.ldap.connect.pool=true
```
kellyrob99: Should these be turned on by default and were they supposed to be on by default previously?

cstamas: [This issue](https://issues.sonatype.org/browse/NEXUS-5064), (linked to "this" one NXCM-5067) done in 2.1 was meant to enable connection pooling for ssl connections, but it did it wrong it seems. So I believe answer is "yes".

jdillon: If this can live in nexus.properties (or default.properties) it should be there not here.

peterlynch: I don't suggest adding any of these in nexus.properties, since JVM system properties are usually read ( and confirmed in this case)  into static variables and 'when' is not predictable.

jdillon: Dude... the point of nexus.properties was to have a simple place to put ... um... system properties so that users don't have to add this .2 shit to wrapper conf.

This is loaded very early and is appropriate for most system properties.

FTR I'm only talking about this one right one:

```
wrapper.java.additional.2=-Dcom.sun.jndi.ldap.connect.pool=true
```

This should be able to live in nexus.properties with out any problem IIUC.

peterlynch: Agree some, but not any of these. We cannot guarantee the class loading order in the JVM and when static variables will get populated.

jdillon: You could be right... I don't know :-P

If its required here, the we have to put it here, but if its not then we shouldn't.

System properties should be loaded very very early.  If ldap classes are getting initialized before then... well that is a problem in itself.

peterlynch: @cstamas actually 'com.sun.jndi.ldap.connect.pool' is an LDAP context "environment" property - not a system property. 

It is set [here](https://github.com/sonatype/nexus/blob/master/plugins/ldap/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java#L279).

It is read [when making a connection](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/com/sun/jndi/ldap/LdapCtx.java#189).

I think this means we should remove the setting of this property in wrapper.conf.

---

Issue NXCM-5067 is about SSL ldap connections not being pooled. The protocol selection is determined in a static block by [LdapPoolManager.java](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/com/sun/jndi/ldap/LdapPoolManager.java#148) reading the _system_ property 'com.sun.jndi.ldap.connect.pool.protocol' as you have stated. Therefore

> wrapper.java.additional.3=-Dcom.sun.jndi.ldap.connect.pool.protocol=plain ssl

should be changed to 

> wrapper.java.additional.2=-Dcom.sun.jndi.ldap.connect.pool.protocol=plain ssl

after removing 

> wrapper.java.additional.2=-Dcom.sun.jndi.ldap.connect.pool=true

plugins/ldap/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java
```diff
@@ -272,16 +263,6 @@ public LdapContext getLdapContext(String username, String password, boolean syst
         env.put( Context.PROVIDER_URL, url );
         env.put( Context.REFERRAL, referral );
 
-        // Only pool connections for system contexts
-        if ( usePooling && username != null && systemContext )
-        {
-            // Enable connection pooling
-            env.put(SUN_CONNECTION_POOLING_PROPERTY, "true");
```
peterlynch, line 279: I think removing this is wrong - unless I am missing some larger context to how this is called.

SUN_CONNECTION_POOLING_PROPERTY is a context property. Not a system property. See other comment.

plugins/ldap/ldap-common/src/main/java/org/sonatype/security/ldap/realms/DefaultLdapContextFactory.java
```diff
@@ -272,16 +263,6 @@ public LdapContext getLdapContext(String username, String password, boolean syst
...
-            // Enable pooling for plain and ssl connections
-            env.put( SUN_CONNECTION_POOLING_PROTOCOL_PROPERTY, "plain ssl" );
```
peterlynch, line 281: This makes sense removing since it is a system property.

cstamas:
@peterlynch is right, one of the removed properties _is system property_ while another _is env property_. I find it very confusing, and related doco seems very confusing too (by using almost same keys making hard to distinguish among them). 

I was reading the "tutorial" (and some stackoverflow questions about similar problems), and there were statements like _"You configure connection pooling by using a number of system properties at program startup time. Note that these are **system properties, not environment properties** and that they affect all connection pooling requests."_ from [tutorial](http://docs.oracle.com/javase/jndi/tutorial/ldap/connect/config.html) made me think that _both are system properties_.

So, to recap, this pull is wrong.
- `com.sun.jndi.ldap.connect.pool.protocol` is a System property (in code being changed it was used as env property, hence had no effect, LDAPS connections were NOT pooled)
- `com.sun.jndi.ldap.connect.pool` is an Environment property (in code being changed it was used correctly, this pull would break it).

Closing this pull and will fix it.

cstamas:
Fixes applied, thanks @peterlynch!

wrapper.conf removed the one wrongly added entry, code fixed and env property use returned. UT returned.

PR #846, opened 2013-04-29
TRIVIAL [NXCM-4665]- apply RepositoryTarget to individual files as well as GAV (folders)

kellyrob99:
https://issues.sonatype.org/browse/NXCM-4665

NXCM-4665 - Scheduled Task to remove releases

Tamas pointed out that this task was only working at the GAV or folder level and ignoring potential for a RepositoryTarget to match against individual files. While this is a pretty small edge case, it's also easy to accomodate so here's the impl and an associated IT.


nexus-core/src/test/java/org/sonatype/nexus/maven/tasks/DefaultReleaseRemoverIT.java
```diff
@@ -60,12 +65,12 @@ public void testDeletionOfReleases()
...
             releaseRemover.removeReleases( new ReleaseRemovalRequest( releases.getId(), 2, "" ) );
         // pom + jar + sha1 for both
-        assertThat( releaseRemovalResult.getDeletedFileCount(), is( 4 ) );
+        assertThat( releaseRemovalResult.getDeletedFileCount(), is( 6 ) );
         assertThat( releaseRemovalResult.isSuccessful(), is( true ) );
         StorageItem item = null;
         try
         {
-            releases.retrieveItem( new ResourceStoreRequest( "org/sonatype/test/1.0" ) );
+            item = releases.retrieveItem( new ResourceStoreRequest( "org/sonatype/test/1.0" ) );
```
kellyrob99, line 73: Found this error in previous test logic and fixed it as well

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF19-1

kellyrob99:
Updated as per testing feedback, this should be retested to ensure no regressions have been introduced.

PR #847, opened 2013-04-30
[REVIEW] NEXUS-5511 Refine reasoning

cstamas:
Original fix for NEXUS-5511, but left some loose endings.

Conditions like "request is local-only", "proxy mode is not ALLOW" and "rejected by auto-routing" were all put under same message stating "...remote storage access is prevented...". Clearly, this is no good.

This pull separates now these scenarios, but also improves the messages by aligning how repository is "rendered" (unifying the message).

See issue for "validation scenario" where new (but also old) reasoning messages are represented:
https://issues.sonatype.org/browse/NEXUS-5511

### DIFF
--- a/nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
+++ b/nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
@@ -1215,8 +1215,7 @@ public abstract class AbstractRepository
                     }
 
                     throw new ItemNotFoundException( reasonFor( request, this,
-                        "The path %s is in NFC of repository %s.", request.getRequestPath(),
-                        RepositoryStringUtils.getHumanizedNameString( this ) ) );
+                        "The path %s is in NFC of repository %s.", request.getRequestPath(), this ) );
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


nexus-core/src/main/java/org/sonatype/nexus/proxy/repository/AbstractRepository.java
```diff
@@ -1215,8 +1215,7 @@ public void maintainNotFoundCache( ResourceStoreRequest request )
                     }
 
                     throw new ItemNotFoundException( reasonFor( request, this,
-                        "The path %s is in NFC of repository %s.", request.getRequestPath(),
-                        RepositoryStringUtils.getHumanizedNameString( this ) ) );
+                        "The path %s is in NFC of repository %s.", request.getRequestPath(), this ) );
```
nabcos: minor: don't use acronym for NFC? e.g. "The path %s is still cached as not found for repository %s."

cstamas: Sounds good.

nabcos:
+1

PR #848, opened 2013-04-30
REVIEW: NXCM-4867 - Index update concurrency problem

scarlucci:
https://issues.sonatype.org/browse/NXCM-4867

These changes fix a synchronization problem caused by attempting an incremental index update when an incremental update is not possible, and a full index update must be performed. In this case, since the maven indexer did not support a way to request an incremental-only update, it was not possible to know the appropriate locking mechanism to use for the update.

Previously, we were grabbing a read-only lock to perform the incremental update. But in cases where an incremental was not possible, and a full index update was performed, there was a window towards the end of the full index update where another thread attempting to read the index could experience an NPE.

The maven indexer has been updated to allow clients to specify an incremental-only index update. These changes take advantage of this new support. If an incremental update is not possible, then a full index update is performed with the appropriate locking

CI build on the way


plugins/indexer/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
```diff
@@ -2800,4 +2858,14 @@ public IndexingContext getRepositoryIndexContext( String repositoryId )
     {
         return mavenIndexer.getIndexingContexts().get( getContextId( repositoryId ) );
     }
+
```
ifedorenko, line 2886: I recommend changing exception class name to IndexUpdateException. Please also mark the class as static as it does not and should not need access any member of the outer class.

ifedorenko:
+1

Couple of cosmetic comments, but I don't feel too strong about them so feel free to ignore.

plugins/indexer/nexus-indexer-lucene-plugin/pom.xml
```diff
@@ -30,8 +30,7 @@
   <properties>
     <pluginName>Nexus Indexer Lucene Plugin</pluginName>
     <pluginDescription>Adds search capabilities for repository content.</pluginDescription>
-
-    <mavenIndexer.version>5.1.2-c65990c</mavenIndexer.version>
+    <mavenIndexer.version>5.1.2-bf3d48b</mavenIndexer.version>
```
kellyrob99, line 33: Just out of curiousity, why the strange version number?

cstamas, line 33: These are "pseudo" releases of ASF Maven Indexer. ASF MI 5.1.1 was released last, codebase is currently on 5.1.2-SNAPSHOT. The "bf3d48b" suffix is basically the Git sha1 commit hash out of which the pseudo release of MI was done.

plugins/indexer/nexus-indexer-lucene-plugin/src/main/java/org/sonatype/nexus/index/DefaultIndexerManager.java
scarlucci, line 2886: Changed the class to be static. Left the name to be as explicit as possible about the conditions under which this exception is used, but can be convinced to change if someone has a strong opinion about this name

kellyrob99:
+1

ifedorenko:
+1

Please merge this if the feature build is successful

PR #851, opened 2013-05-03
RFC: Add support for application/vnd.json-jackson to use Jackson

jdillon:
Nexus-side to this change:

https://github.com/sonatype/sisu-restlet-bridge/pull/7


cstamas:
I'd add a Filter too (for very similar example of overriding MIME types see the [browser sensing filter](https://github.com/sonatype/nexus/blob/master/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/BrowserSensingFilter.java)) on top of all Restlet REST resources, to have one single spot controlling used MIME type.... 

It could be a trivial Filter sensing "new" client/UI and making it use vnd.json-jackson or in case of "old" client/UI simply send it to vnd.json-xstream.

jdillon:
@cstamas I don't understand why a filter would help here.  Current clients all use 'Accept: application/xml' which maps the the restlet xstream format.  I don't see any point at all in magically transforming this to another media type for those clients.

FTR we'll probably have to retain the xstream format via application/xml so that newer clients can still talk to older servers, but I've ripped it out atm so that I could be used that it was not being used anywhere, and also to find any places which are currently hardcoded to use xstream (and there are a few unfortunately).

jdillon:
closing lets kill restlet instead of doing this

PR #852, opened 2013-05-06
[REVIEW] Making Nexus570IndexArchetypeIT stable.

cstamas:
Changes:
- As legacy ITs already had (unsatisfied) expectation about
  "sync" nature of Indexer related operations, added a flag
  to IndexerManagerEventInspector that makes it "blocking" or
  "async" depending on a system property. To be used in ITs.
- removed obsolete "mavenIndexerBlockingCommits" property
  from Jetty8NexusBooter, as code handling this was removed along
  with other notable changes happened in MI 5.x. Simply put,
  this property had no effect since then.
- Using the newly added key in Jetty8NexusBooter, to make
  Index related event inspector become "blocking".

### DIFF
--- a/nexus-test/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty8NexusBooter.java
+++ b/nexus-test/nexus-test-harness-launcher/src/main/java/org/sonatype/nexus/test/booter/Jetty8NexusBooter.java
@@ -147,15 +147,8 @@ public class Jetty8NexusBooter
         // guice finalizer
         System.setProperty( "guice.executor.class", "NONE" );
         
-        // Note: in ITs we want to make Indexer perform blocking commits.
-        // Since MavenIndexer 4.0, it performs async commits by default, meaning that no "helper" from Nexus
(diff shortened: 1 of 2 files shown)


ifedorenko:
+0. Probably makes the test more reliable, but looks backwards. If the test needs index to be in certain state, why not explicitly call DefaultIndexerManager#addRepositoryIndexContext as part of the test?

cstamas:
@ifedorenko this test actually tests a "successful transition": it deploys two artifacts, one "plain jar" and one with packaging "maven-archetype", and then performs search to verify that MI did "pick up" the fact that one of them is an Archetype. And from these ITs, doing DefaultIndexerManager#addRepositoryIndexContext would be possible only if this IT would have some server side code (plugin?) support, as IT is able to reach REST API only, but IMO that would be an overkill for this test.

ifedorenko:
I thought old-style ITs could use nexus components directly.

cstamas:
@ifedorenko no, none of the ITs are able to reach anything else thab REST API only. The main difference is that "old-style" ITs were running Nexus in-process but in isolated classloader, while "new-style" ITs are running Nexus as separate process.

PR #854, opened 2013-05-07
[REVIEW] NEXUS-5710: Status and Content CSS files were not loaded

cstamas:
The URL generated by template (content and status template)
were always wrong, it contained "//" between nexusRoot
and rest of the path info.

But the changes how the resource is served up now
made the "//" appear as error, Restlet refuses
to serve it up (was 404).

The two places of use of getContextRoot(req) where
templates are evaluated are fixed (third place related
to openSearch already had a fix for this), as all the
other users of above mentioned method EXPECT that
returned reference contains path info ending with "/",
unlike the cases related to Velocity templates.

### DIFF
--- a/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
+++ b/plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
@@ -597,7 +597,13 @@ public abstract class AbstractResourceStoreContentPlexusResource
 
             dataModel.put( "nexusVersion", getNexus().getSystemStatus().getVersion() );
 
-            dataModel.put( "nexusRoot", getContextRoot( req ).toString() );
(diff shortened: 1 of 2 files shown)


adreghiciu:
+1

plugins/restlet1x/nexus-restlet1x-plugin/src/main/java/org/sonatype/nexus/rest/AbstractResourceStoreContentPlexusResource.java
```diff
@@ -597,7 +597,13 @@ protected Representation serialize( Context context, Request req, Variant varian
 
             dataModel.put( "nexusVersion", getNexus().getSystemStatus().getVersion() );
 
-            dataModel.put( "nexusRoot", getContextRoot( req ).toString() );
+            // getContentRoot(req) always returns Reference with "/" as last character
```
kellyrob99, line 600: minor but comment refers to getContentRoot() and actual method is getContextRoot()

kellyrob99:
A little late to the party, but is there a place this duplicated code could be pulled up to and shared?

PR #855, opened 2013-05-13
[RFC for 2.5.0-02 release]- include log configuration changes to show beginning of thread name

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-5712
 change logging pattern layout to include first 15 characters of thread name instead of last 15

### DIFF
--- a/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundleConfiguration.java
+++ b/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundleConfiguration.java
@@ -65,7 +65,7 @@ public class DefaultNexusBundleConfiguration
     /**
      * Default logging pattern.
      */
-    private static final String DEFAULT_LOG_PATTERN = "%4d{yyyy-MM-dd HH:mm:ss} %-5p [%-15.15t] - %c - %m%n";
+    private static final String DEFAULT_LOG_PATTERN = "%4d{yyyy-MM-dd HH:mm:ss} %-5p [%-15.-15t] - %c - %m%n";
 
     /**
      * File task builder.
--- a/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundleConfiguration.java
+++ b/nexus-launcher/src/main/java/org/sonatype/nexus/bundle/launcher/support/DefaultNexusBundleConfiguration.java
@@ -65,7 +65,7 @@ public class DefaultNexusBundleConfiguration
     /**
      * Default logging pattern.
      */
-    private static final String DEFAULT_LOG_PATTERN = "%4d{yyyy-MM-dd HH:mm:ss} %-5p [%-15.-15t] - %c - %m%n";
+    private static final String DEFAULT_LOG_PATTERN = "%4d{yyyy-MM-dd HH:mm:ss} %-5p [%thread] - %c - %m%n";
 
     /**
      * File task builder.
--- a/nexus-logging-extras/src/main/resources/META-INF/log/logback.properties
+++ b/nexus-logging-extras/src/main/resources/META-INF/log/logback.properties
@@ -12,5 +12,5 @@
 #
 
 root.level=INFO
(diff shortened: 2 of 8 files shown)


kellyrob99:
There was some discussion on the Jira about whether or not we should increase the size of what we log for thread name, which I'm not sure is entirely necessary.
If we're not overly concerned with increasing size of debug logs we could remove the size restriction and log the entire url, but I'm not convinced it's worth the additional verbosity.
Any votes in either direction?

ifedorenko:
Is this user-configurable?

jdillon:
I'd rather have the full thread name... then again I'd also like to not have the thread names change too; I hate that technique.

/me shrugs

kellyrob99:
With these settings we get thread names like:
[jetty-main-thre]
[qtp1880441913-8]
[ar-4-thread-1  ]
[qtp1880441913-7]

jdillon:
FTR I still don't like truncating detail like this, even if some libraries crap on the thread names

removed my vote, as I think this just moves the problem it does not resolve it.

peterlynch:
See my comments on issue basically it looks like it should go to 18 chars
to pick up pxpool-1-thread-1 at least.

jdillon:
@peterlynch why stop at 18?  we have threads with names like:
- BrokerService[BE9B2656-17DDA7F8-C265D039-0CA0D616-9FEB556E] Task-35081

By changing the pattern from how it was before, you make it easier to identify threads which tack on junk dynamically (bad libraries for doing that) and make it harder to identify threads which actually have the important detail near the ends.

If I am looking at a log to try and debug what is going on, I need all the information not just the first few chars or the last few chars.

peterlynch:
Sounds like the solution is just change the pattern layout for the one
affected logger only - ok?

jdillon:
@peterlynch layouts are for appenders, not for loggers.  to change the layout for a single logger would add significant complexity to the logback configuration.

kellyrob99:
If we provide no limitation on the size of the thread name we get this:
`jvm 1    | 2013-05-13 19:48:58 DEBUG [qtp1380189880-71 - /nexus/content/groups/public/org/apache/maven/shared/maven-shared-components/17/maven-shared-components-17.pom]`

If our intention in making this change is to capture the actual thread name information for jetty requests then we need 17-18 characters from the beginning of the thread name. Presently the actual thread name is being truncated at 15, regardless of DEBUG setting.

The alternative is to remove the truncation completely and include the uri in the output, and I'm kind of inclined to go that way. Debug is meant to provide more information, and this tells us exactly what is being requested in that context. Seems like a win-win.

What does the rest of the team think?

cstamas:
Just to recap and remind us why are we did this:
- Jetty in DEBUG _changes_ how threads are named, by doing `threadQtpName + requestUri`. At INFO level Jetty QTP threads are named only with `threadQtpName`.
- DEBUG logs are hard to follow due to this, as Jetty QTP thread names are "clipped", only the last few characters of request URI are visible (currently impossible to distinguish two threads processing same URI)
- we have no other problems in this regard, so again, only _Jetty threads only in DEBUG mode_ (Jetty logger in DEBUG level).

But, there is a commit
https://github.com/sonatype/nexus/commit/d0745ecdbc371b298bdfcea4689a007d77494b9e

that makes this a non-issue anymore (well, at least when Nexus put into DEBUG over UI, as Jetty will remain INFO). In fact, I see no reason why would we want to run Jetty itself at DEBUG level (as happened before, in older versions of Nexus).

Still, just FTR the related Jetty code is here (from 8.1.8 release we currently use):
https://github.com/eclipse/jetty.project/blob/3a99c0cea082aa6baebcb8fd1607c954444e435a/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractHttpConnection.java#L430

So, my stance now is simply to provide no limitation on size of the thread name (last @kellyrob99 comment).

peterlynch:
I could live with no limitation to thread name size - although I will defer to @rseddon as to if this is OK from a support perspective.

I still think many of our thread names we control are poorly named.

rseddon:
I'd rather have no limitation on thread name size.  More information is better.  And now that we have compression enabled for rotated logs the larger size will be less of an issue.

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF1-4

PR #856, opened 2013-05-14
[REVIEW] NXCM-5067 jsw quote fix

kellyrob99:
Add a directive to the Java Service Wrapper to strip quotes for a JVM property on *nix.

https://issues.sonatype.org/browse/NXCM-5067

### DIFF
--- a/nexus-oss-webapp/src/main/resources/content/bin/jsw/conf/wrapper.conf
+++ b/nexus-oss-webapp/src/main/resources/content/bin/jsw/conf/wrapper.conf
@@ -28,12 +28,12 @@ wrapper.java.library.path.1=bin/jsw/lib
 # Additional JVM parameters (tune if needed, but match the sequence of numbers!)
 wrapper.java.additional.1=-Djava.net.preferIPv4Stack=true
 wrapper.java.additional.2=-Dcom.sun.jndi.ldap.connect.pool.protocol="plain ssl"
-wrapper.java.additional.3.stripquotes=TRUE
-#wrapper.java.additional.4=-Xdebug
-#wrapper.java.additional.5=-Xnoagent
-#wrapper.java.additional.6=-Djava.compiler=NONE
-#wrapper.java.additional.7=-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
-#wrapper.java.additional.8=-XX:+HeapDumpOnOutOfMemoryError
+wrapper.java.additional.2.stripquotes=TRUE
+#wrapper.java.additional.3=-Xdebug
+#wrapper.java.additional.4=-Xnoagent
+#wrapper.java.additional.5=-Djava.compiler=NONE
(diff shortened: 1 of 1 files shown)


kellyrob99:
I'm also testing the nexus.properties alternative out on a separate branch to ensure we don't see the same kind of problems on Windows that led to the quotes in the first place.
We can discuss as a team which alternative is preferable once CI has its say.

cstamas:
CI liked both solutions:

This pull "NXCM-5067_jswQuoteFix":
https://bamboo.zion.sonatype.com/browse/NXO-OSSF6-2

And the branch "NXCM-5067_jswQuoteFix2"
https://bamboo.zion.sonatype.com/browse/NXO-OSSF21-1

Now only to validate that both does what it should: adds pooling to LDAPS while retaining it for plain LDAP connection.

Personally I have no preference either way, but I do "feel" like wrapper is the proper place to do this, as nexus.properties is dragged by WAR also, and doing JVM-wide config from a WAR sounds awkward to me, that's the responsibility of the one starting the java process itself (in case of bundle, the JSW).

ifedorenko:
I am not sure I understand concerns about user changes to wrapper.conf. The proposed change only affects single additional parameter and does not change how jsw treats user customizations of the file. If anything, it'll show users how to customize wrapper.conf properly, which is a good thing. What am I missing?

peterlynch:
This change looks good to me

+1

jdillon:
@ifedorenko the point I was making was that its very easy to fuck up the indexes, so easy that we who know how this stuff works fucked it up, so users are almost certainly going to fuck it up too.  Its no longer a matter of simply going line by line an incrementing the index number, but you have to be aware of what each line is.

nexus.properties was setup to avoid this complication and I don't know why you folks are pushing so hard for the complex property configuration in wrapper.conf.

anyways, I'm 100% sure this will cause support problems

ifedorenko:
I agree jsw will cause support problems. I do not see how this particular change make things worse.

jdillon:
just means folks will have to be aware when reindexing that .2 has two lines which need to have the same index.  folks that aren't already intimately aware of this shit config format will almost certainly not catch this and will mess up the configuration that will cause ldap pooling configuration to stop working again for ssl.  We still tell folks to put properties in this file and then reindex the wraper.java.additional.\* lines.

whatever... this is obvious to me, and I don't see how this isn't obvious to everyone else... blah

do what you want, seems like nexus does thrive on unneeded complexity anyways, so this change fits that norm.

kellyrob99:
I think that we should discuss when/how to get rid of jsw completely if we see it as that much of a problem. 
Ideally all of the jvm parameters should be specified in the same place, but so long as we utilize jsw this seems to be the correct place to ensure that we don't have any potential race conditions when setting this property.
We should definitely discuss how to approach deprecation of jsw usage as part of 2.6 activities, whether or not we agree to undertake the work in that timeframe.

PR #858, opened 2013-05-14
[2.6] NEXUS-5704: 500 Internal Server Error when "If-None-Match" in header

cstamas:
Restlet 1.1 is very picky (and have a bug) how it handles ETags: it requires "by the book" how HTTP request formats ETags, they must be quoted. In other case, the header presence is detected (condition list is non-empty), but unexpected `null` value is returned for tag value (the condition list will have one element, which is `null`). This causes NPE in Nexus code.

"Vanilla" Nexus (even without this change) _works properly_ for HTTP conditional GETs using `if-modified-since` headers, as shown in the linked issue. Still, with unquoted ETag NPE does happens in Nexus codebase, and this pull fixes it.

This pull merely fixes that single point of NPE by adding nullcheck, and also simplifies the `renderStorageFileItem()` method, but DOES NOT fix the Restlet library in this manner. Nexus will still NPE if HTTP request carries badly formatted ETag. After this change, badly formatted ETags will continue to cause 500 Internal Server Error responses, but this time deeply in Restlet code (where also Tag inspection is tried, without nullchecks around it).

Relates to issue
https://issues.sonatype.org/browse/NEXUS-5704

### DIFF
--- a/plugins/restlet1x/nexus-restlet1x-plugin/src/test/java/org/sonatype/nexus/rest/ResourceStoreContentPlexusResourceTest.java
+++ b/plugins/restlet1x/nexus-restlet1x-plugin/src/test/java/org/sonatype/nexus/rest/ResourceStoreContentPlexusResourceTest.java
@@ -12,24 +12,35 @@
  */
 package org.sonatype.nexus.rest;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.sameInstance;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.Map;
 
 import com.google.common.collect.Maps;
 import com.noelios.restlet.http.HttpResponse;
 import com.noelios.restlet.http.HttpServerCall;
+
+import org.hamcrest.Matcher;
+import org.hamcrest.MatcherAssert;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
 import org.restlet.Context;
+import org.restlet.data.ClientInfo;
 import org.restlet.data.Conditions;
 import org.restlet.data.Form;
 import org.restlet.data.Reference;
 import org.restlet.data.Request;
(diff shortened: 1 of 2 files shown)


ifedorenko:
Is it hard to unit test this change? And I do mean "unit test", I don't believe full-blown integration test is a good idea in this case.

cstamas:
I believe we have UTs for some resources, will take take a peek into those, and how hard would be to cover this change...

cstamas:
Added UT that verifies the two spots changed.

cstamas:
CI
https://bamboo.zion.sonatype.com/browse/NXO-OSSF25-1

plugins/restlet1x/nexus-restlet1x-plugin/src/test/java/org/sonatype/nexus/rest/ResourceStoreContentPlexusResourceTest.java
```diff
@@ -214,4 +239,111 @@ public void testNexus5155OmitCacheHeadersForLinkedFileItems()
...
+    {
+        try
+        {
+            final ResourceStoreRequest rsr = fileItem.getResourceStoreRequest();
+            rsr.setIfNoneMatch( "{SHA1{1234567890}}" );
+            when( attributes.containsKey( StorageFileItem.DIGEST_SHA1_KEY ) ).thenReturn( true );
+            when( attributes.get( StorageFileItem.DIGEST_SHA1_KEY ) ).thenReturn( "1234567890" );
+
+            underTest.renderStorageFileItem( request, fileItem );
+            assertThat( "ResourceException was expected to be thrown", false );
```
kellyrob99: The intent of the test would probably be clearer here if you just used fail() with the same message

kellyrob99:
So IIUC the particular user that reported this issue would still have to properly format their Etag in order to use it successfully, but at least now they won't get an NPE?

cstamas:
CI +1'd
https://bamboo.zion.sonatype.com/browse/NXO-OSSF25-1

cstamas:
@kellyrob99 Will change the UT to make it more clearer.

And yes, Restlet framework suffers from "unquoted ETag" bug, but after this fix the nasty NPE will not happen in Nexus codebase, but even with this, users will still have to "properly" format ETags.

adreghiciu:
+1

cstamas:
Moved to collapsed project as https://github.com/sonatype/nexus-oss/pull/8

PR #860, opened 2013-05-16
[REVIEW] NXCM-5302 Protect against staging.xml corruption

kellyrob99:
Enhance existing code to provide temp file handling for storing configuration files.

https://issues.sonatype.org/browse/NXCM-5302 - Protect against staging.xml corruption

Nexus already had interfaces for this use-case, although I couldn't find any existing implementations. I did rename a misspelled interface so if anyone knows of code that implemented ConfigurationWritter, please speak up now :)
I did try to confirm first by searching https://github.com/search?q=ConfigurationWritter&ref=cmdform&type=Code


nexus-core/src/main/java/org/sonatype/nexus/util/configurationreader/DefaultConfigurationHelper.java
```diff
@@ -144,19 +145,38 @@
...
+            File tempFile = new File(configurationFile.getParentFile(), configurationFile.getName() + ".tmp");
+            File copy = new File(configurationFile.getParentFile(), configurationFile.getName() + ".old");
+
+            fw = new OutputStreamWriter( new FileOutputStream( tempFile ) );
 
             writer.write( fw, configuration );
+            fw.flush();
+
+            //move original content out of the way, if it exists
+            if ( configurationFile.exists() )
```
kellyrob99: In the case of staging when starting up a new system there is no configuration file. Presumably that is a fairly common pattern as config files appear largely to be lazily initialized.

nexus-core/src/main/java/org/sonatype/nexus/util/configurationreader/DefaultConfigurationHelper.java
```diff
@@ -164,19 +184,7 @@
...
-                    fw.flush();
-
-                    fw.close();
-                }
-                catch ( IOException e )
-                {
-                    // just closing if open
-                }
-            }
+            IOUtils.closeQuietly( fw );
```
kellyrob99: If preferred we can use guava Closables here, but in any case this is an improvement

kellyrob99:
+1 by CI https://bamboo.zion.sonatype.com/browse/NXO-OSSF17-1

jdillon:
eh, I don't mean to be an ass... but ...

-1

I think org.sonatype.nexus.util.configurationreader.\* should be removed.  Its overly complicated for what its doing, and from briefly looking around nothing is using these classes.  I don't think this is the time to bring these back into usage.

The new usage of this component in staging only uses half of the framework, and the half it is uses its not using well as this framework wants to lock/unlock and the usage in staging pull/100 which manages the locks outside of this component can lead to confusion if developers are not aware that the lock is being released at the end of save() instead of int he finally block where they are currently (and properly) are unlocked.

nexus-core/src/main/java/org/sonatype/nexus/util/configurationreader/DefaultConfigurationHelper.java
```diff
@@ -144,19 +145,38 @@
...
         try
         {
-            fw = new OutputStreamWriter( new FileOutputStream( configurationFile ) );
+            configurationFile.getParentFile().mkdirs();
+            File tempFile = new File(configurationFile.getParentFile(), configurationFile.getName() + ".tmp");
+            File copy = new File(configurationFile.getParentFile(), configurationFile.getName() + ".old");
+
+            fw = new BufferedWriter( new FileWriter( tempFile ) );
 
             writer.write( fw, configuration );
```
kellyrob99: If we were on Java 7 this whole block could be reduced to:

```
Files.move( Paths.get( tempFile.toURI() ), Paths.get( configurationFile.toURI() ),
                        StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING );
```

kellyrob99:
No worries @jdillon your points are well taken.
I've simplified this down to only the bare essentials and removed the old dead code, so hopefully it fits the need a little better now.

nexus-core/src/main/java/org/sonatype/nexus/util/configuration/DefaultConfigurationHelper.java
```diff
@@ -0,0 +1,53 @@
...
+        try
+        {
+            configurationFile.getParentFile().mkdirs();
+
+            File tempFile = new File(configurationFile.getParentFile(), configurationFile.getName() + ".tmp");
+            fw = new BufferedWriter( new FileWriter( tempFile ) );
+            writer.write( fw, configuration );
+            fw.flush();
+
+            Files.move( tempFile, configurationFile );
```
cstamas, line 47: Moving file while it's writer is open.... will this work on Windows?

kellyrob99, line 47: No, as a matter of fact it won't. CI gave me false confidence I'm afraid. Tested and if we close first it works as expected on Windows, thanks @cstamas

nexus-core/src/main/java/org/sonatype/nexus/util/configuration/DefaultConfigurationHelper.java
```diff
@@ -0,0 +1,54 @@
...
+        {
+            configurationFile.getParentFile().mkdirs();
+
+            File tempFile = new File(configurationFile.getParentFile(), configurationFile.getName() + ".tmp");
+            fw = new BufferedWriter( new FileWriter( tempFile ) );
+            writer.write( fw, configuration );
+            fw.flush();
+            fw.close();
+
+            Files.move( tempFile, configurationFile );
```
kellyrob99, line 47: This is still system dependent as to what it actually does under the hood. If the system supports it a rename is done, if not a copy takes place, in which case the operation is no longer atomic. There didn't seem to be a consistent way to accomplish this in Java 6, and both common.io.FileUtils.move() and guava Files.move() depend on the native File.renameTo() method internally.
Anyone know a better way to do this until we can utilize the nio enhancements in Java 7?

In any event this should solve the problem on *nix systems.

kellyrob99:
Rendered obsolete by https://github.com/sonatype/nexus-staging-plugin/pull/102

